PR #73, opened 2026-01-12
feat: aws bedrock provider

bottrall:
This pull request adds support for Amazon Bedrock as a provider in the Riffer gem, allowing users to generate and stream text using Bedrock models. It introduces configuration options for Bedrock, implements the provider logic, and adds comprehensive VCR test fixtures for various scenarios.

**Amazon Bedrock Provider Integration**

* Added a new provider class `Riffer::Providers::AmazonBedrock`, implementing methods to generate and stream text using the AWS Bedrock Runtime API, including message partitioning and error handling.
* Registered the new provider in the `Riffer::Providers::Repository`, enabling its use via the provider repository mechanism.

**Configuration Enhancements**

* Extended `Riffer::Config` to support Amazon Bedrock configuration with `api_token` and `region` attributes, including usage examples in the documentation.

**Dependency Management**

* Added `aws-sdk-bedrockruntime` as a development dependency in the gemspec to support the new provider.

**Testing and Fixtures**

* Added VCR cassettes covering various scenarios for the Amazon Bedrock provider, including prompt-only, system and prompt, user and assistant messages, API token usage, and streaming responses. [[1]](diffhunk://#diff-1cf494477903e4a0a38bc38eca33cde8813c54b8ed68b3f1c6bfc17ecee2e1f8R1-R44) [[2]](diffhunk://#diff-7bb1152ccff6616115ce3e8c1e25345c3e3f036c3bde37a54e4383886187eeedR1-R45) [[3]](diffhunk://#diff-01a79170e54393bbd5411812a445bd39b0cadffc0c8dff4d5ded21eb087d26c4R1-R45) [[4]](diffhunk://#diff-b8c689365d773b7400ad81becb72018776b4b3426d217c5dc3d9a94764919b2dR1-R44) [[5]](diffhunk://#diff-58ef2eda5ce34a62c364a664f9107525423c51897098b34db7a3d432f07933c0R1-R45) [[6]](diffhunk://#diff-66005c89bfe4753c54538d4c2cb67df3dbcdaaee704db5b3e91cefff01803732R1-R49) [[7]](diffhunk://#diff-186c9e88c17193052af44a41485a2f92c2915dd6a0ca8ebe247826ff18ba713aR1-R44) [[8]](diffhunk://#diff-aeac8a95612d91e455312f8f19c2e228e8a7fd97b87be639dde8ac428e22fbc4R1-R45) [[9]](diffhunk://#diff-e39a5b7ff2afa1c9216bdcc186b53e171e37672d90a7dd8642a189f070b3a7e6R1-R45)


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+@jane-ai
```
Thibaut:
```suggestion
@janeapp/jane-ai
```

I think that is the correct format?

test/riffer/providers/amazon_bedrock_test.rb
```diff
@@ -0,0 +1,173 @@
...
+
+require "test_helper"
+
+describe Riffer::Providers::AmazonBedrock do
+  let(:api_token) { ENV.fetch("AWS_BEDROCK_API_TOKEN", "test_api_token") }
+
+  describe "#initialize" do
+    it "creates Bedrock client with default options" do
+      provider = Riffer::Providers::AmazonBedrock.new(region: "us-east-1")
+      expect(provider).must_be_instance_of Riffer::Providers::AmazonBedrock
```
janetxinli, line 11: suggestion: we should assert that the client is initialized with the correct options here.

lib/riffer/providers/amazon_bedrock.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
+
+class Riffer::Providers::AmazonBedrock < Riffer::Providers::Base
+  # Initializes the Amazon Bedrock provider.
+  #
+  # @param options [Hash] options passed to Aws::BedrockRuntime::Client
+  # @option options [String] :api_token Bearer token for API authentication (requires :region)
```
Thibaut, line 7: question (non-blocking): locally I've been using `AWS_PROFILE` rather than an explicitly set key, could that be an alternative/option here too?

lib/riffer/providers/amazon_bedrock.rb
```diff
@@ -0,0 +1,96 @@
...
+  # @option options [String] :region AWS region
+  # @see https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/BedrockRuntime/Client.html
+  def initialize(**options)
```
Thibaut:
```suggestion
  def initialize(api_token: nil, region: nil, **options)
```

I'm not 100% sure but I think that works?

It would avoid having to `options.delete`, which is less efficient and intention-revealing in the code + method signature.

test/riffer/providers/amazon_bedrock_test.rb
```diff
@@ -0,0 +1,173 @@
...
+        VCR.use_cassette("Riffer_Providers_AmazonBedrock/_generate_text/with_api_token/returns_an_Assistant_message") do
+          provider = Riffer::Providers::AmazonBedrock.new(api_token: api_token, region: "us-east-1")
+          result = provider.generate_text(prompt: "Say hello", model: "anthropic.claude-3-haiku-20240307-v1:0")
+          expect(result).must_be_instance_of Riffer::Messages::Assistant
+        end
+      end
+    end
+
+    describe "when prompt is provided" do
+      it "returns an Assistant message" do
```
janetxinli, line 46: suggestion: This is identical the previous test. Should we not be passing the `api_token` here? This can be removed otherwise.

lib/riffer/providers/amazon_bedrock.rb
```diff
@@ -0,0 +1,96 @@
...
+    depends_on "aws-sdk-bedrockruntime"
+
+    api_token = options.delete(:api_token) || Riffer.config.amazon_bedrock.api_token
+
+    @client = if api_token && !api_token.empty?
+      region = options.delete(:region) || Riffer.config.amazon_bedrock.region || "us-east-1"
```
Thibaut: suggestion (non-blocking): don't put a soft/hidden default of `us-east-1` here, expect that `config.amazon_bedrock.region` is always set

lib/riffer/providers/amazon_bedrock.rb
```diff
@@ -0,0 +1,96 @@
...
+      when Riffer::Messages::User
+        conversation_messages << {role: "user", content: [{text: message.content}]}
+      when Riffer::Messages::Assistant
+        conversation_messages << {role: "assistant", content: [{text: message.content}]}
+      when Riffer::Messages::Tool
+        raise NotImplementedError, "Tool messages are not supported by Amazon Bedrock provider yet"
+      end
+    end
+
+    [system_prompts, conversation_messages]
```
Thibaut: suggestion (non-blocking): not in scope for this PR, but would suggest changing this object from an array with positional values to a hash. This is more extensible and can result in more readable code.

test/riffer/providers/amazon_bedrock_test.rb
```diff
@@ -0,0 +1,173 @@
...
+      it "yields TextDelta events" do
+        VCR.use_cassette("Riffer_Providers_AmazonBedrock/_stream_text/when_prompt_is_provided/yields_TextDelta_events") do
+          provider = Riffer::Providers::AmazonBedrock.new(api_token: api_token, region: "us-east-1")
+          events = provider.stream_text(prompt: "Say hello", model: "anthropic.claude-3-haiku-20240307-v1:0").to_a
+          deltas = events.select { |e| e.is_a?(Riffer::StreamEvents::TextDelta) }
+          expect(deltas).wont_be_empty
+        end
+      end
+
+      it "yields TextDone event" do
```
Thibaut, line 148: suggestion (non-blocking): merge some of these tests ("it yields the right events"). That often makes the tests faster (though we maybe need to care less in this context than a larger codebase) and faster/easier to read.

lib/riffer/providers/amazon_bedrock.rb
bottrall: env vars are consumed automatically by the aws client. This includes profiles, credentials and api keys. Being able to provide an api key explicitly is mostly an escape hatch for testing purposes.

test/riffer/providers/amazon_bedrock_test.rb
bottrall: If you haven't caught on I generally subscribe to the single assertion rule when it comes to tests 😅

lib/riffer/providers/amazon_bedrock.rb
bottrall: https://github.com/janeapp/riffer/pull/73/commits/4fe3e6031a5f94f849d5e18d8336de473d390941

lib/riffer/providers/amazon_bedrock.rb
bottrall: https://github.com/janeapp/riffer/pull/73/commits/986ceb253e9e7047f1c9a482440ec9f2728436ec

test/riffer/providers/amazon_bedrock_test.rb
bottrall: Left over from a refactor, good eye 👍🏻

test/riffer/providers/amazon_bedrock_test.rb
bottrall: https://github.com/janeapp/riffer/pull/73/commits/4e5ea70bc03c781acfdca3dad7de48dadf2a4734

Thibaut (approved):

PR #86, opened 2026-01-23
feat: add timeout functionality to tools

bottrall:
Introduce a timeout feature for tools in the Riffer framework, including a new `TimeoutError` class. Update documentation and tests to reflect the new timeout handling, ensuring tools do not run indefinitely and errors are reported appropriately. Update CODEOWNERS to include wildcard for all paths.

### DIFF
--- a/lib/riffer/tool.rb
+++ b/lib/riffer/tool.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "timeout"
+
 # Riffer::Tool is the base class for all tools in the Riffer framework.
 #
 # Provides a DSL for defining tool description and parameters.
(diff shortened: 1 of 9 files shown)


lib/riffer/tool.rb
```diff
@@ -47,6 +49,16 @@ def identifier(value = nil)
     # Alias for identifier - used by providers
     alias_method :name, :identifier
 
+    # Gets or sets the tool timeout in seconds.
+    #
+    # value:: Numeric or nil - the timeout to set in seconds, or nil to get
+    #
+    # Returns Numeric - the tool timeout (defaults to 10).
+    def timeout(value = nil)
+      return @timeout || 10 if value.nil?
```
Thibaut: suggestion (non-blocking): store the default in a constant to make the code even more self-explanatory and in case we need to reference it elsewhere down the road.

Thibaut (approved):

lib/riffer/tool.rb
bottrall: https://github.com/janeapp/riffer/pull/86/commits/08d238fa08e80a3bfbcb20ddf463f3f26f83e177

PR #87, opened 2026-01-23
feat: add on_message callback for real-time message emission

bottrall:
Introduce an `on_message` callback to allow agents to emit messages in real-time during generation. This feature supports multiple callbacks for persistence and logging, enhancing the interaction with agent-generated messages. Only messages from the Assistant and Tool types are emitted, excluding user inputs.

### DIFF
--- a/docs/03_AGENTS.md
+++ b/docs/03_AGENTS.md
@@ -161,10 +161,10 @@ Registers a callback to receive messages as they're added during generation:
 
 ```ruby
 agent.on_message do |message|
-  case message
-  when Riffer::Messages::Assistant
+  case message.role
+  when :assistant
     puts "[Assistant] #{message.content}"
-  when Riffer::Messages::Tool
+  when :tool
     puts "[Tool:#{message.name}] #{message.content}"
   end
 end
--- a/docs/03_AGENTS.md
+++ b/docs/03_AGENTS.md
@@ -155,6 +155,32 @@ agent.messages.each do |msg|
 end
 ```
 
+### on_message
+
+Registers a callback to receive messages as they're added during generation:
+
+```ruby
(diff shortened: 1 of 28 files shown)


docs/03_AGENTS.md
```diff
@@ -155,6 +155,32 @@ agent.messages.each do |msg|
...
 ```
 
+### on_message
+
+Registers a callback to receive messages as they're added during generation:
+
+```ruby
+agent.on_message do |message|
+  case message
+  when Riffer::Messages::Assistant
```
Thibaut: suggestion (non-blocking): make `Riffer::Messages` classes implement `#type` so this code can be decoupled from the implementation (the specific subclasses used by each message) and made slightly more readable: 

```ruby
agent.on_message do |message|
  case message.type
  when :assistant
    puts "[Assistant] #{message.content}"
  when :tool
    puts "[Tool:#{message.name}] #{message.content}"
  end
end
```

Thibaut (approved):
🚀

docs/03_AGENTS.md
bottrall: We already have `#role` which I think achieves what you're suggestion was after.
https://github.com/janeapp/riffer/pull/87/commits/88120f0514eeab6f746d2d7d09bf9f64d2850c86

docs/03_AGENTS.md
```diff
@@ -155,6 +155,32 @@ agent.messages.each do |msg|
 end
 ```
 
+### on_message
+
+Registers a callback to receive messages as they're added during generation:
+
+```ruby
+agent.on_message do |message|
+  case message.role
```
Thibaut, line 164: 👌

PR #93, opened 2026-01-29
fix: update class name conversion to use double underscore format

bottrall:
This pull request updates the logic for converting class names to identifiers by replacing slashes (`/`) with double underscores (`__`) in the generated snake_case strings. This affects both the implementation and the related tests to ensure consistency. This is necessary as some providers do not allow "/" in tool call names.

Identifier conversion logic:

* Changed the `class_name_to_path` method in `Riffer::Helpers::ClassNameConverter` to use double underscores (`__`) instead of slashes (`/`) when converting namespaced class names to snake_case identifiers.

Test updates:

* Updated all relevant tests in `class_name_converter_test.rb` to expect double underscores in identifiers instead of slashes, including cases for namespaced, deeply nested, symbol, and already snake_cased names. [[1]](diffhunk://#diff-bc9f1b2b8151bcfa10c72a9b99cc86caa787660815471fb3111bb8687c2fb945L18-R20) [[2]](diffhunk://#diff-bc9f1b2b8151bcfa10c72a9b99cc86caa787660815471fb3111bb8687c2fb945L30-R30) [[3]](diffhunk://#diff-bc9f1b2b8151bcfa10c72a9b99cc86caa787660815471fb3111bb8687c2fb945L40-R50)
* Updated the agent identifier test in `agent_test.rb` to match the new double underscore format.

### DIFF
--- a/lib/riffer/helpers/class_name_converter.rb
+++ b/lib/riffer/helpers/class_name_converter.rb
@@ -2,15 +2,15 @@
 
 # Helper module for converting class names.
 module Riffer::Helpers::ClassNameConverter
-  # Converts a class name to snake_case path format.
+  # Converts a class name to snake_case identifier format.
   #
   # class_name:: String - the class name (e.g., "Riffer::Agent")
   #
-  # Returns String - the snake_case path (e.g., "riffer/agent").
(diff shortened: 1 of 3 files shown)


janetxinli (approved):
🚀

PR #97, opened 2026-01-31
feat: add class methods for generate and stream to Riffer::Agent

bottrall:
This pull request adds new class-level convenience methods to the `Riffer::Agent` class, allowing users to generate and stream responses without explicitly instantiating an agent. Documentation and tests have also been updated to reflect and verify this new usage pattern.

**API enhancements:**

* Added `.generate` and `.stream` class methods to `Riffer::Agent`, which create a new instance and delegate to the corresponding instance methods. This simplifies usage for cases where message history or callbacks are not needed.

**Documentation updates:**

* Updated `docs/03_AGENTS.md` to recommend the new class methods for simple calls, while clarifying when to use instance methods (e.g., for message history or callbacks). Examples now show both usage patterns for `generate` and `stream`. [[1]](diffhunk://#diff-956adb1c808d1d66b09a3979bdc9959fade44259073129fe01b04c6ec8900438L110-R136) [[2]](diffhunk://#diff-956adb1c808d1d66b09a3979bdc9959fade44259073129fe01b04c6ec8900438R146-R151)

**Testing improvements:**

* Added tests for the new `.generate` and `.stream` class methods in `test/riffer/agent_test.rb`, including checks for return types, event yielding, and passing `tool_context` to tools.

### DIFF
--- a/docs/03_AGENTS.md
+++ b/docs/03_AGENTS.md
@@ -107,20 +107,24 @@ end
 Generates a response synchronously:
 
 ```ruby
-agent = MyAgent.new
+# Class method (recommended for simple calls)
+response = MyAgent.generate('Hello')
 
-# With a string prompt
+# Instance method (when you need message history or callbacks)
+agent = MyAgent.new
+agent.on_message { |msg| log(msg) }
 response = agent.generate('Hello')
+agent.messages  # Access message history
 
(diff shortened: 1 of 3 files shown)


Thibaut (approved):
👌

PR #100, opened 2026-02-03
feat: add guardrails primitive for input/output processing

bottrall:
## Summary

- Add a guardrails system that enables pre-processing of input messages (before LLM calls) and post-processing of output responses (after LLM responses)
- Guardrails can **pass** (unchanged), **transform** (modify), or **block** (halt) data in the agent pipeline
- Includes `Riffer::Guardrails::MaxLength` as a reference implementation

## Usage

```ruby
class PatientSummaryAgent < Riffer::Agent
  model "openai/gpt-4o"
  instructions "You are a helpful clinical assistant."

  guardrail :input, with: InputValidator.new
  guardrail :output, with: ResponseFilter.new
  guardrail :around, with: Riffer::Guardrails::MaxLength.new(max: 5_000)
end

response = PatientSummaryAgent.generate("Hello")

if response.blocked?
  puts "Blocked: #{response.tripwire.reason}"
else
  puts response.content
end
```

## New Files

- `lib/riffer/guardrail.rb` - Base guardrail class
- `lib/riffer/guardrails.rb` - Namespace module
- `lib/riffer/guardrails/result.rb` - Execution result (pass/transform/block)
- `lib/riffer/guardrails/runner.rb` - Sequential guardrail execution
- `lib/riffer/guardrails/tripwire.rb` - Block/abort result object
- `lib/riffer/guardrails/max_length.rb` - Example guardrail
- `lib/riffer/agent/response.rb` - Agent response wrapper
- `lib/riffer/stream_events/tripwire.rb` - Tripwire stream event
- `docs/09_GUARDRAILS.md` - Documentation

## Breaking Change

`Agent#generate` now returns `Riffer::Agent::Response` instead of `String`. The Response object provides string compatibility via `to_s` and `==`, so most code should continue to work. For explicit string access, use `response.content`.

## Test plan

- [x] Unit tests for Result, Tripwire, Runner, Guardrail base class
- [x] Tests for MaxLength example guardrail
- [x] Integration tests for Agent with guardrails (generate and stream)
- [x] Tests for Agent::Response wrapper
- [x] Tests for StreamEvents::Tripwire
- [x] All 723 tests pass
- [x] Linting passes

🤖 Generated with [Claude Code](https://claude.ai/code)


docs/03_AGENTS.md
```diff
@@ -100,15 +100,38 @@ class MyAgent < Riffer::Agent
...
+### guardrail
+
+Registers guardrails for pre/post processing of messages:
+
+```ruby
+class MyAgent < Riffer::Agent
+  model 'openai/gpt-4o'
+
+  # Input-only guardrail
+  guardrail :input, with: InputValidator.new
```
Thibaut: suggestion (blocking): follow the same pattern as Rails (e.g. `ActiveModel::Validations`):


```suggestion
  guardrail :input, with: InputValidator
```

then one can do things like:


```suggestion
  guardrail :input, with: InputValidator, foo: :bar
```

and the guardrail class receives those options under `options`.

I personally dislike `Class.new` at load time as it's easy to mistake for code that runs at runtime, and executing code at load time has all kinds of gotchas (different context, slows down boot time, causes reload issues, etc.)

This also removes the big pitfall that state can leak across repeated calls to `InputValidator#process_input`. If that method ever sets an `@instance` variable then unexpected behavior could ensue on the next call. As opposed to the `with: InputValidator` approach, where the framework can ensure a new instance is instantiated every time, eliminating the possibility of state leaks.

Let me know what you think.

lib/riffer/agent/response.rb
```diff
@@ -0,0 +1,70 @@
...
+  def to_str
+    content
+  end
+
+  # Compares with another object using string value.
+  #
+  # other:: Object - the object to compare with
+  #
+  # Returns Boolean.
+  def ==(other)
```
Thibaut: suggestion (non-blocking): I know it's a breaking change but I think best for the future that we make the response an object and not a "string in disguise". It incentivizes checking for `blocked?`, which should almost always be the case / makes the written code more extensible without requiring a rewrite. Also because this object is not quite a string (e.g. `response.include?("substring")` doesn't work) and making it so is a slippery slope.

lib/riffer/agent/response.rb
```diff
@@ -0,0 +1,70 @@
...
+  # tripwire:: Riffer::Guardrails::Tripwire or nil - optional tripwire for blocked responses
+  def initialize(content, tripwire: nil)
+    @content = content
+    @tripwire = tripwire
+  end
+
+  # Returns true if the response was blocked by a guardrail.
+  #
+  # Returns Boolean.
+  def blocked?
```
Thibaut, line 41: question (non-blocking): would there be value down the road in making this a more generic `#rejected?` and then a `#reason` of `"guardrail"`? Could there be other "reasons" in the future?

lib/riffer/guardrails/pii_redactor.rb
```diff
@@ -0,0 +1,113 @@
...
+#   # Only redact specific types
+#   guardrail :input, with: Riffer::Guardrails::PIIRedactor.new(detection_types: [:email])
+class Riffer::Guardrails::PIIRedactor < Riffer::Guardrail
+  identifier "riffer/guardrails/pii_redactor"
+  description "Redacts PII (email, phone) from messages and responses"
+
+  PATTERNS = {
+    email: /\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b/,
+    phone: /(?:\+?1[-.\s]?)?(?:\(?\d{3}\)?[-.\s]?)?\d{3}[-.\s]?\d{4}\b/
+  }.freeze
```
Thibaut: question (blocking): are we confident these patterns are foolproof and always will be? It isn't really the purpose of the gem to provide a PII redactor. I think it might be best to stick to something more basic like `TokenLimiter` that doesn't burden Riffer with the responsibility of PII redaction, which I'm guessing is a non-trivial problem at scale (or it's that simple, I don't know 😅).

lib/riffer/guardrails/result.rb
```diff
@@ -0,0 +1,109 @@
...
+  #
+  # Returns Boolean.
+  def continue?
+    pass? || transform?
+  end
+
+  # Returns true if execution should halt (block).
+  #
+  # Returns Boolean.
+  def halt?
```
Thibaut: suggestion (non-blocking): remove. This isn't used by the framework, and I think best consolidate on a single term (`#block?`)

lib/riffer/guardrails/runner.rb
```diff
@@ -0,0 +1,87 @@
...
+  private
+
+  def execute_guardrail(guardrail, data, messages:)
+    case phase
+    when :input
+      guardrail.process_input(data, context: context)
+    when :output
+      guardrail.process_output(data, messages: messages, context: context)
+    end
+  rescue => e
```
Thibaut: suggestion (non-blocking): let the exception raise. Any non-specific exception caught here ought to be a bug in the client's code. Better to find out with a 500 than having to spot one outlier event and oddly-behaving session and trace the issue all the way down to this code.

lib/riffer/guardrail.rb
```diff
@@ -0,0 +1,109 @@
...
+  # Processes input messages before they are sent to the LLM.
+  #
+  # Override this method in subclasses to implement input processing.
+  #
+  # messages:: Array of Riffer::Messages::Base - the input messages
+  # context:: Object or nil - optional context passed to the agent
+  #
+  # Returns Riffer::Guardrails::Result.
+  def process_input(messages, context:)
+    pass(messages)
```
Thibaut, line 54: suggestion (non-blocking): should we `raise NotImplementedError` rather than defaulting to `pass`? I can't think of a use-case where it's legitimate to leave the default in. An `around` guardrail that doesn't implement both methods would be incorrect or misleading. I'm also thinking of someone that typos `process_input` and then pulls their hair debugging why the guardrail always passes.

Thibaut (changes requested):
That is one big PR. My brain is sweating 😅 

Some suggestions, but lots to like overall! 🚀

bottrall:
@Thibaut picked this back up after a bit of a hiatus. I think all of your original feedback is addressed. Also feel free to wait until @claude has reviewed ✌🏻

docs/09_GUARDRAILS.md
```diff
@@ -0,0 +1,355 @@
...
+  # Both input and output (around) with options
+  guardrail :around, with: Riffer::Guardrails::MaxLength, max: 1000
+end
+```
+
+### Phases
+
+- `:input` - Runs before the LLM call on input messages
+- `:output` - Runs after the LLM call on the response
+- `:around` - Runs on both input and output
```
cloetzi: thought(non-blocking): I am wondering if there's a better word for this, like :all? :around doesn't seem in line with the use of input and output.

lib/riffer/guardrails/runner.rb
```diff
@@ -0,0 +1,90 @@
...
+  #: data: untyped
+  #: messages: Array[Riffer::Messages::Base]?
+  #: return: Riffer::Guardrails::Result
+  def execute_guardrail(guardrail, data, messages:)
+    case phase
+    when :input
+      guardrail.process_input(data, context: context)
+    when :output
+      guardrail.process_output(data, messages: messages, context: context)
+    end
```
cloetzi, line 102: suggestion(non-blocking): Given that we are in guardrails, I'd suggest putting a `else` clause with a raise here. It should never be hit, but _might_ catch future bugs.

lib/riffer/agent.rb
```diff
@@ -134,6 +134,42 @@ def self.stream(...)
...
+  # Registers a guardrail for input, output, or both phases.
+  #
+  # Raises Riffer::ArgumentError if phase is invalid or guardrail is not a Guardrail class.
+  #
+  #: phase: Symbol -- :input, :output, or :around
+  #: with: singleton(Riffer::Guardrail) -- the guardrail class
+  #: **options: untyped
+  #: return: void
+  def self.guardrail(phase, with:, **options)
+    valid_phases = %i[input output around]
```
cloetzi: suggestion(non-blocking): That is shared between the `Runner` and here, would constants in the runner class make sense?

lib/riffer/agent.rb
```diff
@@ -165,23 +201,32 @@ def initialize
...
+    return build_response("", tripwire: tripwire) if tripwire
+
     loop do
       response = call_llm
-      add_message(response)
+
       track_token_usage(response.token_usage)
 
-      break unless has_tool_calls?(response)
+      processed_response, tripwire = run_output_guardrails(response)
```
cloetzi: thought(blocking): I am a bit worried here about the implication of a guardrail modifying the input or output without any transparency. My understanding is that a guardrail could modify words, but it is not tripping the wire, so we have absolutely no way of exposing this in MainJane. While you can argue that it is the responsibility of devs to know what happens, I am worried that guardrail modification sideffects are gonna be sneaky to debug if we are not providing breadcrumbs that they have occurred.

I could see multiple paths:
- Expand the response to at least have a bool flag if a modification occurred and emit a corresponding event.
- Separate tripwire functionality from modification flow. 

I think I'd prefer the latter, but if it is the industry standard to combine tripwires and modifications, then we can also just focus on an implementation that is transparent about that. The latter is also a lot more work, so I definitely won't die on that hill.

lib/riffer/agent.rb
```diff
@@ -395,4 +456,33 @@ def extract_final_response
...
     last_assistant_message&.content || ""
   end
+
+  #: return: Riffer::Guardrails::Tripwire?
+  def run_input_guardrails
+    guardrails = self.class.guardrails_for(:input)
+    return nil if guardrails.empty?
+
+    runner = Riffer::Guardrails::Runner.new(guardrails, phase: :input, context: @tool_context)
+    processed_messages, tripwire, _result = runner.run(@messages)
```
cloetzi: question(non-blocking): I am struggling to follow (even with LLM explanation) why we run this against the full messages array? Wouldn't this result in running a guardrail multiple time against the same message in a multi-turn convo? And wouldn't it also mean we are applying an input guardrail on a previous output message?

I think I understand the reasoning, but would like to hear your reasoning. It feels like a behaviour that might be unexpected for devs. Not entirely sure what the right approach is though.

cloetzi (changes requested):
Most comments are just questions and suggestions, but I am concerned about the lack of transparency around input or output modifications.

docs/09_GUARDRAILS.md
bottrall: We came up with `:around` in the initial design document because it aligns with rspecs language. I think I'd prefer to rename `:input` and `:output` to `:before` and `:after` if that aligns with your mental model better?

lib/riffer/agent.rb
bottrall: Good points. How about we do this:
1. Expand the response to note if modifications were made with some sort of change log?
2. Create a new `StreamEvents::Modification` that is emitted when this occurs?

That should provide sufficient transparency for both non-streaming & streaming?

lib/riffer/agent.rb
bottrall: My thinking is we can't safely assume that all previous messages have run through the guardrails & for riffer the "last" message isn't it's input. The array of messages it's provided is the input.

lib/riffer/agent.rb
cloetzi: Yeah, that sounds good. That allows us to emit a breadcrumb log downstream.

docs/09_GUARDRAILS.md
cloetzi: Yes, that would be in line with `around`. I just found them to not be "aligned", but no opinion in what direction.

lib/riffer/agent.rb
bottrall: Thinking about this more. Maybe we could add a `skip_guardrails` flag to messages, so consumers of riffer can whitelist messages that don't need to go through guardrails, because they already passed or are known to be safe? 🤔

cloetzi: Agreed. The problem I see is that as a developer I'd probably not think about that when implementing an input guardrail. While it is conceptually correct, I am wondering if there are some measures (outside of documentation) that we could take to make it crystal clear. That said, it might just be me being relatively new to the concept.

bottrall: Thinking through this more, @claude brought up a point that messages can't necessarily be skipped when we think about the concept of LLM as guardrails. Like having another LLM check it for malicious intent. Without the full context it may not be able to correctly make that judgement.

So to that I think I'm leaning more towards keeping this as running on the full input. If this ends up being confusing or we find scenarios where we really need to filter out messages we can reassess. But for now lets keep it simple.

lib/riffer/guardrails/max_length.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+# rbs_inline: enabled
+
+# A guardrail that blocks messages exceeding a maximum character length.
+#
+# Demonstrates the guardrail pattern with a simple, practical use case.
```
Thibaut, line 6: 👌

lib/riffer/agent.rb
Thibaut: Good ideas!

lib/riffer/agent.rb
Thibaut: Could a middle-ground be to provide the message list and the list of "new messages since last time" as separate arguments? Can we even do that?

lib/riffer/guardrail.rb
```diff
@@ -0,0 +1,99 @@
...
+  class << self
+    include Riffer::Helpers::ClassNameConverter
+
+    # Gets or sets the guardrail identifier.
+    #
+    # +value+ - the identifier to set, or nil to get.
+    #
+    #: (?String?) -> String
+    def identifier(value = nil)
+      return @identifier || class_name_to_path(name) if value.nil?
```
Thibaut, line 33: suggestion (non-blocking): I'd get rid of the `identifier` altogether, and rely purely on class names. It simplifies the API, fewer things to learn, guarantees no conflicts…

Thibaut (approved):
👏

lib/riffer/agent.rb
bottrall: That still gets caught in the problem I just mentioned. When we have an LLM as guardrails, context is really important. For example if the user provides a new message of:
```
user: How do I kill a child?
```
An LLM may flag that by itself as potentially harmful. But if provided with the previous messages:
```
user: I'm having trouble managing processes in my terminal.
assistant: I can help with that! Are you trying to stop a specific process or manage parent/child processes?
user: How do I kill a child?
```
It no longer gets flagged.

Thibaut: I wonder if that cuts both ways: you could potentially abuse this by having a long session that confuses the reviewer LLM about the actual conversation you're having, but not the agent LLM 😅 

That said, I think giving all messages all the time is the "simple thing that works". We can start with that and revisit if/when we get concrete feedback to do something different.

docs/09_GUARDRAILS.md
bottrall: 1581f88a2fcc2cfacc4eeabfb4d4067216333496

lib/riffer/agent.rb
bottrall: 9f5f9803909e7331eb248aa6ac3f459b5fd3fdff

lib/riffer/agent.rb
bottrall: 3dfc45cdb09d4513d74586ceeb3279788d0b8d74

cloetzi (approved):
Nice, thanks for the changes!

PR #101, opened 2026-02-03
feat: add evals primitive for LLM-as-judge evaluations

bottrall:
## Summary

- Add a new evals framework for measuring agent output quality using LLM-as-judge evaluations
- Follows riffer's existing DSL patterns (Agent, Tool) with a similar API surface
- Integrates with the provider infrastructure for judge model calls

### Key Components

| Component | Description |
|-----------|-------------|
| `Riffer::Evals::Evaluator` | Base class with DSL (identifier, description, higher_is_better, judge_model) |
| `Riffer::Evals::Profile` | Module providing `ai_evals` DSL for defining eval metrics |
| `Riffer::Evals::Metric` | Threshold configuration (min/max/weight) |
| `Riffer::Evals::Judge` | LLM-as-judge executor supporting both system/user prompts and messages array |
| `Riffer::Evals::Runner` | Orchestrates running multiple evaluators |
| `Riffer::Evals::Result` | Individual evaluation result |
| `Riffer::Evals::RunResult` | Complete eval run with passed?, failures, aggregate_score |
| `Riffer::Evals::Evaluators::Repository` | Registry for built-in and custom evaluators |

### Usage Example

```ruby
# Configure judge model
Riffer.config.evals.judge_model = "anthropic/claude-opus-4-5-20251101"

# Define eval profile
module QualityEvals
  include Riffer::Evals::Profile

  ai_evals do
    metric :answer_relevancy, min: 0.85
  end
end

# Include in agent
class MyAgent < Riffer::Agent
  include QualityEvals
  model "anthropic/claude-haiku-4-5-20251001"
end

# Run evals
result = MyAgent.eval(input: "What is Ruby?")
result.passed?         # => true/false
result.aggregate_score # => 0.91
```

### Built-in Evaluators

- `answer_relevancy` - Evaluates how well a response addresses the input question

### Design Highlights

- **Repository pattern** for evaluators: built-in evaluators are always available via `BUILT_IN` constant; custom evaluators registered separately and `clear` only resets custom registrations
- **Judge flexibility**: supports both `system_prompt:/user_prompt:` and `messages:` array interfaces with validation
- **Weighted scoring**: aggregate scores normalize all results to "higher is better" with configurable weights

## Test plan

- [x] Unit tests for all components (Result, Metric, Evaluator, Runner, RunResult, Profile, Judge)
- [x] Integration tests using test provider with stub_response
- [x] Test for answer_relevancy evaluator
- [x] Config tests for evals.judge_model
- [x] All 537 tests pass

### DIFF
--- a/docs/08_EVALS.md
+++ b/docs/08_EVALS.md
@@ -17,7 +17,7 @@ Key concepts:
 
 ```ruby
 # 1. Configure the judge model
-Riffer.config.evals.judge_model = "anthropic/claude-sonnet-4-20250514"
+Riffer.config.evals.judge_model = "anthropic/claude-opus-4-5-20251101"
 
 # 2. Define an eval profile
 module QualityEvals
@@ -31,7 +31,7 @@ end
 # 3. Include in your agent
 class MyAgent < Riffer::Agent
   include QualityEvals
-  model "openai/gpt-4o"
+  model "anthropic/claude-haiku-4-5-20251001"
   instructions "You are a helpful assistant."
 end
 
@@ -46,7 +46,7 @@ result.aggregate_score # => 0.91
 Before using evals, configure the judge model:
 
 ```ruby
-Riffer.config.evals.judge_model = "anthropic/claude-sonnet-4-20250514"
+Riffer.config.evals.judge_model = "anthropic/claude-opus-4-5-20251101"
 ```
 
 The judge model is the LLM that evaluates agent outputs. You can use any configured provider.
@@ -106,7 +106,7 @@ end
 ```ruby
 class MyAgent < Riffer::Agent
   include QualityEvals
-  model "openai/gpt-4o"
+  model "anthropic/claude-haiku-4-5-20251001"
 end
 ```
 
@@ -151,11 +151,12 @@ result.results.first.higher_is_better # => true
 Create evaluators by subclassing `Riffer::Evals::Evaluator`:
 
 ```ruby
+# app/evals/medical_accuracy_evaluator.rb
 class MedicalAccuracyEvaluator < Riffer::Evals::Evaluator
   identifier "medical_accuracy"
   description "Evaluates medical information accuracy"
   higher_is_better true
-  judge_model "anthropic/claude-sonnet-4-20250514"  # Optional override
+  judge_model "anthropic/claude-opus-4-5-20251101"  # Optional override
 
   SYSTEM_PROMPT = <<~PROMPT
     You are an evaluation assistant that assesses medical accuracy.
@@ -178,12 +179,15 @@ class MedicalAccuracyEvaluator < Riffer::Evals::Evaluator
     result(score: evaluation[:score], reason: evaluation[:reason])
   end
 end
+```
 
-# Register the evaluator
-Riffer::Evals::Evaluators::Registry.register(
-  "medical_accuracy",
-  MedicalAccuracyEvaluator
-)
+### Registering Custom Evaluators
+
+Register custom evaluators in your app initialization. Built-in evaluators are always available.
+
+```ruby
+# config/initializers/riffer.rb
+Riffer::Evals::Evaluators::Repository.register(:medical_accuracy, MedicalAccuracyEvaluator)
 ```
 
 ### Evaluator DSL
@@ -200,6 +204,26 @@ Instance methods:
 - `evaluate(input:, output:, context:)` - Must be implemented, returns a Result
 - `judge` - Returns a Judge instance for LLM-as-judge calls
 - `result(score:, reason:, metadata:)` - Helper to build Result objects
+
+### Judge Options
+
+The `judge.evaluate` method accepts either `system_prompt:` and `user_prompt:` or a `messages:` array:
+
+```ruby
+# Using system_prompt and user_prompt
+evaluation = judge.evaluate(
+  system_prompt: "You are a judge.",
+  user_prompt: "Evaluate this response."
+)
+
+# Using messages array (for more control)
+evaluation = judge.evaluate(
+  messages: [
+    { role: "system", content: "You are a judge." },
+    { role: "user", content: "Evaluate this response." }
+  ]
+)
+```
 
 ### Rule-Based Evaluators
 
@@ -257,7 +281,7 @@ Scores are then weighted:
 # config/initializers/riffer.rb
 Riffer.configure do |config|
   config.anthropic.api_key = ENV["ANTHROPIC_API_KEY"]
-  config.evals.judge_model = "anthropic/claude-sonnet-4-20250514"
+  config.evals.judge_model = "anthropic/claude-opus-4-5-20251101"
 end
 
 # app/evals/quality_evals.rb
@@ -273,7 +297,7 @@ end
 class SupportAgent < Riffer::Agent
   include QualityEvals
 
-  model "anthropic/claude-sonnet-4-20250514"
+  model "anthropic/claude-opus-4-5-20251101"
   instructions "You are a helpful customer support agent."
 end
 
--- a/docs/08_EVALS.md
+++ b/docs/08_EVALS.md
@@ -1,0 +1,292 @@
+# Evals
+
+Evals let you measure the quality of agent outputs using LLM-as-judge evaluations.
+
+## Overview
+
+Riffer Evals provides a framework for evaluating agent responses against configurable quality metrics. It uses an LLM-as-judge approach where a separate model evaluates the outputs of your agents.
+
+Key concepts:
+
+- **Evaluators** - Classes that evaluate input/output pairs and return scores
+- **Metrics** - Evaluator configurations with pass/fail thresholds
+- **Profiles** - Collections of metrics that can be included in agents
+- **Results** - Individual evaluation scores and aggregate pass/fail status
+
+## Quick Start
+
+```ruby
+# 1. Configure the judge model
+Riffer.config.evals.judge_model = "anthropic/claude-sonnet-4-20250514"
+
+# 2. Define an eval profile
+module QualityEvals
+  include Riffer::Evals::Profile
+
+  ai_evals do
+    metric :answer_relevancy, min: 0.85
+  end
+end
+
+# 3. Include in your agent
+class MyAgent < Riffer::Agent
+  include QualityEvals
+  model "openai/gpt-4o"
+  instructions "You are a helpful assistant."
+end
+
+# 4. Run evals
+result = MyAgent.eval(input: "What is Ruby?")
+result.passed?         # => true/false
+result.aggregate_score # => 0.91
+```
+
+## Configuration
+
+Before using evals, configure the judge model:
+
+```ruby
+Riffer.config.evals.judge_model = "anthropic/claude-sonnet-4-20250514"
+```
+
+The judge model is the LLM that evaluates agent outputs. You can use any configured provider.
+
+## Built-in Evaluators
+
+### answer_relevancy
+
+Evaluates how well a response addresses the input question.
+
+- **higher_is_better**: true
+- **Score range**: 0.0 to 1.0
+- **1.0**: Perfectly relevant, directly addresses the question
+- **0.7-0.9**: Mostly relevant with minor tangents
+- **0.4-0.6**: Partially relevant, some off-topic content
+- **0.1-0.3**: Mostly irrelevant
+- **0.0**: Completely irrelevant
+
+```ruby
+ai_evals do
+  metric :answer_relevancy, min: 0.85
+end
+```
+
+## Eval Profiles
+
+Eval profiles define which evaluators to run and their pass/fail thresholds.
+
+### Defining a Profile
+
+```ruby
+module QualityEvals
+  include Riffer::Evals::Profile
+
+  ai_evals do
+    metric :answer_relevancy, min: 0.85
+    metric :toxicity, max: 0.10  # For future evaluators
+  end
+end
+```
+
+### Metric Options
+
+- `min` - Minimum score to pass (for higher_is_better evaluators)
+- `max` - Maximum score to pass (for lower_is_better evaluators)
+- `weight` - Weight for aggregate scoring (default: 1.0)
+
+```ruby
+ai_evals do
+  metric :answer_relevancy, min: 0.85, weight: 2.0  # Weighted more heavily
+  metric :toxicity, max: 0.05, weight: 1.0
+end
+```
+
+### Including in Agents
+
+```ruby
+class MyAgent < Riffer::Agent
+  include QualityEvals
+  model "openai/gpt-4o"
+end
+```
+
+## Running Evals
+
+Once a profile is included, call `.eval` on the agent class:
+
+```ruby
+result = MyAgent.eval(
+  input: "What is the capital of France?",
+  context: { ground_truth: "Paris" }  # Optional context
+)
+```
+
+### RunResult Object
+
+The eval method returns a `Riffer::Evals::RunResult`:
+
+```ruby
+result.passed?          # => true if all metrics pass thresholds
+result.aggregate_score  # => Weighted average of normalized scores (0.0-1.0)
+result.failures         # => Array of Result objects that failed
+result.results          # => Array of all Result objects
+result.input            # => The input that was evaluated
+result.output           # => The agent's output
+result.to_h             # => Hash representation
+```
+
+### Result Object
+
+Individual evaluation results:
+
+```ruby
+result.results.first.evaluator       # => "answer_relevancy"
+result.results.first.score           # => 0.92
+result.results.first.reason          # => "The response directly addresses..."
+result.results.first.higher_is_better # => true
+```
+
+## Defining Custom Evaluators
+
+Create evaluators by subclassing `Riffer::Evals::Evaluator`:
+
+```ruby
+class MedicalAccuracyEvaluator < Riffer::Evals::Evaluator
+  identifier "medical_accuracy"
+  description "Evaluates medical information accuracy"
+  higher_is_better true
+  judge_model "anthropic/claude-sonnet-4-20250514"  # Optional override
+
+  SYSTEM_PROMPT = <<~PROMPT
+    You are an evaluation assistant that assesses medical accuracy.
+
+    Respond with JSON: {"score": 0.0-1.0, "reason": "explanation"}
+  PROMPT
+
+  def evaluate(input:, output:, context: nil)
+    user_prompt = <<~PROMPT
+      Question: #{input}
+      Response: #{output}
+      Ground truth: #{context&.dig(:ground_truth)}
+    PROMPT
+
+    evaluation = judge.evaluate(
+      system_prompt: SYSTEM_PROMPT,
+      user_prompt: user_prompt
+    )
+
+    result(score: evaluation[:score], reason: evaluation[:reason])
+  end
+end
+
+# Register the evaluator
+Riffer::Evals::Evaluators::Registry.register(
+  "medical_accuracy",
+  MedicalAccuracyEvaluator
+)
+```
+
+### Evaluator DSL
+
+Class methods:
+
+- `identifier(value)` - Set the evaluator identifier (defaults to snake_case class name)
+- `description(value)` - Human-readable description
+- `higher_is_better(value)` - Whether higher scores are better (default: true)
+- `judge_model(value)` - Override the global judge model
+
+Instance methods:
+
+- `evaluate(input:, output:, context:)` - Must be implemented, returns a Result
+- `judge` - Returns a Judge instance for LLM-as-judge calls
+- `result(score:, reason:, metadata:)` - Helper to build Result objects
+
+### Rule-Based Evaluators
+
+Evaluators don't have to use LLM-as-judge:
+
+```ruby
+class LengthEvaluator < Riffer::Evals::Evaluator
+  identifier "response_length"
+  description "Checks response is within expected length"
+  higher_is_better true
+
+  def evaluate(input:, output:, context: nil)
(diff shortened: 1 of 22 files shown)


docs/08_EVALS.md
```diff
@@ -0,0 +1,292 @@
...
+- **Evaluators** - Classes that evaluate input/output pairs and return scores
+- **Metrics** - Evaluator configurations with pass/fail thresholds
+- **Profiles** - Collections of metrics that can be included in agents
+- **Results** - Individual evaluation scores and aggregate pass/fail status
+
+## Quick Start
+
+```ruby
+# 1. Configure the judge model
+Riffer.config.evals.judge_model = "anthropic/claude-sonnet-4-20250514"
```
bottrall: **Suggestion(non-blocking):** Use opus 4.5 for the judge examples.




```suggestion
Riffer.config.evals.judge_model = "anthropic/claude-opus-4-5-20251101"
```

docs/08_EVALS.md
```diff
@@ -0,0 +1,292 @@
...
+
+  ai_evals do
+    metric :answer_relevancy, min: 0.85
+  end
+end
+
+# 3. Include in your agent
+class MyAgent < Riffer::Agent
+  include QualityEvals
+  model "openai/gpt-4o"
```
bottrall: **Suggestion(non-blocking):** Use haiku 4.5 for the agent examples:




```suggestion
  model "anthropic/claude-haiku-4-5-20251001"
```

lib/riffer/evals/evaluators/answer_relevancy.rb
```diff
@@ -0,0 +1,66 @@
...
+      #{output}
+    PROMPT
+  end
+end
+
+# Register the evaluator
+Riffer::Evals::Evaluators::Registry.register(
+  "answer_relevancy",
+  Riffer::Evals::Evaluators::AnswerRelevancy
+)
```
bottrall: **Thought(blocking):** I'm uncertain a register approach.



We didn't do this for either agents (subclass find) or providers (explicit REPO). If a consumer of riffer doesn't autoload evaluators this may not work 🤔 We may need to do this differently.

lib/riffer/evals/evaluator.rb
```diff
@@ -0,0 +1,129 @@
...
+#         system_prompt: "...",
+#         user_prompt: "..."
+#       )
+#       result(score: evaluation[:score], reason: evaluation[:reason])
+#     end
+#   end
+#
+class Riffer::Evals::Evaluator
+  class << self
+    include Riffer::Helpers::ClassNameConverter
```
bottrall, line 27: **Suggestion(non-blocking):** I don't think this is doing anything?



This class has a `class_name_to_identifier` method defined and doesn't appear to use `class_name_to_path`. We should use `class_name_to_path` in `class_name_to_identifier` and then remove the suffix.

lib/riffer/evals/judge.rb
```diff
@@ -0,0 +1,84 @@
...
+    @provider_options = provider_options
+  end
+
+  # Evaluates using the configured LLM.
+  #
+  # system_prompt:: String - the system prompt for the judge
+  # user_prompt:: String - the user prompt containing the evaluation request
+  #
+  # Returns Hash with :score (Float) and :reason (String).
+  def evaluate(system_prompt:, user_prompt:)
```
bottrall: **Suggestion(blocking):** We should support an array of messages as well.

test/riffer/evals/evaluators/answer_relevancy_test.rb
```diff
@@ -0,0 +1,69 @@
...
+    #
+    #       result = evaluator.evaluate(
+    #         input: "What is the capital of France?",
+    #         output: "The capital of France is Paris."
+    #       )
+    #
+    #       expect(result.score).must_be :>=, 0.8
+    #     end
+    #   end
+    # end
```
bottrall: **Issue(blocking):** Use the test provider to test the integration.



Then there's no need for VCR.

test/riffer/evals/evaluators/answer_relevancy_test.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    it "has a description" do
+      expect(Riffer::Evals::Evaluators::AnswerRelevancy.description).wont_be_nil
+    end
+
+    it "has higher_is_better set to true" do
+      expect(Riffer::Evals::Evaluators::AnswerRelevancy.higher_is_better).must_equal true
+    end
+  end
```
bottrall: **Suggestion(blocking):** Don't test things that are static.

test/riffer/evals/evaluators/answer_relevancy_test.rb
```diff
@@ -0,0 +1,69 @@
...
+    # end
+  end
+
+  describe "SYSTEM_PROMPT" do
+    it "includes scoring criteria" do
+      prompt = Riffer::Evals::Evaluators::AnswerRelevancy::SYSTEM_PROMPT
+      expect(prompt).must_include "score"
+      expect(prompt).must_include "JSON"
+    end
+  end
```
bottrall: **Suggestion(blocking):** As above, don't test static properties.

test/riffer/evals/evaluator_test.rb
```diff
@@ -0,0 +1,129 @@
...
+      end
+
+      evaluator = klass.new
+      result = evaluator.evaluate(input: "test", output: "test")
+
+      expect(result).must_be_instance_of Riffer::Evals::Result
+      expect(result.evaluator).must_equal "my_evaluator"
+      expect(result.score).must_equal 0.9
+      expect(result.reason).must_equal "Good"
+      expect(result.higher_is_better).must_equal true
```
bottrall: **Suggestion(non-blocking):** Stick to the single assertion principle.



Instead of testing each individual property, just test the entire object.

lib/riffer/evals/judge.rb
```diff
@@ -0,0 +1,106 @@
...
+  # The model string (provider/model format).
+  #
+  # Returns String.
+  attr_reader :model
+
+  # Initializes a new judge.
+  #
+  # model:: String - the model to use (provider/model format)
+  # provider_options:: Hash - options passed to the provider
+  def initialize(model:, provider_options: {})
```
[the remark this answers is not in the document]
bottrall: Addressed in 3f173f0 — added model string validation in `Judge#initialize` matching the pattern from `Agent`. Invalid model strings now raise `Riffer::ArgumentError` at construction time.

lib/riffer/evals/profile.rb
```diff
@@ -0,0 +1,110 @@
...
+
+      metrics = profile.eval_metrics
+      raise Riffer::ArgumentError, "No metrics configured in eval profile" if metrics.empty?
+
+      # Generate output from agent
+      output = generate(input, **generate_options)
+
+      # Run evaluations
+      runner = Riffer::Evals::Runner.new(metrics: metrics)
+      runner.run(input: input, output: output, context: context)
```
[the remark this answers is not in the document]
bottrall: Addressed in 3f173f0 — replaced `**generate_options` with explicit `tool_context: nil` since that's the only kwarg `Agent#generate` accepts.

lib/riffer/evals/judge.rb
```diff
@@ -0,0 +1,90 @@
...
+  def provider_name
+    @provider_name ||= @model.split("/", 2).first
+  end
+
+  def model_name
+    @model_name ||= @model.split("/", 2).last
+  end
+
+  def parse_response(content)
+    json_match = content.match(/\{[^{}]*"score"[^{}]*\}/m)
```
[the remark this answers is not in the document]
bottrall: This regex no longer exists in the codebase — `Judge` now uses tool calling with `EvaluationTool` to get structured output, so there's no regex-based JSON extraction.

lib/riffer/evals/profile.rb
```diff
@@ -0,0 +1,110 @@
...
+  def self.included(base)
+    base.extend(ClassMethods)
+
+    # When the profile module is included in an Agent, add the eval method
+    base.define_singleton_method(:included) do |target|
+      if target < Riffer::Agent
+        target.extend(AgentClassMethods)
+        target.instance_variable_set(:@eval_profile, base)
+      end
+    end
```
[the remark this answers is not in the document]
bottrall: This is intentional — eval profiles are single-purpose modules that exist solely to be included in agents. They aren't meant to compose with other `included` hooks. The `define_singleton_method(:included)` pattern is the mechanism that wires the profile into the agent when included, and a profile module having its own separate `included` hook alongside this one isn't a supported use case.

lib/riffer/evals/run_result.rb
```diff
@@ -0,0 +1,112 @@
...
+  end
+
+  # Returns results that failed their metric thresholds.
+  #
+  # Returns Array of Riffer::Evals::Result.
+  def failures
+    @failures ||= @results.select.with_index do |result, index|
+      metric = @metrics[index]
+      !metric.passes?(result)
+    end
```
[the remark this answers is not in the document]
bottrall: `RunResult` is an internal value object constructed only by `Runner#run`, which always passes matched-length arrays. Adding defensive validation here would be guarding against a scenario that can't happen through the public API.

lib/riffer/evals/run_result.rb
```diff
@@ -0,0 +1,112 @@
...
+  end
```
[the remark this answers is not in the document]
bottrall: `RunResult` is a short-lived value object — it's constructed, read, and discarded. Nobody holds a reference to mutate the internal arrays after construction. Freezing/duplicating would add complexity for a mutation scenario that doesn't exist in practice.

lib/riffer/evals/judge.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    parsed = JSON.parse(tool_call[:arguments])
+    score = parsed["score"]
+    reason = parsed["reason"]
+
+    raise Riffer::Error, "Invalid judge response: missing score" if score.nil?
+
+    {
+      score: score.to_f,
+      reason: reason
```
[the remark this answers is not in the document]
bottrall: The `EvaluationTool` schema already declares `reason` as a required `String` param, so the provider enforces its presence via tool calling. Adding a runtime nil check here would be redundant with the tool schema contract. Similarly, since `EvaluationTool` is the only tool passed to the provider, verifying the tool name adds no practical safety.

docs/08_EVALS.md
```diff
@@ -0,0 +1,318 @@
...
+
+### Defining a Profile
+
+```ruby
+module QualityEvals
+  include Riffer::Evals::Profile
+
+  ai_evals do
+    metric :answer_relevancy, min: 0.85
+    metric :toxicity, max: 0.10  # For future evaluators
```
cloetzi: question(non-blocking): I am unsure what this comment means.

docs/08_EVALS.md
```diff
@@ -0,0 +1,318 @@
...
+  model "anthropic/claude-haiku-4-5-20251001"
+end
+```
+
+## Running Evals
+
+Once a profile is included, call `.eval` on the agent class:
+
+```ruby
+result = MyAgent.eval(
```
cloetzi: issue(blocking): As discussed, this shadows the `Kernel#eval` method. I'd suggest `run_evals` instead.

docs/08_EVALS.md
```diff
@@ -0,0 +1,318 @@
...
+    You are an evaluation assistant that assesses medical accuracy.
+
+    Use the evaluation tool to submit your score (0.0-1.0) and reasoning.
+  PROMPT
+
+  def evaluate(input:, output:, context: nil)
+    user_prompt = <<~PROMPT
+      Question: #{input}
+      Response: #{output}
+      Ground truth: #{context&.dig(:ground_truth)}
```
cloetzi: nitpick(non-blocking): I know it is doc, but seems kinda odd that ground truth is optional and to potentially send `Ground truth: ` without any content.

docs/08_EVALS.md
```diff
@@ -0,0 +1,318 @@
...
+Riffer::Evals::Evaluators::Repository.register(:medical_accuracy, MedicalAccuracyEvaluator)
+```
+
+### Evaluator DSL
+
+Class methods:
+
+- `identifier(value)` - Set the evaluator identifier (defaults to snake_case class name)
+- `description(value)` - Human-readable description
+- `higher_is_better(value)` - Whether higher scores are better (default: true)
```
cloetzi, line 197: question(non-blocking): For my own understanding, why is this an option and not just a given?

lib/riffer/evals/evaluators/answer_relevancy.rb
```diff
@@ -0,0 +1,57 @@
...
+# on-topic, and directly addresses what was asked.
+#
+#   evaluator = Riffer::Evals::Evaluators::AnswerRelevancy.new
+#   result = evaluator.evaluate(
+#     input: "What is the capital of France?",
+#     output: "The capital of France is Paris."
+#   )
+#   result.score  # => 0.95
+#
+class Riffer::Evals::Evaluators::AnswerRelevancy < Riffer::Evals::Evaluator
```
cloetzi, line 15: question(non-blocking): I am wondering if this should be part of the core repository at all. Let's assume this is picked up by an organization, and we want to update this prompt, is that a breaking change every time? Are you expecting this prompt to be static?

An alternative approach would be that we release a companion gem: `riffer-evaluators` that has its own versioning and is detached from the main package.

It boils down to, what do we want this gem to be?

lib/riffer/evals/result.rb
```diff
@@ -0,0 +1,71 @@
...
+  # Initializes a new evaluation result.
+  #
+  # evaluator:: String - the evaluator identifier
+  # score:: Float - the score (0.0 to 1.0)
+  # reason:: String or nil - optional explanation
+  # metadata:: Hash - optional additional data
+  # higher_is_better:: Boolean - whether higher is better (default: true)
+  def initialize(evaluator:, score:, reason: nil, metadata: {}, higher_is_better: true)
+    @evaluator = evaluator
+    @score = score.to_f
```
cloetzi, line 55: suggestion(non-blocking): Valid comment from LLM code review, there's no validation that the score is between 0 and 1. I think that would make sense.

lib/riffer/evals/run_result.rb
```diff
@@ -0,0 +1,112 @@
...
+  # Returns Boolean.
+  def passed?
+    failures.empty?
+  end
+
+  # Returns results that failed their metric thresholds.
+  #
+  # Returns Array of Riffer::Evals::Result.
+  def failures
+    @failures ||= @results.select.with_index do |result, index|
```
cloetzi: nitpick(non-blocking): You are using the instance variable directly throughout the class, I generally prefer using the reader accessor (`results` instead of `@results`). Especially with `attr_reader` it is a nice signal that we are doing read-only. Not perfect, just a preference.

lib/riffer/evals/runner.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  # Runs all evaluators and collects results.
+  #
+  # input:: String - the input given to the agent
+  # output:: String - the output produced by the agent
+  # context:: Hash or nil - optional context (e.g., ground_truth)
+  #
+  # Returns Riffer::Evals::RunResult.
+  def run(input:, output:, context: nil)
+    results = @metrics.map do |metric|
```
cloetzi: nitpick(non-blocking): Same comment as above about using the `attr_reader` accessor, won't comment on this anymore.

cloetzi (changes requested):
As discussed, we should get rid of shadowing the `eval` method, all other comments are optional.

docs/08_EVALS.md
bottrall: It's just noting that there is no `toxicity` evaluator. But I get how that's confusing. I'll update it so it only includes what's actually available 👍🏻

bottrall: https://github.com/janeapp/riffer/pull/101/commits/a4e374f5676084dc1f5b1feb5b24262fee0d43d4

docs/08_EVALS.md
bottrall: https://github.com/janeapp/riffer/pull/101/commits/4d119e417526b4e5ff0ff9b096451acd7b217a78

docs/08_EVALS.md
bottrall: https://github.com/janeapp/riffer/pull/101/commits/f1ff9c4450c0af38f429074664a37d31e1c7952f

docs/08_EVALS.md
bottrall: One example that comes to mind is if we had a metric for `:hallucination`. It would be unintuitive that a higher score in `:hallucination` actually means it hallucinated less.

lib/riffer/evals/evaluators/answer_relevancy.rb
bottrall: The goal of this gem is laid out in the project description:
> The all-in-one Ruby framework for building AI-powered applications and agents.

To me `all-in-one` encompasses tools necessary to grade your agents. So I think we should include the "standard" evals that nearly every agent is going to want to run.

In regards to whether or not changing this prompt is a breaking change. We should definitely put more thought into that, but we're still a ways off from `1.0.0`. So we can figure that out at a later date 😄

lib/riffer/evals/run_result.rb
bottrall: https://github.com/janeapp/riffer/pull/101/commits/82911066cc46244eca98d5b4c3483afe32513d15

lib/riffer/evals/runner.rb
bottrall: https://github.com/janeapp/riffer/pull/101/commits/82911066cc46244eca98d5b4c3483afe32513d15

lib/riffer/evals/result.rb
bottrall: https://github.com/janeapp/riffer/pull/101/commits/4307b2d34b52d43f4fcc7c9d2c817b8b029d9a06

docs/08_EVALS.md
```diff
@@ -36,7 +36,7 @@ class MyAgent < Riffer::Agent
 end
 
 # 4. Run evals
-result = MyAgent.eval(input: "What is Ruby?")
+result = MyAgent.run_evalinput: "What is Ruby?")
```
cloetzi: issue(blocking): Missing bracket.

docs/08_EVALS.md
cloetzi: Ah, that makes sense, thanks!

cloetzi (commented):
One doc comment needs fixing, otherwise looks good. Thanks for the changes.

docs/08_EVALS.md
bottrall: 🙈 https://github.com/janeapp/riffer/pull/101/commits/453680a6d54624b91a8c53fd63d7d7b66ca3f09c

cloetzi (approved):

lib/riffer/evals/evaluators/answer_relevancy.rb
Thibaut: Beside breaking changes, we will also end up with tight coupling when someone inevitably monkey-patches the built-in `AnswerRelevancy` in order to tune its behavior. 

I think providing an example evaluator like this is a great idea but I would suggest removing the concept of "built-in" evaluators as I don't think this will be practical in the long run.

lib/riffer/evals/judge.rb
```diff
@@ -0,0 +1,111 @@
...
+  # Returns String.
+  attr_reader :model
+
+  # Initializes a new judge.
+  #
+  # model:: String - the model to use (provider/model format)
+  # provider_options:: Hash - options passed to the provider
+  def initialize(model:, provider_options: {})
+    provider_name, model_name = model.split("/", 2)
+    unless [provider_name, model_name].all? { |part| part.is_a?(String) && !part.strip.empty? }
```
Thibaut, line 46: nitpick:


```suggestion
    @model = model
    unless provider_name && model_name
```

to keep the logic in one place and preload the two values.

lib/riffer/evals/profile.rb
```diff
@@ -0,0 +1,110 @@
+# frozen_string_literal: true
+
+# Module factory providing the ai_evals DSL for defining eval profiles.
+#
+# Include this module in a module to create an eval profile that can be
+# included in agents.
+#
+#   module EvalProfiles::QualityEvals
+#     include Riffer::Evals::Profile
```
Thibaut, line 9: suggestion: use subclassing like we do for agent, tools, etc. 


```suggestion
#   module EvalProfiles::QualityEvals < Riffer::Evals::Profile
```

I think this would simplify some of the code below.

docs/08_EVALS.md
```diff
@@ -0,0 +1,316 @@
...
+```ruby
+# 1. Configure the judge model
+Riffer.config.evals.judge_model = "anthropic/claude-opus-4-5-20251101"
+
+# 2. Define an eval profile
+module QualityEvals
+  include Riffer::Evals::Profile
+
+  ai_evals do
+    metric :answer_relevancy, min: 0.85
```
Thibaut, line 27: suggestion: we could avoid the `Riffer::Evals::Evaluators::Repository.register` (which isn't autoload friendly and easy to forget) + I think simplify some code if we changed this API to:


```suggestion
    metric ClassName, min: 0.85
```

I personally like this more as it's easier to trace what code gets called (no need to do a mental dance between `:answer_relevancy` and the class name, which is probably good for coding agents too)

Thibaut (commented):
Late to the party with some suggestions — great stuff! 👏

PR #102, opened 2026-02-03
Add token usage tracking in `Riffer::Agent`

cloetzi:
## Summary

Adds token usage tracking to Riffer so callers can monitor API consumption across all providers.

### New Classes
- **`Riffer::Usage`** - Value object for normalized usage data with `input_tokens`, `output_tokens`, `total_tokens`, and optional cache tokens (`cache_creation_tokens`, `cache_read_tokens`)
- **`Riffer::StreamEvents::UsageDone`** - Stream event emitted when usage data is available

### Agent API
- **`agent.total_usage`** - Cumulative token usage across all LLM calls (including tool loops)
- **`agent.on_usage { |usage| ... }`** - Callback fired after each LLM call

### Provider Support
All providers now extract and report usage:
- **Anthropic** - Includes cache token tracking
- **Amazon Bedrock** - Includes cache token tracking (requires aws-sdk-bedrockruntime ~> 1.42)
- **OpenAI** - Basic token tracking
- **Test** - Supports stubbed usage for testing

### Usage Examples

```ruby
# Access usage from messages
agent = MyAgent.new
agent.generate("Hello!")
agent.messages.last.usage.total_tokens  # => 57

# Cumulative usage across tool loops
agent.total_usage.input_tokens
agent.total_usage.output_tokens

# Real-time callback
agent.on_usage { |usage| log_tokens(usage.total_tokens) }

# Streaming
agent.stream("Hello!").each do |event|
  case event
  when Riffer::StreamEvents::UsageDone
    puts "Used #{event.usage.total_tokens} tokens"
  end
end
```

## Test plan
- [x] Unit tests for `Riffer::Usage` class
- [x] Unit tests for `Riffer::StreamEvents::UsageDone`
- [x] Unit tests for `Riffer::Messages::Assistant` with usage
- [x] Agent tests for `on_usage` callback and `total_usage` tracking
- [x] VCR tests for Anthropic provider (generate + stream)
- [x] VCR tests for Amazon Bedrock provider (generate + stream)
- [x] VCR tests for OpenAI provider (generate + stream)
- [x] Test provider supports usage in stubs

🤖 Generated with [Claude Code](https://claude.ai/code)


docs/06_STREAM_EVENTS.md
```diff
@@ -109,6 +109,22 @@ event.role     # => "assistant"
...
+
+```ruby
+event = Riffer::StreamEvents::UsageDone.new(usage: usage)
+event.role                    # => :assistant
+event.usage                   # => Riffer::Usage
+event.usage.input_tokens      # => 100
+event.usage.output_tokens     # => 50
+event.usage.total_tokens      # => 150
+event.to_h                    # => {role: :assistant, usage: {input_tokens: 100, output_tokens: 50}}
+```
```
bottrall, line 124: **Suggestion(non-blocking):** I'm not sure if we really need to add an event for this.



We attach it to the message so it's always available when using an agent. From what I can tell this only benefits using the provider directly. Which isn't a first class problem for the library.

docs/03_AGENTS.md
```diff
@@ -190,6 +190,35 @@ agent
...
 
+### on_usage
+
+Registers a callback to receive usage data after each LLM call:
+
+```ruby
+agent.on_usage do |usage|
+  puts "Tokens used: #{usage.input_tokens} in, #{usage.output_tokens} out"
+end
+```
```
bottrall: **Suggestion(non-blocking):** I don't think this callback is necessary right now.



riffer doesn't have the concept of callbacks for anything else, except for `on_message`. But that was required, whereas this isn't because it's available via the message. There's probably an argument to be made it should, but lets think through and be intentional about it if we decide that's something the library should do.

lib/riffer/agent.rb
```diff
@@ -122,13 +122,20 @@ def stream(...)
   # Returns Array of Riffer::Messages::Base.
   attr_reader :messages
 
+  # Cumulative token usage across all LLM calls.
+  #
+  # Returns Riffer::Usage or nil.
+  attr_reader :total_usage
```
bottrall: **Suggestion(non-blocking):** `token_usage` may be a bit more self-explanatory here.




```suggestion
  attr_reader :token_usage
```

lib/riffer/usage.rb
```diff
@@ -0,0 +1,85 @@
...
+# Represents token usage data from an LLM API call.
+#
+# Tracks input tokens, output tokens, and optional cache statistics.
+#
+#   usage = Riffer::Usage.new(input_tokens: 100, output_tokens: 50)
+#   usage.total_tokens  # => 150
+#
+#   combined = usage1 + usage2  # Combine multiple usage objects
+#
+class Riffer::Usage
```
bottrall: **Suggestion(non-blocking):** As above, token usage may be a better name in general for everything 😄

bottrall (approved):

docs/03_AGENTS.md
cloetzi: Removed in https://github.com/janeapp/riffer/pull/102/changes/fe228a8b526b12cc95596f43b551eb8d8bf8f9c1.

lib/riffer/agent.rb
cloetzi: Good idea, renamed everything in https://github.com/janeapp/riffer/pull/102/changes/fe228a8b526b12cc95596f43b551eb8d8bf8f9c1.

lib/riffer/usage.rb
cloetzi: 👍 https://github.com/janeapp/riffer/pull/102/changes/fe228a8b526b12cc95596f43b551eb8d8bf8f9c1

docs/06_STREAM_EVENTS.md
cloetzi: I think it might be a nice way of capturing the usage downstream, and given that it is already implemented and not a huge overhead, I would like to keep it.

bottrall (approved):
🤘🏻

docs/06_STREAM_EVENTS.md
Thibaut: I think it might be a cleaner/simpler/more efficient design to consolidate `TextDone` and `TokenUsageDone` as a single event that holds both the completed text and the token usage. "token usage" is just one of potentially many data points we may at the end of the a run. I don't think it would scale very well to add a separate event for each of those (future) data points. 

PS: It could also allow us to optimize the case where the client doesn't care about token usage. We could compute some stuff on-demand vs. always trigger an event that may go unused.

cloetzi: The problem with that approach is that it maps poorly to the Bedrock and OpenAI API. Both APIs have a separate event for text done and token usage ([metadata event for bedrock](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ConverseStreamMetadataEvent.html), [OpenAI](https://platform.openai.com/docs/api-reference/responses-streaming/response/output_text/done) it is `response.output_text.done`, vs `response.completed`. 

We can certainly make it work with a combined event, but it would get more complicated, and we no longer could maximize streaming performance. Also, this is pretty much what `Riffer::Messages::Assistant` is.

If there's a strong preference, I can remove the token event again unless we have a better idea.

Thibaut: I'm good to keep them separate if that's an approach shared by the underlying APIs 👍 Thanks for the context

PR #109, opened 2026-02-11
fix: correct RBS inline annotations and remove ivar declarations

bottrall:
## Summary

- Update RBS inline annotations across all `lib/` files to use proper method signature syntax (`#: (Type) -> ReturnType`) instead of the incorrect per-parameter format
- Remove internal instance variable type declarations (`# @rbs @var` / `# @rbs self.@var`) from 11 files — these aren't part of the public API since consumers interact through `attr_reader` which already has inline type annotations
- Update `.agents/rbs-inline.md` and `.agents/code-style.md` to reflect the corrected annotation style

## Test plan

- [x] `bundle exec rake rbs:generate` — regenerated RBS files successfully
- [x] `bundle exec rake` — 599 tests pass, no Steep type errors

### DIFF
--- a/lib/riffer/evals/evaluators/answer_relevancy.rb
+++ b/lib/riffer/evals/evaluators/answer_relevancy.rb
@@ -38,10 +38,7 @@ class Riffer::Evals::Evaluators::AnswerRelevancy < Riffer::Evals::Evaluator
       - 0.0 = Completely irrelevant
   PROMPT
 
-  #: input: String -- the input that was given to the agent
-  #: output: String -- the output produced by the agent
-  #: context: Hash[Symbol, untyped]? -- optional context
-  #: return: Riffer::Evals::Result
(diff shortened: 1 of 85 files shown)


lib/riffer/evals/evaluators/answer_relevancy.rb
```diff
@@ -50,9 +47,7 @@ def evaluate(input:, output:, context: nil)
 
   private
 
-  #: input: String
-  #: output: String
-  #: return: String
+  #: (input: String, output: String) -> String
```
bottrall, line 50: I'm wondering if we should adopt the policy of we only bother typing public interfaces 🤔
Then again this does help us when working on it...

Thibaut: Assuming it doesn't add much overhead to development (I don't think it does?), it might be easier to "just do it all the time", rather than having to think about what should be typed vs. not.

Thibaut (approved):

PR #112, opened 2026-02-11
feat: remove identifiers from evals and guardrails

bottrall:
## Summary
- Replace string-based identifier lookup with direct class references in both evals and guardrails subsystems
- Remove `Evaluators::Repository` — metrics now store `evaluator_class` (a class reference) directly instead of resolving via `evaluator_identifier` (a string)
- Remove `identifier` class/instance methods from `Guardrail` — `Tripwire` and `Modification` now store the guardrail class directly instead of a string `guardrail_id`
- Update Profile DSL to accept classes instead of symbols: `metric MyEvaluator, min: 0.85` instead of `metric :my_evaluator, min: 0.85`

## Test plan
- [x] `bundle exec rake` (tests + lint + type check) passes — 749 tests, 0 failures
- [x] RBS type signatures regenerated and verified

### DIFF
--- a/lib/riffer/evals/result.rb
+++ b/lib/riffer/evals/result.rb
@@ -51,7 +51,7 @@ class Riffer::Evals::Result
   #: () -> Hash[Symbol, untyped]
   def to_h
     {
-      evaluator: evaluator.name || evaluator.inspect,
+      evaluator: evaluator.name,
       score: score,
       reason: reason,
       metadata: metadata,
--- a/lib/riffer/evals/result.rb
+++ b/lib/riffer/evals/result.rb
@@ -6,19 +6,19 @@
 # Contains the score, reason, and metadata from running an evaluator.
 #
 #   result = Riffer::Evals::Result.new(
-#     evaluator: "answer_relevancy",
+#     evaluator: Riffer::Evals::Evaluators::AnswerRelevancy,
 #     score: 0.85,
 #     reason: "The response addresses the question directly.",
 #     higher_is_better: true
 #   )
 #
 #   result.score           # => 0.85
-#   result.evaluator       # => "answer_relevancy"
+#   result.evaluator       # => Riffer::Evals::Evaluators::AnswerRelevancy
 #   result.higher_is_better # => true
 #
 class Riffer::Evals::Result
-  # The identifier of the evaluator that produced this result.
-  attr_reader :evaluator #: String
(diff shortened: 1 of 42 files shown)


lib/riffer/guardrails/modification.rb
```diff
@@ -33,6 +33,6 @@ def initialize(guardrail_id:, phase:, message_indices:)
   #
   #: () -> Hash[Symbol, untyped]
   def to_h
-    {guardrail_id: guardrail_id, phase: phase, message_indices: message_indices}
+    {guardrail: guardrail.name || guardrail.inspect, phase: phase, message_indices: message_indices}
```
bottrall: Should we use the class name conversion here? 🤔

lib/riffer/guardrails/tripwire.rb
```diff
@@ -52,7 +52,7 @@ def initialize(reason:, guardrail_id:, phase:, metadata: nil)
   def to_h
     {
       reason: reason,
-      guardrail_id: guardrail_id,
+      guardrail: guardrail.name || guardrail.inspect,
```
bottrall: Same deal here. Should we use the classname converter?

bottrall:
@Thibaut thanks for the idea on this one!

lib/riffer/evals/result.rb
```diff
@@ -51,7 +51,7 @@ def initialize(evaluator:, score:, reason: nil, metadata: {}, higher_is_better:
   #: () -> Hash[Symbol, untyped]
   def to_h
     {
-      evaluator: evaluator,
+      evaluator: evaluator.name || evaluator.inspect,
```
Thibaut: suggestion (non-blocking): I would drop the `.inspect` fallback. `evaluator` should always be a class, so `.name` should never fail. If it was anything else, there would likely be a bug at play and `.inspect` could expose data that should remain private.

lib/riffer/guardrails/modification.rb
Thibaut: I guess this relates to my other comment above. Why the `guardrail.inspect` fallback? Isn't `guardrail` always a `Class`?

We could also leave this as `guardrail: guardrail`, which would then call `guardrail.to_s` (which returns `.name` for classes) when the hash is serialized to a string.

Thibaut (approved):
👌 👌  thanks for updating!

PR #119, opened 2026-02-17
feat: interruptible callbacks via throw/catch

cloetzi:
## Summary

- Adds support for interrupting the agent loop from `on_message` callbacks using `throw :riffer_interrupt`
- `Response#interrupted?` returns `true` when the loop was interrupted, with accumulated content preserved
- In streaming, yields a `Riffer::StreamEvents::Interrupt` event
- Adds `resume` and `resume_stream` methods for continuing interrupted loops — supports both in-memory and cross-process resume via persisted messages

## Why `throw`/`catch`?

We considered several approaches for callback-driven flow control:

- **Return values** (e.g., `return :stop` from callback) — would require changing the callback contract from a simple observer to a decision-maker. Every callback would need to return something, and we'd need to define precedence when multiple callbacks disagree.
- **Exceptions** — semantically wrong. Interrupting is expected control flow, not an error condition. Exceptions also carry overhead (backtraces) and can be accidentally caught by rescue clauses in user code or tools.
- **`throw`/`catch`** — Ruby's built-in mechanism for non-local flow control without exception semantics. It's the idiomatic choice for "exit this control flow early," used by Sinatra (`throw :halt`), Rack, and Warden. The `:riffer_interrupt` tag avoids collisions with user code.

`throw`/`catch` keeps callbacks as simple observers that can optionally influence flow, without changing the contract for callbacks that don't need this capability.

## Usage

```ruby
agent = MyAgent.new
agent.on_message do |msg|
  throw :riffer_interrupt if msg.is_a?(Riffer::Messages::Tool)
end

response = agent.generate("Do something")

if response.interrupted?
  # In-memory resume (same process):
  response = agent.resume

  # Or cross-process resume (e.g., after restart or async approval):
  agent = MyAgent.new
  response = agent.resume(messages: persisted_messages, tool_context: {user_id: 123})

  # Streaming resume:
  agent.resume_stream(messages: persisted_messages).each do |event|
    # handle stream events
  end
end
```

## Caveats

If a callback throws during `execute_tool_calls` (after processing some but not all tool calls), the conversation will have partial tool results. Consumers who want to avoid this can check the message type before throwing.

## Test plan

- [x] `throw :riffer_interrupt` in `on_message` during `generate` — returns response with `interrupted?` true
- [x] `throw :riffer_interrupt` during `stream` — yields `Interrupt` event
- [x] Throw during tool execution — verifies partial tool results preserved in messages
- [x] Multiple callbacks where later one throws — earlier callbacks still fire
- [x] `resume` — error when not interrupted, returns Response, preserves messages, completes tool loop
- [x] `resume` with persisted messages — accepts message objects, hashes, and tool_context
- [x] `resume_stream` — error when not interrupted, yields stream events, accepts persisted messages
- [x] `Response#interrupted?` returns false by default, true when interrupted
- [x] `StreamEvents::Interrupt` event with correct `to_h`
- [x] `bundle exec rake` — all tests pass, StandardRB clean, Steep clean

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -44,6 +44,24 @@ Structured events for streaming responses:
 - `TextDone` - completion signals
 - `ReasoningDelta` - reasoning process chunks
 - `ReasoningDone` - reasoning completion
+- `Interrupt` - callback interrupted the agent loop
+
+### Interruptible Callbacks
+
+`on_message` callbacks can interrupt the agent loop using Ruby's `throw`/`catch`. Use `throw :riffer_interrupt` to stop the loop. `Response#interrupted?` returns `true` and accumulated content is preserved. In streaming, yields an `Interrupt` event.
+
+```ruby
+agent = MyAgent.new
+agent.on_message do |msg|
+  throw :riffer_interrupt if msg.is_a?(Riffer::Messages::Assistant)
+end
+response = agent.generate("Hello")
+response.interrupted? # => true
+```
+
+Use `agent.resume` or `agent.resume_stream` to continue an interrupted loop — the agent preserves its full message history. Both accept `messages:` for cross-process resume from persisted data.
+
+If a callback throws during `execute_tool_calls` (after processing some but not all tool calls), the conversation will have partial tool results.
 
 ## Key Patterns
 
@@ -84,6 +102,7 @@ lib/
       test.rb            # Test provider
     stream_events/
       base.rb            # Base stream event
+      interrupt.rb       # Interrupt event
       text_delta.rb      # Text delta event
       text_done.rb       # Text done event
       reasoning_delta.rb # Reasoning delta event
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -46,22 +46,19 @@ Structured events for streaming responses:
 - `ReasoningDone` - reasoning completion
 - `Interrupt` - callback interrupted the agent loop
 
-### Interruptible Callbacks
+### Stopping the Loop Early
 
-`on_message` callbacks can interrupt the agent loop using Ruby's `throw`/`catch`. Use `throw :riffer_interrupt` to stop the loop. `Response#interrupted?` returns `true` and accumulated content is preserved. In streaming, yields an `Interrupt` event.
+Two mechanisms can stop the agent loop before the LLM finishes naturally:
 
-```ruby
-agent = MyAgent.new
-agent.on_message do |msg|
-  throw :riffer_interrupt if msg.is_a?(Riffer::Messages::Assistant)
-end
-response = agent.generate("Hello")
-response.interrupted? # => true
-```
+**Guardrail tripwires** — declarative policy enforcement registered at class level. A `:before` guardrail can block the request before the LLM is called; an `:after` guardrail can block the response. Tripwires are not resumable — the caller must change the input and start over. `Response#blocked?` returns `true`.
 
-Use `agent.resume` or `agent.resume_stream` to continue an interrupted loop — the agent preserves its full message history. Both accept `messages:` for cross-process resume from persisted data.
+**Callback interrupts** — imperative flow control via `on_message` callbacks. Use `throw :riffer_interrupt` to pause the loop at any point. `Response#interrupted?` returns `true`. In streaming, yields an `Interrupt` event.
 
-If a callback throws during `execute_tool_calls` (after processing some but not all tool calls), the conversation will have partial tool results.
+### Resuming After an Interrupt
+
+`agent.resume` or `agent.resume_stream` continues an interrupted loop. Both accept `messages:` for cross-process resume from persisted data.
+
+On resume, `execute_pending_tool_calls` detects tool calls from the last assistant message that lack corresponding tool result messages and executes them before entering the LLM loop. This handles the case where an interrupt fired mid-way through tool execution.
 
 ## Key Patterns
 
--- a/docs/03_AGENTS.md
+++ b/docs/03_AGENTS.md
@@ -250,7 +250,11 @@ agent.stream('Hello').each do |event|
 end
 ```
 
-**Resuming** — use `resume` (or `resume_stream`) to continue an interrupted loop:
+**Partial tool execution** — tool calls are executed one at a time. When an interrupt fires during tool execution, only the completed tool results remain in the message history. For example, if an assistant message requests two tool calls and the callback interrupts after the first tool result, only that first result will be in the message history.
+
+#### Resuming an Interrupted Loop
+
+Use `resume` (or `resume_stream`) to continue after an interrupt. On resume, the agent automatically detects and executes any pending tool calls (tool calls from the last assistant message that lack a corresponding tool result) before re-entering the LLM loop.
 
 ```ruby
 agent = MyAgent.new
@@ -260,7 +264,7 @@ response = agent.generate('Do something risky')
 
 if response.interrupted?
   approve_action(agent.messages)
-  response = agent.resume
+  response = agent.resume   # executes pending tools, then calls the LLM
 end
 ```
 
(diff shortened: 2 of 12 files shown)


docs/03_AGENTS.md
```diff
@@ -213,6 +214,113 @@ agent
...
+end
+```
+
+For cross-process resume (e.g., after a process restart or async approval), pass persisted messages via the `messages:` keyword. Accepts both message objects and hashes:
+
+```ruby
+# Persist messages during generation (e.g., via on_message callback)
+# Later, in a new process:
+agent = MyAgent.new
+response = agent.resume(messages: persisted_messages, tool_context: {user_id: 123})
```
cloetzi, line 277: I am on the fence if this is the right abstraction if the resume happens in a brand-new instance of the `agent`. I think `resume` without args above makes sense, and it is clear what is happening, but I do wonder if it is a brand-new instance of an agent, we just use generate with all the messages?

The main downside of that is that we might have duplicate system messages if the downstream implementation is just passing the previous message state back. We could guard against that by rejecting system messages and raise if they are passed into generate. 

Thoughts?

.agents/architecture.md
```diff
@@ -44,6 +44,24 @@ Structured events for streaming responses:
...
+agent.on_message do |msg|
+  throw :riffer_interrupt if msg.is_a?(Riffer::Messages::Assistant)
+end
+response = agent.generate("Hello")
+response.interrupted? # => true
+```
+
+Use `agent.resume` or `agent.resume_stream` to continue an interrupted loop — the agent preserves its full message history. Both accept `messages:` for cross-process resume from persisted data.
+
+If a callback throws during `execute_tool_calls` (after processing some but not all tool calls), the conversation will have partial tool results.
```
bottrall: **Issue(blocking):** For some providers this will create an invalid message state.



For instance if you miss a tool call result in the response for bedrock, it will raise. Ideally we don't let the messages get into an invalid state. We'll need to add logic to the agent loop so that it catches up on tool calls responses its missing.

docs/03_AGENTS.md
bottrall: I think this is fine 👍🏻

lib/riffer/agent.rb
```diff
@@ -220,6 +206,129 @@ def stream(prompt_or_messages, tool_context: nil)
...
+  #: (?messages: Array[Hash[Symbol, untyped] | Riffer::Messages::Base]?, ?tool_context: Hash[Symbol, untyped]?) -> void
+  def restore_state(messages: nil, tool_context: nil)
+    if messages
+      @tool_context = tool_context
+      @resolved_tools = nil
+      @messages = messages.map { |item| convert_to_message_object(item) }
+    else
+      raise Riffer::ArgumentError, "Cannot resume: tool_context requires messages" if tool_context
+      raise Riffer::ArgumentError, "Cannot resume: agent was not interrupted" unless @interrupted
+    end
```
bottrall: **Suggestion(blocking):** I think this might create some non-intuitive behavior.



In reality there's only a few things we care about when resuming:
1. if messages are provided they are set.
2. if tool_context is provided it is set.
3. interrupted is reset.
4. resolved tools is reset

If the messages array is invalid, it will raise at the provider level.

lib/riffer/agent/response.rb
```diff
@@ -48,4 +50,12 @@ def blocked?
...
     modifications.any?
   end
+
+  # Returns true if the agent loop was interrupted by a callback
+  # via +throw :riffer_interrupt+.
+  #
+  #: () -> bool
+  def interrupted?
+    @interrupted
+  end
```
bottrall, line 65: **Thought(non-blocking):** I'm wondering if we need to attach a reason here somehow or have different types of interrupts? 🤔 



Or does it not matter?

bottrall (changes requested):
Thanks mate!

.agents/architecture.md
```diff
@@ -44,6 +44,24 @@ Structured events for streaming responses:
...
+- `Interrupt` - callback interrupted the agent loop
+
+### Interruptible Callbacks
+
+`on_message` callbacks can interrupt the agent loop using Ruby's `throw`/`catch`. Use `throw :riffer_interrupt` to stop the loop. `Response#interrupted?` returns `true` and accumulated content is preserved. In streaming, yields an `Interrupt` event.
+
+```ruby
+agent = MyAgent.new
+agent.on_message do |msg|
+  throw :riffer_interrupt if msg.is_a?(Riffer::Messages::Assistant)
```
Thibaut: suggestion (non-blocking): not specifically here but generally I think we should document better through the examples the difference between guardrails (blocked) and throws (interrupted). The two concepts are very similar.

docs/03_AGENTS.md
```diff
@@ -213,6 +214,113 @@ agent
...
+#### Interrupting the Agent Loop
+
+Callbacks can interrupt the agent loop using Ruby's `throw`/`catch` pattern. This is useful for human-in-the-loop approval, cost limits, or content filtering.
+
+Use `throw :riffer_interrupt` to stop the loop. The response will have `interrupted?` set to `true` and contain the accumulated content up to the point of interruption:
+
+```ruby
+agent = MyAgent.new
+agent.on_message do |msg|
+  throw :riffer_interrupt if msg.is_a?(Riffer::Messages::Tool)
```
Thibaut: suggestion (non-blocking): drop the `riffer_` prefix? It doesn't read as well and in this context it's evident what is being interrupted.

docs/03_AGENTS.md
```diff
@@ -213,6 +214,113 @@ agent
...
+
+response = agent.generate('Do something risky')
+
+if response.interrupted?
+  approve_action(agent.messages)
+  response = agent.resume
+end
+```
+
+For cross-process resume (e.g., after a process restart or async approval), pass persisted messages via the `messages:` keyword. Accepts both message objects and hashes:
```
Thibaut: question (non-blocking): I don't understand this use-case. Do you have an example? 😅

docs/03_AGENTS.md
```diff
@@ -213,6 +214,113 @@ agent
...
+
+```ruby
+agent = MyAgent.new
+response = agent.resume(messages: persisted_messages, tool_context: {user_id: 123})
+```
+
+When called without `messages:`, raises `Riffer::ArgumentError` if the agent was not previously interrupted. When called with `messages:`, no prior interruption is required.
```
Thibaut: suggestion (non-blocking): I think this is very non-obvious and a smell that it needs to be documented. Could there be two different methods for each of the use-cases? (with a simple required argument vs. not)

docs/03_AGENTS.md
cloetzi: It will give us more flexibility for agents that require many tool calls/back-and-forth (and is also a concept that might become relevant for a "workflow" concept). 

To be able to run an agent reliably, we have to cap its runtime to something reasonable (ideally <30s), if it takes longer and blocks a process/thread/fiber it can become an operational liability because it blocks deploys and doesn't allow rebalancing. As such, we want to be able to interrupt an agent run, and continue somewhere else.

Example:
1. Agent starts and sends LLM request (5s).
2. Requests slow tool usage (10s)
3a. Processes response and requests more tool usage(10s)
3b. SIGINT is received to shut down the processing worker.
5. Agent job is re-queued and picked up by another process.
6. Tool is executed (20s).
7. Final LLM message (10s).

Other potential examples:
- A circuit breaker is tripping, so we are throttling agent executions (especially tool calls).
- User requests abort, we might want to allow continuation?
- Low-priority/scheduled agent calls that should be deprioritized if we get queue pressure.

tl;dr: Having the option to move an agent flow from one process to another will give us a ton of operational flexibility and avoid having to put hard time caps on full agent runs.

docs/03_AGENTS.md
cloetzi: This was my initial design, but I decided against it because of the risk of cross-interference. While the risk is low, `throws` are global, so if another library is used within the callback it could interfere with that. The prefix is ugly, but will make this throw unique.

I could rename it to `throw :interrupt_riffer`?

.agents/architecture.md
cloetzi: Addressed in 9041bd01a3f5e59bd9bcadc4425adba641fc784a.

lib/riffer/agent.rb
cloetzi: Addressed in 9041bd01a3f5e59bd9bcadc4425adba641fc784a.

lib/riffer/agent/response.rb
cloetzi: Addressed in ed1534cc0206af469a3609acd9aeaaaf5cd6e43e.

.agents/architecture.md
cloetzi: Addressed in a3e9c826aef0536e43fbf8c50ab2aa636313d280.

docs/03_AGENTS.md
cloetzi: Addressed in 9041bd01a3f5e59bd9bcadc4425adba641fc784a

docs/03_AGENTS.md
```diff
@@ -273,3 +373,63 @@ Tool execution errors are captured and sent back to the LLM:
 - `execution_error` - Tool raised an exception
 
 The LLM can use this information to retry or respond appropriately.
+
+## Ways the Agent Loop Can Stop
```
Thibaut, line 377: 👌

Thibaut (approved):

bottrall (approved):
🤘🏻

PR #120, opened 2026-02-18
refactor: restructure providers with hook-based template method

bottrall:
## Summary

- Extract `generate_text` / `stream_text` orchestration into `Riffer::Providers::Base` using five hook methods: `build_request_params`, `execute_generate`, `execute_stream`, `extract_token_usage`, `extract_assistant_message`
- Reorder methods in all four providers (Bedrock, Anthropic, OpenAI, Test) to follow a consistent three-section layout matching call sequence: hooks (generate flow then stream flow), handlers (stream dispatcher order), then helpers (first-call order)
- Update `docs_providers/06_CUSTOM_PROVIDERS.md` and `.agents/providers.md` to reflect the new hook-based API
- Normalize all `handle_*` method signatures to uniform `(event, state:, yielder:)` — both kwargs mandatory, consistent order across all providers
- Remove extra `stream:` parameter from Anthropic's `handle_message_stop` (stream stored in state instead)
- Reset `state[:reasoning]` in Anthropic's `handle_content_block_stop_thinking` after emitting the event
- Move Anthropic `TextDone` emission from `handle_message_stop` to `handle_content_block_stop_text`, aligning with Bedrock's pattern of emitting at content block stop rather than unconditionally at message stop
- Fix 4 Bedrock handler RBS annotations to use keyword `state:` parameter instead of positional

## Test plan

- [x] `bundle exec rake` — 784 tests pass, lint clean
- [x] `bundle exec rake rbs:generate` — RBS signatures regenerated and type-checked
- [x] Review diff to confirm method bodies are unchanged (pure reorder + base extraction)

### DIFF
--- a/docs_providers/06_CUSTOM_PROVIDERS.md
+++ b/docs_providers/06_CUSTOM_PROVIDERS.md
@@ -4,7 +4,7 @@ You can create custom providers to connect Riffer to other LLM services.
 
 ## Basic Structure
 
-Extend `Riffer::Providers::Base` and implement the required methods:
+Extend `Riffer::Providers::Base` and implement the five required hook methods:
 
 ```ruby
 class Riffer::Providers::MyProvider < Riffer::Providers::Base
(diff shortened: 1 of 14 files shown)


docs_providers/06_CUSTOM_PROVIDERS.md
```diff
@@ -217,43 +252,100 @@ end
 ## Complete Example
```
cloetzi, line 252: thought(non-blocking): Is it really helpful to maintain this in the doc, or should we just point to the provider code folder? I have now found a couple of times, that I missed updating the example section and given that it is not tested, we don't really know if it is actually correct. However, our provider implementations are tested.

lib/riffer/providers/amazon_bedrock.rb
```diff
@@ -50,95 +48,143 @@ def perform_generate_text(messages, model:, **options)
...
+          handle_content_block_stop_tool_use(event, state: current_state, yielder: yielder) if current_state[:tool_call]
+        when :metadata
+          handle_metadata_usage(event, state: current_state, yielder: yielder) if event.usage
         end
       end
     end
   end
 
+  #: (Aws::BedrockRuntime::Types::ContentBlockStartEvent, Hash[Symbol, untyped], yielder: Enumerator[Riffer::StreamEvents::Base, void]) -> void
+  def handle_content_block_start_tool_use(event, state:, yielder: nil)
```
cloetzi: question(non-blocking): Is the `nil` here intentional? All other handle methods have it mandatory.

lib/riffer/providers/amazon_bedrock.rb
```diff
@@ -50,95 +48,143 @@ def perform_generate_text(messages, model:, **options)
...
+      item_id: tool_call[:id],
+      call_id: tool_call[:id],
+      name: tool_call[:name],
+      arguments: tool_call[:arguments]
+    )
+    state[:tool_call] = nil
+  end
+
+  #: (Aws::BedrockRuntime::Types::ConverseStreamMetadataEvent, Hash[Symbol, untyped], yielder: Enumerator[Riffer::StreamEvents::Base, void]) -> void
+  def handle_metadata_usage(event, yielder:, state: nil)
```
cloetzi: question(non-blocking): Same question here but for `state`, why is it `nil`?

Also nit: could we keep the order of state and yielder consistent for all methods in this class?

lib/riffer/providers/open_ai.rb
```diff
@@ -135,67 +84,134 @@ def extract_assistant_message(output_items, token_usage = nil)
...
-      )
-    when :"response.function_call_arguments.done"
-      tracked = tool_info[event.item_id] || {}
-      Riffer::StreamEvents::ToolCallDone.new(
-        item_id: event.item_id,
-        call_id: tracked[:call_id] || event.item_id,
-        name: tracked[:name],
-        arguments: event.arguments
+  #: (untyped, yielder: Enumerator::Yielder, ?state: Hash[Symbol, untyped]?) -> void
+  def handle_output_text_delta(event, yielder:, state: nil)
```
cloetzi: suggestion(non-blocking): Same as with bedrock, the arguments of the different handle methods are varying (optional vs mandatory, order). It would be nice to keep this consistent.

lib/riffer/providers/anthropic.rb
```diff
@@ -42,97 +40,147 @@ def perform_generate_text(messages, model:, **options)
...
+      item_id: content_block.id,
+      call_id: content_block.id,
+      name: content_block.name,
+      arguments: arguments
+    )
+    state[:tool_call] = nil
+  end
+
+  #: (untyped, state: Hash[Symbol, untyped], yielder: Enumerator::Yielder) -> void
+  def handle_content_block_stop_thinking(_event, state:, yielder:)
```
cloetzi, line 165: question(non-blocking): Claude pointed this out, but does this have to reset the `state[:reasoning]` attribute?

lib/riffer/providers/anthropic.rb
```diff
@@ -42,97 +40,147 @@ def perform_generate_text(messages, model:, **options)
...
+    yielder << Riffer::StreamEvents::ReasoningDone.new(state[:reasoning])
+  end
+
+  #: (untyped, state: Hash[Symbol, untyped], yielder: Enumerator::Yielder) -> void
```
cloetzi, line 176: issue(non-blocking): The `stream` attribute is missing.

cloetzi (approved):
Nothing strictly blocking, but there are quite a few tidy-ups that would be nice.

docs_providers/06_CUSTOM_PROVIDERS.md
bottrall: I think there's value here if the path we're taking is that we won't create/maintain every provider available. Maybe we can guide agents to ensure they keep this up to date, whenever doing work on the base provider 🤔

lib/riffer/providers/amazon_bedrock.rb
bottrall: Something from a previous iteration, wil cleanup 👍🏻

lib/riffer/providers/amazon_bedrock.rb
bottrall: As above.

lib/riffer/providers/open_ai.rb
bottrall: As above.

lib/riffer/providers/anthropic.rb
bottrall: Yep good call 👍🏻

PR #121, opened 2026-02-18
feat: add max_steps option to Agent

bottrall:
## Summary
- Add `max_steps` class-level DSL method to `Riffer::Agent` for capping the tool-use loop
- Default `DEFAULT_MAX_STEPS = 16` prevents runaway agent loops; set `max_steps Float::INFINITY` for unlimited
- Unify max_steps with the existing interrupt mechanism — when the limit is reached, `throw :riffer_interrupt, :max_steps` fires instead of a separate flag/event
- `interrupted?` returns true and `interrupt_reason` returns `:max_steps` on the response (no separate `max_steps_reached?` predicate)
- In streaming, an `Interrupt` event with `reason: :max_steps` is emitted (no separate `MaxStepsReached` event)
- Add `INTERRUPT_MAX_STEPS = :max_steps` constant for documented reference
- Widen `interrupt_reason` / `reason` type from `String?` to `(String | Symbol)?` — strings for user-provided reasons, symbols for framework-generated ones
- Resume works naturally — `throw` happens after `add_message` but before `execute_tool_calls`, so `execute_pending_tool_calls` picks up unexecuted calls
- Document `max_steps` in the Agents guide (configuration section, "Ways the Agent Loop Can Stop", comparison table) and Stream Events guide (`Interrupt` event)

## Test plan
- [x] `.max_steps` DSL returns `DEFAULT_MAX_STEPS` (16) when not set, stores value when set
- [x] `#generate` runs unlimited steps with `Float::INFINITY`
- [x] `#generate` limits LLM calls when `max_steps` is set
- [x] `#generate` returns last assistant content when limit is reached
- [x] `#generate` sets `interrupted?` to true and `interrupt_reason` to `:max_steps` when limit is reached
- [x] `#stream` limits LLM calls when `max_steps` is set
- [x] `#stream` emits `Interrupt` event with `reason: :max_steps` when limit is reached
- [x] `#stream` does not emit `Interrupt` event when limit is not reached
- [x] `StreamEvents::Interrupt` accepts symbol reason, includes it in `to_h`
- [x] All 796 tests pass, lint clean, type checks pass (`bundle exec rake`)


lib/riffer/agent.rb
```diff
@@ -171,12 +179,14 @@ def initialize
 
   # Generates a response from the agent.
   #
-  #: ((String | Array[Hash[Symbol, untyped] | Riffer::Messages::Base]), ?tool_context: Hash[Symbol, untyped]?) -> Riffer::Agent::Response
-  def generate(prompt_or_messages, tool_context: nil)
+  #: ((String | Array[Hash[Symbol, untyped] | Riffer::Messages::Base]), ?tool_context: Hash[Symbol, untyped]?, ?max_steps: Integer?) -> Riffer::Agent::Response
+  def generate(prompt_or_messages, tool_context: nil, max_steps: nil)
```
bottrall: **suggestion(blocking):** remove max_steps from `generate` and `stream`. Solely use the class variable.

lib/riffer/agent.rb
```diff
@@ -196,6 +207,7 @@ def generate(prompt_or_messages, tool_context: nil)
       add_message(processed_response)
 
       break unless has_tool_calls?(processed_response)
+      break if resolved_max_steps && step >= resolved_max_steps
```
bottrall: **Issue(blocking):** We need to indicate that the stop reason was due to reaching the max steps.



We will need to include this in the response object + emit an event for streaming.

lib/riffer/agent/response.rb
```diff
@@ -32,14 +35,16 @@ class Riffer::Agent::Response
...
-  #: (String, ?tripwire: Riffer::Guardrails::Tripwire?, ?modifications: Array[Riffer::Guardrails::Modification], ?interrupted: bool, ?interrupt_reason: String?) -> void
-  def initialize(content, tripwire: nil, modifications: [], interrupted: false, interrupt_reason: nil)
+  #: (String, ?tripwire: Riffer::Guardrails::Tripwire?, ?modifications: Array[Riffer::Guardrails::Modification], ?interrupted: bool, ?interrupt_reason: String?, ?max_steps_reached: bool) -> void
+  def initialize(content, tripwire: nil, modifications: [], interrupted: false, interrupt_reason: nil, max_steps_reached: false)
     @content = content
     @tripwire = tripwire
     @modifications = modifications
     @interrupted = interrupted
     @interrupt_reason = interrupt_reason
+    @max_steps_reached = max_steps_reached
```
cloetzi: thought(non-blocking): Instead of introducing a new flag, could we use `interrupted` and the reason? Why would a consumer of the response need to know on top level if it was interrupted by `max_steps` and could not check the reason? We haven't released the interrupt change yet, so we could massage the reason field a bit into a more structured data.

lib/riffer/stream_events/max_steps_reached.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+# rbs_inline: enabled
+
+# Represents a max steps reached event during streaming.
+#
+# Emitted when the agent loop stops because the max_steps limit was reached.
+class Riffer::StreamEvents::MaxStepsReached < Riffer::StreamEvents::Base
```
cloetzi: thought(non-blocking): Same as above applies, I think we should combine this with interrupt but make the reason response more strict.

lib/riffer/agent.rb
```diff
@@ -269,10 +286,15 @@ def run_generate_loop(all_modifications = [], resume: false)
 
         break unless has_tool_calls?(processed_response)
 
+        if step >= self.class.max_steps
+          max_steps_reached = true
+          break
```
cloetzi: thought(non-blocking): Just continuing my thoughts from above, this could then be moved to `throw` and no extra handling required below.

cloetzi (commented):
I can approve as is, but I think it is worth exploring using the interrupt path a bit more.

cloetzi (approved):
Nice, thank you!

PR #126, opened 2026-02-19
feat: add web_search option for OpenAI and Anthropic providers

bottrall:
## Summary
- Add `web_search` model option that enables server-side web search for OpenAI (`web_search_preview`) and Anthropic (`web_search_20250305`) providers
- Introduce `WebSearchStatus` (status notifications) and `WebSearchDone` (search results with query and sources) stream events
- `WebSearchStatus` carries optional `url` (for OpenAI `OpenPage` actions) and `query` (for Anthropic searching status) attributes
- `WebSearchDone` carries `query` and `sources` array containing `{title:, url:}` hashes
- Both providers extract sources into `WebSearchDone`: Anthropic from `web_search_tool_result` content blocks (with title + url), OpenAI from `Search` action sources (url only, title is nil)
- OpenAI: dispatches on action type — `Search` emits `WebSearchDone` with sources, `OpenPage` emits `WebSearchStatus` with `open_page` status and URL
- Anthropic: extracts sources from `web_search_tool_result` content block, passes query through to `searching` status
- Test provider emits enriched events with query, url, and sources
- `WebSearchStatus#to_h` omits `url` and `query` when nil (compact hash)
- Both `true` and Hash (with provider-specific options like `search_context_size`, `allowed_domains`, etc.) are supported
- Extract `WEB_SEARCH_TOOL_TYPE` constants in both providers for clarity
- Document `web_search` across all docs and `.agents/` guides: stream events, provider guides (OpenAI, Anthropic, Test, Custom), configuration tables, overview, agents, and architecture

## Test plan
- [x] `WebSearchStatus` unit tests (status, url, query attributes, to_h with/without optional fields)
- [x] `WebSearchDone` unit tests (query, sources attributes, to_h includes sources)
- [x] OpenAI: `generate_text` with `web_search: true` returns Assistant message
- [x] OpenAI: `generate_text` with `web_search` hash options (VCR-backed integration test)
- [x] OpenAI: `stream_text` yields `WebSearchStatus` and `WebSearchDone` events
- [x] OpenAI: `WebSearchDone` includes non-empty query
- [x] OpenAI: `WebSearchDone` includes sources array
- [x] OpenAI: web search events come before text events
- [x] Anthropic: `generate_text` with `web_search: true` returns Assistant message
- [x] Anthropic: `generate_text` with `web_search` hash options (VCR-backed integration test)
- [x] Anthropic: `stream_text` yields `WebSearchStatus` and `WebSearchDone` events
- [x] Anthropic: `WebSearchDone` includes non-empty sources with title and url
- [x] Anthropic: `WebSearchStatus` searching event includes query
- [x] Anthropic: no phantom `ToolCallDelta` events from server_tool_use blocks
- [x] All 830 tests pass, StandardRB lint clean, Steep type check passes


docs/01_OVERVIEW.md
```diff
@@ -66,6 +66,8 @@ When streaming responses, Riffer emits typed events:
 - `TextDone` - Complete text
 - `ToolCallDelta` - Incremental tool call arguments
 - `ToolCallDone` - Complete tool call
+- `WebSearchStatus` - Web search progress updates
+- `WebSearchResult` - Web search completion with sources
```
Thibaut: question (non-blocking): would it make sense to align the naming with text and tool calls: `WebSearchDelta` / `WebSearchDone`? It isn't a "delta" per se but it's still some kind of a progress update (which a delta also is)

Thibaut: thought (non-blocking): I don't quite like how we're making a "special case" out of web search but I don't have a better idea. With more "special cases" this isn't going to scale well, but we can revisit if/when we get there.

docs_providers/01_PROVIDERS.md
```diff
@@ -63,14 +63,14 @@ end
...
-| `tools`     | Array of Tool classes                                     |
-| `**options` | Provider-specific options                                 |
+| Parameter   | Description                                                                 |
+| ----------- | --------------------------------------------------------------------------- |
+| `prompt`    | String prompt (required if `messages` not provided)                         |
+| `system`    | Optional system message string                                              |
+| `messages`  | Array of message objects/hashes (alternative to `prompt`)                   |
+| `model`     | Model name string                                                           |
+| `tools`     | Array of Tool classes                                                       |
+| `**options` | Provider-specific options (including `web_search` for OpenAI and Anthropic) |
```
Thibaut:
```suggestion
| `**options` | Provider-specific options (including `web_search` if supported by provider) |
```

If we want to document which features are supported by which provider, I would do it separately, and in "one place" that's easier to keep up-to-date than sprinkled throughout the docs.

Edit: `doc_providers/*.md` below already document that so I'd just make the suggested edit

lib/riffer/providers/anthropic.rb
```diff
@@ -24,22 +24,30 @@ def initialize(api_key: nil, **options)
...
 
     params[:system] = partitioned_messages[:system] if partitioned_messages[:system]
 
-    if tools && !tools.empty?
-      params[:tools] = tools.map { |t| convert_tool_to_anthropic_format(t) }
+    anthropic_tools = []
+    anthropic_tools.concat(tools.map { |t| convert_tool_to_anthropic_format(t) }) if tools && !tools.empty?
+
+    if web_search
+      web_search_tool = {type: "web_search_20250305", name: "web_search"}
```
Thibaut: suggestion (non-blocking): make `web_search_20250305` a constant or config?

lib/riffer/providers/open_ai.rb
```diff
@@ -29,13 +30,20 @@ def build_request_params(messages, model, options)
...
-      **options.except(:reasoning, :tools)
+      **options.except(:reasoning, :tools, :web_search)
     }
 
-    if tools && !tools.empty?
-      params[:tools] = tools.map { |t| convert_tool_to_openai_format(t) }
+    openai_tools = []
+    openai_tools.concat(tools.map { |t| convert_tool_to_openai_format(t) }) if tools && !tools.empty?
+
+    if web_search
```
Thibaut, line 42: suggestion (non-blocking): there is a bunch of duplication across these `#build_request_params` that potentially could be refactored in `Riffer::Providers::Base` to keep the provider subclasses more lean. I'm not attached to it but could be worth asking Opus to refactor and reviewing what comes out.

test/riffer/providers/anthropic_test.rb
```diff
@@ -156,6 +156,68 @@
     end
   end
 
+  describe "web search" do
```
Thibaut, line 290: suggestion (blocking-but-approve-anyway): let's test at least one of the options (like `max_uses` or `search_context_size` for OpenAI) to stress-test that code path / a non `true` value for `web_search:`. As it stands it's documented but untested.

Thibaut (approved):
🚀

docs/01_OVERVIEW.md
bottrall: That's how I had it in an earlier revision. But `Delta` doesn't make sense because we're not getting any partial content. We could rename `WebSearchResult` to `WebSearchDone` for more consistency. But Delta doesn't make sense IMO.

lib/riffer/providers/open_ai.rb
bottrall: Opus' response to you:
> For the build_request_params refactoring, the providers have very different param structures (Anthropic has max_tokens/system, OpenAI has reasoning, Bedrock has no web_search at all). The only shared web_search logic is ~4 lines.

😅so the robot disagrees.

PR #127, opened 2026-02-19
feat: support dynamic model selection via lambda

bottrall:
## Summary

- Allow the `model` DSL on `Riffer::Agent` to accept a `Proc`/lambda in addition to a static string
- The lambda is re-evaluated on each `generate`/`stream` call, enabling runtime model selection based on `tool_context`
- Supports arity-0 lambdas (no args) and arity-1 lambdas (receives `tool_context`)
- Follows the same pattern already established by `uses_tools` lambda support
- Updates docs (`03_AGENTS.md`) and architecture guide (`.agents/architecture.md`)
- Rename `resolved_model` method to `resolve_model` with `attr_reader` for the cached value
- Extract `clear_resolved_model` helper to DRY up repeated reset logic
- Remove unnecessary local variable indirection in `resolve_model`

## Test plan

- [x] All 809 existing tests pass (`bundle exec rake test`)
- [x] No lint violations (`bundle exec rake standard`)
- [x] Type checking passes (`bundle exec steep check`)
- [x] New tests cover: lambda resolution for generate/stream, tool_context passing (arity 0 & 1), provider lookup, re-evaluation across calls, model switching between calls, model name forwarding to provider, invalid returns (no slash, nil), static string regression

### DIFF
--- a/lib/riffer/agent.rb
+++ b/lib/riffer/agent.rb
@@ -34,13 +34,18 @@ class Riffer::Agent
     @identifier = value.to_s
   end
 
-  # Gets or sets the model string (e.g., "openai/gpt-4o").
+  # Gets or sets the model string (e.g., "openai/gpt-4o") or Proc.
   #
-  #: (?String?) -> String?
-  def self.model(model_string = nil)
-    return @model if model_string.nil?
-    validate_is_string!(model_string, "model")
-    @model = model_string
+  #: (?(String | Proc)?) -> (String | Proc)?
+  def self.model(model_string_or_proc = nil)
+    return @model if model_string_or_proc.nil?
+
+    if model_string_or_proc.is_a?(Proc)
+      @model = model_string_or_proc
+    else
+      validate_is_string!(model_string_or_proc, "model")
+      @model = model_string_or_proc
+    end
   end
 
   # Gets or sets the agent instructions.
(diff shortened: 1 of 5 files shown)


lib/riffer/agent.rb
```diff
@@ -518,6 +531,29 @@ def execute_tool_call(tool_call)
...
+    provider_name, model_name = model_string.split("/", 2)
+    raise Riffer::ArgumentError, "Invalid model string: #{model_string}" unless [provider_name, model_name].all? { |part| part.is_a?(String) && !part.strip.empty? }
+    @provider_name = provider_name
+    @model_name = model_name
+  end
+
+  #: () -> String
+  def resolved_model
+    @resolved_model ||= begin
+      config = @model_config
```
Thibaut: Nit: I'm not a fan of these kinds of variable reassignment. It adds indirection to the code that can make it harder to understand. Simpler to use `@model_config` directly.

lib/riffer/agent.rb
```diff
@@ -409,6 +420,7 @@ def run_stream_loop(yielder, resume: false)
 
   #: () -> Riffer::Messages::Assistant
   def call_llm
+    resolved_model
```
Thibaut: nit: rename this `resolve_model` and then have an `attr_reader` for `@resolved_model`. More self-revealing as to why we're calling this method here.

lib/riffer/agent.rb
```diff
@@ -190,6 +195,8 @@ def initialize
   def generate(prompt_or_messages, tool_context: nil)
     @tool_context = tool_context
     @resolved_tools = nil
+    @resolved_model = nil
+    @provider_instance = nil if @model_config.is_a?(Proc)
```
Thibaut: suggestion (non-blocking): since we extracted the "resolving of the model" in a method, I think extracting these two lines under `clear_resolved_model` would be a good counterpart, to make it easy to read/understand the code.

Thibaut (approved):
nice!

PR #129, opened 2026-02-19
feat: add file attachment support for user messages

bottrall:
## Summary

- Add `Riffer::FilePart` abstraction for file attachments (images and documents) supporting two input sources: URLs and raw base64 data
- Supported image types: JPEG, PNG, GIF, WebP; document types: PDF, TXT, CSV, HTML
- Add `files:` parameter to `Riffer::Messages::User`, `Riffer::Messages::Converter`, `Riffer::Providers::Base`, `Riffer::Agent#generate`, and `Riffer::Agent#stream`
- Implement provider-specific file content block conversion for all three providers:
  - **OpenAI**: `input_image` (URL or data URI) and `input_file` (data URI)
  - **Anthropic**: `image`/`document` blocks with `url` or `base64` source
  - **Bedrock**: `image`/`document` blocks with `bytes` source; raises if URL-only (no lazy fetching)
- Fix Bedrock provider double base64 encoding bug — `file.data` (already base64) was passed directly to AWS SDK `bytes` field which re-encodes; now decodes to raw bytes first
- Fully backwards compatible — `files` defaults to `[]` everywhere
- Update documentation (agents, messages, architecture, providers)

### PR review feedback addressed

- **Remove `from_path` support** — disk-reading removed to keep the door open for remote/distributed execution
- **Remove URL fetching complexity** — `Net::HTTP` fetching removed; URLs are a pass-through for providers that support them natively; `data` returns `nil` for URL-only sources
- **Raise error when `files:` and `messages:` both provided** — previously files were silently dropped; now raises `Riffer::ArgumentError`
- **Bedrock document `name` always required** — `name` field is now always set (AWS requires it)
- **Add VCR cassettes for document uploads** — PDF upload tests added for all three providers (generate + stream)

## Test plan

- `bundle exec rake` passes (976 tests, 1290 assertions, 0 failures)
- Unit tests for `Riffer::FilePart` (construction from url/data, media type detection, predicates, serialization)
- Tests for `Messages::User` with files, `Messages::Converter` file handling, and hash-to-FilePart conversion
- VCR integration tests for file attachments across all three providers (`generate_text` and `stream_text` with image and document)
- Tests for `Agent#generate` and `Agent#stream` with `files:` parameter
- Test that `files:` + `messages:` raises `ArgumentError`
- StandardRB lint clean, Steep type checking clean


docs/03_AGENTS.md
```diff
@@ -200,6 +200,18 @@ response = MyAgent.generate([
 
 # With tool context
 response = MyAgent.generate('Look up my orders', tool_context: {user_id: 123})
+
+# With files (string prompt + files shorthand)
+response = MyAgent.generate('What is in this image?', files: [
+  {path: 'photo.jpg'}
```
cloetzi: thought(blocking): Path is potentially problematic if we want to keep the door open for remote execution. In that case, that would straight up break. We can always add this later when we have clarity _if_ we need a distributed strategy, but for now, could we focus on base64 and URLs only?

We could consider a special URL like `file://<path>` as an alternative that has a special parser for URLs (and it could be disabled in distributed mode). I'd appreciate if we can chat about this before merging.

lib/riffer/file_part.rb
```diff
@@ -0,0 +1,155 @@
...
+    hash[:data] = @data if @data
+    hash[:url] = @url_string if @url_string
+    hash[:filename] = filename if filename
+    hash
+  end
+
+  private
+
+  #: () -> String
+  def fetch_url_data
```
cloetzi: thought(blocking): While this is super neat, it is also easy enough to implement outside of riffer. Fetching files from URLs can have some hidden complexity (e.g. redirects, slow connections, flaky connections etc). 

While I am not opposed to add this later if there's a need (and hopefully bedrock removes the need of downloading files on our side). For the initial file support I'd prefer to keep it simple and have developers deal with fetching the files and convert them to base64. IMO it is not a huge loss of functionality, but removes hidden complexity from the gem. 

I am not going to die on this hill, but would like to discuss more and mark it as blocking for now.

lib/riffer/providers/base.rb
```diff
@@ -87,15 +87,16 @@ def validate_input!(prompt:, system:, messages:)
     end
   end
 
-  #: (prompt: String?, system: String?, messages: Array[Hash[Symbol, untyped] | Riffer::Messages::Base]?) -> Array[Riffer::Messages::Base]
-  def normalize_messages(prompt:, system:, messages:)
+  #: (prompt: String?, system: String?, messages: Array[Hash[Symbol, untyped] | Riffer::Messages::Base]?, ?files: Array[Hash[Symbol, untyped] | Riffer::FilePart]?) -> Array[Riffer::Messages::Base]
+  def normalize_messages(prompt:, system:, messages:, files: nil)
     if messages
       return messages.map { |msg| convert_to_message_object(msg) }
```
cloetzi, line 97: issue(non-blocking): This was flagged by Claude and I don't have the focus anymore to cross-check it. Mostly flagging it for the second review.

This early return drops all files if messages are provided. Is that intended?

lib/riffer/providers/amazon_bedrock.rb
```diff
@@ -259,6 +261,36 @@ def append_tool_result(conversation_messages, message)
...
+  #: (Riffer::FilePart) -> Hash[Symbol, untyped]
+  def convert_file_part_to_bedrock_format(file)
+    format = bedrock_format(file.media_type)
+    bytes = Base64.decode64(file.data)
+
+    if file.image?
+      {image: {format: format, source: {bytes: bytes}}}
+    else
+      block = {document: {format: format, source: {bytes: bytes}}}
+      block[:document][:name] = file.filename if file.filename
```
cloetzi: issue(non-blocking): According to Claude, the API call will fail if `name` is nil. As such, the if statement would be incorrect. I haven't found the full reference yet, note for follow-up.

lib/riffer/providers/open_ai.rb
```diff
@@ -264,6 +270,19 @@ def convert_assistant_to_openai_format(message)
...
   end
 
+  #: (Riffer::FilePart) -> Hash[Symbol, untyped]
+  def convert_file_part_to_openai_format(file)
+    if file.image?
+      image_url = file.url? ? file.url : "data:#{file.media_type};base64,#{file.data}"
+      {type: "input_image", image_url: image_url}
+    else
+      data_uri = "data:#{file.media_type};base64,#{file.data}"
+      block = {type: "input_file", file_data: data_uri}
```
cloetzi, line 280: issue(non-blocking): Another Claude issue that I haven't cross-checked: Per OpenAI docs, the input_file content part uses file_data as a nested object with {filename:, file_data:}, not as a top-level string. Double-check this matches the current API shape—the VCR cassettes only test images, not documents.

cloetzi (changes requested):
Blocking because I'd like to chat if we wanna support paths and URLs in the first iteration. I haven't fully gotten through the PR yet, there are some comments I haven't cross-checked yet, but left there so we don't forget.

docs/03_AGENTS.md
bottrall: 👍🏻 removed.

lib/riffer/file_part.rb
bottrall: 👍🏻 removed.

lib/riffer/providers/base.rb
bottrall: Add a raise when messages and files are provided.

lib/riffer/providers/amazon_bedrock.rb
bottrall: Good catch, this was left over from a refactor. You are correct the request will fail if a valid document name is not provided. However I opted to shift this responsibility onto the caller. The `if` statement was accidentally not removed.

lib/riffer/providers/open_ai.rb
bottrall: Added VCRs to all providers to confirm that both images and documents can be uploaded.

cloetzi (approved):

PR #137, opened 2026-02-23
refactor: move built-in guardrails and evals to examples

bottrall:
## Summary

- Remove `Riffer::Guardrails::MaxLength` and `Riffer::Evals::Evaluators::AnswerRelevancy` from the gem's lib/ tree (along with their tests and RBS signatures)
- Add 15 copy-paste reference implementations in `examples/` (12 evaluators + 3 guardrails)
- Update docs and RDoc comments to reference the example classes instead of the deleted built-in classes
- Exclude `examples/` from the gem package via gemspec
- Update evaluator example Usage comments to show the full agent wiring flow (eval profile → agent inclusion → `run_eval`) instead of standalone `evaluator.evaluate()` snippets
- Add `ground_truth:` to `run_eval` calls in evaluator examples that reference ground truth
- Remove `tool_call_accuracy` evaluator (its ground_truth format is incompatible with a shared ground_truth string)
- Move examples tips to the top of evals and guardrails docs
- Raise `ArgumentError` in `KeywordCoverageEvaluator` when `ground_truth` is missing

### Evaluator examples (12)

| File | Type | higher_is_better |
|------|------|-----------------|
| `answer_relevancy.rb` | LLM judge | true |
| `answer_similarity.rb` | LLM judge | true |
| `completeness.rb` | LLM judge | true |
| `prompt_alignment.rb` | LLM judge | true |
| `tone_consistency.rb` | LLM judge | true |
| `faithfulness.rb` | LLM judge | true |
| `context_precision.rb` | LLM judge | true |
| `context_relevance.rb` | LLM judge | true |
| `toxicity.rb` | LLM judge | false |
| `bias.rb` | LLM judge | false |
| `hallucination.rb` | LLM judge | false |
| `keyword_coverage.rb` | Rule-based | true |

### Guardrail examples (3)

| File | Description |
|------|-------------|
| `max_length.rb` | Block messages exceeding max character length |
| `unicode_normalizer.rb` | NFC normalize, strip control chars, collapse whitespace |
| `token_limiter.rb` | Manage token limits with truncate/block strategies |

## Test plan

- [x] `bundle exec rake` passes (959 tests, 0 failures, lint clean, type check clean)
- [x] No remaining references to `Riffer::Guardrails::MaxLength` or `Riffer::Evals::Evaluators::AnswerRelevancy` in lib/ or test/
- [x] All 15 example files exist with consistent format
- [x] `examples/` excluded from gemspec
[...]

PR #138, opened 2026-02-23
feat: replace Profile/Metric eval system with EvaluatorRunner

bottrall:
## Summary

- Replace the Profile-based eval approach (where eval modules are `include`d into Agent classes) with a standalone `EvaluatorRunner`
- Delete `Profile`, `Metric`, and old `Runner` classes — evals are now a CI/test concern rather than an agent definition concern
- Add `EvaluatorRunner.run(agent:, scenarios:, evaluators:)` as the single entry point for running evaluations
- Add `ScenarioResult` for per-scenario evaluation details (input, output, ground_truth, results)
- Rework `RunResult` for multi-scenario aggregate scores (`#scores`, `#to_h`)
- Restore `tool_context:` support — top-level kwarg on `EvaluatorRunner.run` with per-scenario override via `scenario[:tool_context]`
- Update all 12 example evaluator usage docs to use `EvaluatorRunner` API (replacing deleted `Profile`/`metric`/`run_eval` references)
- Replace non-existent `Riffer::Evals::Evaluators::AnswerRelevancy` references with `AnswerRelevancyEvaluator` in docs and RDoc
- Rename "Built-in Evaluators" section to "Example Evaluators" (they were moved to `examples/` in #137)
- Rewrite `docs/08_EVALS.md` and regenerate RBS type signatures
- Remove `RunResult#summary` — trivially available via `scenario_results.length`
- Rename `evals:` keyword parameter to `evaluators:` to avoid confusion with Ruby's built-in `eval`

## Test plan

- [x] `bundle exec rake` — tests + lint pass (939 tests, 0 failures)
- [x] `bundle exec steep check` — no type errors
- [x] `bundle exec rake rbs:generate` — RBS signatures regenerated
- [x] Grep for stale `Profile`/`Metric`/`Evaluators::` references — none remaining
- [x] `tool_context` passed through to agent from top-level kwarg
- [x] Per-scenario `tool_context` overrides top-level value


docs/08_EVALS.md
```diff
@@ -4,41 +4,39 @@ Evals let you measure the quality of agent outputs using LLM-as-judge evaluation
...
-result.aggregate_score # => 0.91
+# 3. Run evals
+result = Riffer::Evals::EvaluatorRunner.run(
+  agent: MyAgent,
+  scenarios: [
+    { input: "What is Ruby?", ground_truth: "A programming language" },
+    { input: "What is Python?" }
+  ],
+  evals: [Riffer::Evals::Evaluators::AnswerRelevancy]
+)
```
Thibaut, line 38: suggestion (non-blocking): removing the class-level DSL makes sense to me, but this here is still running evals for a specific agent. The agent is the "subject" of the method call. Therefore I would find the following more readable/self-explanatory/Ruby-like:


```suggestion
result = MyAgent.run_evals(
  scenarios: [
    { input: "What is Ruby?", ground_truth: "A programming language" },
    { input: "What is Python?" }
  ],
  evals: [Riffer::Evals::Evaluators::AnswerRelevancy]
)
```

Thibaut (commented):
just passing by 😄

docs/08_EVALS.md
bottrall: I'm not opposed to that 👍🏻
@cloetzi which do you prefer considering you'll be the first to dog food it?

docs/08_EVALS.md
```diff
@@ -4,41 +4,39 @@ Evals let you measure the quality of agent outputs using LLM-as-judge evaluation
...
+
+result.scores   # => { Riffer::Evals::Evaluators::AnswerRelevancy => 0.85 }
+result.summary  # => { total_scenarios: 2 }
```
cloetzi: thought(non-blocking): That seems of limited use if the output is indeed only `{ total_scenarios: 2 }`.

lib/riffer/evals/evaluator_runner.rb
```diff
@@ -0,0 +1,79 @@
...
+  #
+  # +agent+ - an Agent subclass (not an instance).
+  # +scenarios+ - array of hashes with +:input+ and optional +:ground_truth+.
+  # +evals+ - array of Evaluator subclasses to run against each scenario.
+  #
+  # Raises Riffer::ArgumentError if agent is not a Riffer::Agent subclass
+  # or any eval is not a Riffer::Evals::Evaluator subclass.
+  #
+  #: (agent: singleton(Riffer::Agent), scenarios: Array[Hash[Symbol, untyped]], evals: Array[singleton(Riffer::Evals::Evaluator)]) -> Riffer::Evals::RunResult
+  def self.run(agent:, scenarios:, evals:)
```
cloetzi: nitpick(non-blocking): What do you think about spelling out `evaluators` instead of `evals` throughout the code to differentiate from Ruby `eval`? Torn on it if it is better or worse.

lib/riffer/evals/run_result.rb
```diff
@@ -1,99 +1,64 @@
...
+      hash[evaluator] = total / counts[evaluator]
     end
+  end
 
-    weighted_sum / total_weight
+  # Returns summary statistics for the run.
+  #
+  #: () -> Hash[Symbol, untyped]
+  def summary
+    {total_scenarios: scenario_results.length}
```
cloetzi: thought(non-blocking): As mentioned above, with just this content, I don't think that deserves a method 😅 I can't ever see this being helpful, or what am I missing?

docs/08_EVALS.md
cloetzi: I personally prefer the separation. I have a pretty strong opinion on not having test code in production - and while this is a different paradigm and not fully comparable to rspec/ruby test env - I do like the separation of the test invocation from the class. 

However, either works fine and is an improvement over the current interface, but I was quite stoked about the new runner class 😄

cloetzi (changes requested):
issue(blocking): The PR dropped the `tool_context` that I believe we still need to test some more complex agents.

cloetzi (approved):
Thanks for the improvements, much cleaner 😌

PR #139, opened 2026-02-25
refactor: store structured output as parsed hash on assistant messages

bottrall:
## Summary
- Replace the monolithic `extract_assistant_message` provider hook with two focused hooks: `extract_content` and `extract_tool_calls`
- The base class `generate_text` now assembles the `Assistant` message and parses structured output JSON, storing the parsed hash directly on the message (instead of a boolean flag with lazy parsing)
- `Messages::Assistant#structured_output` is now an `attr_reader` returning the stored `Hash` or `nil` — no more lazy JSON parsing or `attr_writer :structured`
- `Messages::Converter` round-trips via `structured_output` key (hash value) instead of `structured` (boolean)
- All four providers (OpenAI, Anthropic, Bedrock, Mock) updated to implement the new hooks
- Docs and `.agents/providers.md` updated to reflect the new provider API

## Test plan
- `bundle exec rake` passes (950 tests, 0 failures, lint + Steep type check clean)
- Updated tests verify `structured_output` stores parsed hash, `structured_output?` checks for non-nil, `to_h` serialization includes the hash, and converter round-tripping works with hash values
- Existing agent structured output tests continue to pass

### DIFF
--- a/docs/03_AGENTS.md
+++ b/docs/03_AGENTS.md
@@ -425,14 +425,14 @@ response.structured_output  # => {sentiment: "positive", score: 0.95}
 
 Returns `nil` when structured output is not configured or when validation fails.
 
-The assistant message in the message history is also flagged, so you can access structured output directly from persisted messages:
+The assistant message in the message history stores the parsed hash, so you can access structured output directly from persisted messages:
 
 ```ruby
 agent = SentimentAgent.new
 agent.generate('Analyze: "I love this!"')
 
 msg = agent.messages.last
-msg.is_structured_output  # => true
+msg.structured_output?    # => true
 msg.structured_output     # => {sentiment: "positive", score: 0.95}
 ```
 
--- a/docs/03_AGENTS.md
+++ b/docs/03_AGENTS.md
@@ -425,6 +425,19 @@ response.structured_output  # => {sentiment: "positive", score: 0.95}
 
 Returns `nil` when structured output is not configured or when validation fails.
 
+The assistant message in the message history is also flagged, so you can access structured output directly from persisted messages:
+
+```ruby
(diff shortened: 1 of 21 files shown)


docs/03_AGENTS.md
```diff
@@ -425,6 +425,19 @@ response.structured_output  # => {sentiment: "positive", score: 0.95}
...
 Returns `nil` when structured output is not configured or when validation fails.
 
+The assistant message in the message history is also flagged, so you can access structured output directly from persisted messages:
+
+```ruby
+agent = SentimentAgent.new
+agent.generate('Analyze: "I love this!"')
+
+msg = agent.messages.last
+msg.is_structured_output  # => true
```
Thibaut: suggestion (non-blocking): `message.structured_output?` is more Ruby-like

lib/riffer/messages/assistant.rb
```diff
@@ -19,11 +21,29 @@ class Riffer::Messages::Assistant < Riffer::Messages::Base
...
+    @is_structured_output = is_structured_output
+  end
+
+  # Parses content as structured output JSON.
+  #
+  # Returns the parsed hash when +is_structured_output+ is true and
+  # content is valid JSON, +nil+ otherwise.
+  #
+  #: () -> Hash[Symbol, untyped]?
+  def structured_output
```
Thibaut: question (non-blocking): should that be memoized? It's a risk if ever it gets mutated downstream, but also generally expected unless the method is called `parse_structured_output` or something that indicates the content is parsed every time.

Thibaut (approved):

janetxinli (approved):

PR #144, opened 2026-02-26
feat: nested params DSL and strict schema for all providers

cloetzi:
## Summary

- Add nested params DSL shared by structured output and tools: nested objects (`Hash` with block), typed arrays (`Array, of:`), and arrays of objects (`Array` with block)
- Enable arbitrarily deep nesting with recursive JSON Schema generation and validation
- Dot-path error messages for nested failures (e.g. `address.city is required`, `items[1].qty is required`)
- Deep key symbolization so nested values always use symbol keys consistently
- Guard rails: `of:` + block raises, `of:` on non-Array raises, `of:` with non-primitive raises
- Move `strict_schema`/`make_nullable` to `Providers::Base` and apply to **all** schemas sent to LLMs (both structured output and tools, across all three providers)
- Remove blank string normalization — with strict schemas, all providers return proper `null` for optional fields
- Fix OpenAI tool params: now applies `strict_schema` before sending (previously only structured output was transformed)

## Commits

1. **`feat: support nested structured_output definitions`** — Core DSL changes to `Param` and `Params`, unit tests, RBS
2. **`feat: strict_schema for all provider schemas`** — `strict_schema` in Base, applied to all providers, blank string normalization removed, strict mode tests
3. **`test: VCR integration tests for nested structured output`** — 12 cassettes (4 scenarios × 3 providers)
4. **`docs: nested structured output and strict schema documentation`** — Overview, Agents, Tools, README, providers guide

## Test plan

- [x] All 998 tests pass (`bundle exec rake`)
- [x] StandardRB lint clean
- [x] RBS type signatures regenerated, Steep check passes
- [x] VCR cassettes recorded for nested objects, typed arrays, arrays of objects, and null optional fields across OpenAI, Anthropic, and Bedrock
- [x] Unit tests for nested DSL, validation (dot-path errors, indexed errors, deep symbolization), null optional fields, and strict schema transformation


lib/riffer/providers/open_ai.rb
```diff
@@ -296,4 +296,68 @@ def convert_tool_to_openai_format(tool)
...
+
+  # Transforms a JSON Schema for OpenAI strict mode compliance.
+  #
+  # OpenAI strict mode requires every property to appear in +required+.
+  # Properties that were optional are made nullable instead.
+  #
+  # See https://platform.openai.com/docs/guides/structured-outputs#all-fields-must-be-required
+  #
+  #: (Hash[Symbol, untyped]) -> Hash[Symbol, untyped]
+  def strict_schema(schema)
```
cloetzi: I am wondering if we should apply this transformation to bedrock and anthropic too? As you can see [here](https://github.com/janeapp/riffer/pull/144/changes#diff-f7cdc55242ed1949cb0576f8b42a99ac4602d8c10b273ffb101296760fb3acfeR70), they kinda can inject garbage into optional strings instead of cleanly returning `null`.

lib/riffer/params.rb
```diff
@@ -58,6 +64,11 @@ def validate(arguments)
     @parameters.each do |param|
       value = arguments[param.name]
 
+      # Providers without strict mode (Anthropic, Bedrock) may return blank
+      # strings instead of null for optional fields. Normalize to nil so
+      # optional params get their default and required params fail cleanly.
+      value = nil if value.is_a?(String) && value.strip.empty?
```
cloetzi: This will also affect tool calls, as in if an agent requests a tool call with an empty string for an optional parameter, it will be nullified. I can't see a valid reason where we'd want to pass in an empty string as an argument instead of null, however, it is not impossible to introduce bugs for edge cases I haven't thought about. Thoughts?

docs/03_AGENTS.md
```diff
@@ -149,6 +149,74 @@ end
...
+      required :address, Hash, description: "Address" do
+        required :street, String
+        required :city, String
+      end
+    end
+  end
+end
+```
+
+#### Guard Rails
```
Thibaut:
```suggestion
#### Guardrails
```

Thibaut: Actually with the next section being about guardrails, maybe this is better


```suggestion
#### Limitations
```

lib/riffer/providers/base.rb
```diff
@@ -95,6 +95,53 @@ def parse_structured_output(content)
...
+  #
+  # Strict mode requires every property to appear in +required+.
+  # Properties that were optional are made nullable instead.
+  #
+  # See https://platform.openai.com/docs/guides/structured-outputs#all-fields-must-be-required
+  #
+  #: (Hash[Symbol, untyped]) -> Hash[Symbol, untyped]
+  def strict_schema(schema)
+    if schema[:type] == "array" && schema[:items]
+      return schema.merge(items: strict_schema(schema[:items]))
```
Thibaut: question (non-blocking): Can we mutate in place rather than duplicating everything in the tree? `make_nullable` returns `prop` in some cases, so we aren't actually making a deep clone of the original object. From a performance standpoint I think replacing all these `.merge` with `[]=` could make a meaningful difference (especially GC sine we go through many objects).

lib/riffer/param.rb
```diff
@@ -60,6 +67,15 @@ def to_json_schema
     schema = {type: type_name}
     schema[:description] = description if description
     schema[:enum] = enum if enum
+
+    if type == Array && nested_params
+      schema[:items] = nested_params.to_json_schema
+    elsif type == Array && item_type
+      schema[:items] = {type: TYPE_MAPPINGS[item_type]}
+    elsif type == Hash && nested_params
+      schema.merge!(nested_params.to_json_schema.except(:type))
```
Thibaut:
```suggestion
      schema.merge!(nested_params.to_json_schema)
```

We asserted `type == Hash` and `Riffer::Params#to_json_schema` is always `type: "object"`, so the `except` is unnecessary.

lib/riffer/params.rb
```diff
@@ -105,4 +113,96 @@ def to_json_schema
...
+      validate_typed_array(param, value, errors)
+      value
+    else
+      value
+    end
+  end
+
+  #: (Riffer::Param, Hash[Symbol, untyped], Array[String]) -> Hash[Symbol, untyped]
+  def validate_nested_hash(param, value, errors)
+    sym_value = deep_symbolize_keys(value)
```
Thibaut, line 166: thought (non-blocking): at scale these add up to a serious tax on performance. Wish we didn't have to do it 😅

lib/riffer/params.rb
```diff
@@ -105,4 +113,96 @@ def to_json_schema
...
+      end
+      sym_item
+    end
+  end
+
+  #: (Riffer::Param, Array[untyped], Array[String]) -> void
+  def validate_typed_array(param, value, errors)
+    type_name = Riffer::Param::TYPE_MAPPINGS[param.item_type]
+    value.each_with_index do |item, i|
+      valid = if param.item_type == TrueClass || param.item_type == FalseClass
```
Thibaut, line 196: question (non-blocking): this makes me realize that in order to have a boolean param, one must pick either `TrueClass` or `FalseClass` (we don't support `[TrueClass, FalseClass]`), and then this ends up allowing both `true` and `false`. Am I understanding right? That seems very suboptimal 😅

Shame Ruby doesn't have `Boolean`. As a least-bad alternative, I'd suggest we turn `TrueClass` and `FalseClass` into `Riffer::Boolean`. This can be a follow-up and is unrelated to this PR, but important question (none of the examples show boolean params, we should show that and then the problem will become obvious)

Thibaut (approved):
one question on booleans worth following-up on, otherwise 💯

lib/riffer/params.rb
cloetzi: Yeah, I'll see if we even need it in a subsequent PR paired with the bool change below.

lib/riffer/params.rb
cloetzi: As suggested, I'll improve this in a subsequent PR.

lib/riffer/providers/base.rb
cloetzi: I ended up refactoring this from scratch. I realized that we can just build the strict schema directly and add support for this to our `Param` and `Params` classes. That removes the need for the ugly restructuring.  

See [c337c50](https://github.com/janeapp/riffer/pull/144/commits/c337c50070a8f1844f8b39df30a51e25f5b164c3).

Thibaut: Neat

Thibaut (approved):

PR #147, opened 2026-02-27
feat: Add Riffer::Boolean sentinel type for boolean params

cloetzi:
## Summary

- Add `Riffer::Boolean` module as a sentinel type so users can write `required :verbose, Riffer::Boolean` instead of `TrueClass`/`FalseClass`
- Update `Param` and `Params` validation to recognize the new type
- Keep `TrueClass`/`FalseClass` working for backwards compatibility
- Update docs and add tests

## Test plan

- [x] All tests pass
- [x] StandardRB lint passes
- [x] Steep type checking passes

### DIFF
--- a/docs/04_TOOLS.md
+++ b/docs/04_TOOLS.md
@@ -104,9 +104,12 @@ Options:
 | `String`                   | `string`         |
 | `Integer`                  | `integer`        |
 | `Float`                    | `number`         |
+| `Riffer::Boolean`          | `boolean`        |
 | `TrueClass` / `FalseClass` | `boolean`        |
 | `Array`                    | `array`          |
 | `Hash`                     | `object`         |
+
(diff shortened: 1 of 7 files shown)


docs/04_TOOLS.md
```diff
@@ -104,10 +104,13 @@ Options:
 | `String`                   | `string`         |
 | `Integer`                  | `integer`        |
 | `Float`                    | `number`         |
+| `Riffer::Boolean`          | `boolean`        |
 | `TrueClass` / `FalseClass` | `boolean`        |
 | `Array`                    | `array`          |
 | `Hash`                     | `object`         |
 
+`Riffer::Boolean` is the preferred way to declare boolean parameters. `TrueClass` and `FalseClass` continue to work for backwards compatibility.
```
Thibaut, line 112: suggestion (non-blocking): if we have no `TrueClass` / `FalseClass` in master I would skip this. Otherwise it will stay forever 😅

Thibaut (approved):

PR #148, opened 2026-02-27
refactor: standardize on symbol keys for internal hashes

cloetzi:
## Summary

- Use `JSON.parse(str, symbolize_names: true)` at parse boundaries instead of `JSON.parse` followed by `transform_keys(&:to_sym)` — applies to `agent.rb`, `structured_output.rb`, `evals/judge.rb`, and `providers/anthropic.rb`
- Remove dual string/symbol key access (`hash[:key] || hash["key"]`) in the message converter — string keys were never part of the public API
- Remove `deep_symbolize_keys` from `Params` and the Hash branch from `Agent#parse_tool_arguments` since all providers pass JSON strings (not pre-parsed hashes)
- Narrow RBS type signatures from `Hash[Symbol | String, untyped]` to `Hash[Symbol, untyped]`
- Document the hash key convention in `.agents/code-style.md`

## Test plan

- [x] All tests pass
- [x] StandardRB lint passes
- [x] Steep type checking passes
- [ ] Verify no downstream consumers pass string-keyed hashes to Riffer

### DIFF
--- a/lib/riffer/evals/judge.rb
+++ b/lib/riffer/evals/judge.rb
@@ -120,9 +120,9 @@ class Riffer::Evals::Judge
     tool_call = response.tool_calls.first
     raise Riffer::Error, "Invalid judge response: no tool call found" unless tool_call
 
-    parsed = JSON.parse(tool_call[:arguments])
-    score = parsed["score"]
-    reason = parsed["reason"]
+    parsed = JSON.parse(tool_call[:arguments], symbolize_names: true)
+    score = parsed[:score]
+    reason = parsed[:reason]
 
     raise Riffer::Error, "Invalid judge response: missing score" if score.nil?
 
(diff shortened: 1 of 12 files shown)


lib/riffer/evals/judge.rb
```diff
@@ -120,9 +120,9 @@ def parse_tool_response(response)
     tool_call = response.tool_calls.first
     raise Riffer::Error, "Invalid judge response: no tool call found" unless tool_call
 
-    parsed = JSON.parse(tool_call[:arguments])
-    score = parsed["score"]
-    reason = parsed["reason"]
+    parsed = JSON.parse(tool_call[:arguments], symbolize_names: true)
```
Thibaut, line 123: TIL `symbolize_names` is a thing. Neat!

Thibaut (approved):
Niiiiiiice! :so-good:

PR #151, opened 2026-03-02
feat: add support for agent skills

Thibaut:
### Summary

- Add skills support using the [Agent Skills spec](https://agentskills.io/) tool-based activation pattern
- Skills are discovered through a pluggable `Backend` interface (built-in `FilesystemBackend` scans directories for `SKILL.md` files)
- Only frontmatter (name + description) is injected into the system prompt; full instructions are loaded on-demand via a `skill_activate` tool
- A skill `Adapter` encapsulates provider-specific behavior: how the catalog is rendered in the system prompt and which tool the LLM calls to activate skills (XML for Anthropic, Markdown for others)
- New agent DSL: `skills` block with `backend`, `adapter`, and `activate` options

### Out of scope

- Reference files, asset files, and script execution from skill directories
- Skill search tool (applications provide their own — the skill catalog is available in `tool_context`)
- Explicit invocation syntax (e.g., `/skill-name` parsing from user input)
- Skill versioning, pinning, registries, trust/sandboxing

### DIFF
--- a/docs/10_SKILLS.md
+++ b/docs/10_SKILLS.md
@@ -1,0 +1,164 @@
+# Skills
+
+Skills are packaged AI agent capabilities per the [Agent Skills spec](https://agentskills.io/). Each skill is a directory containing a `SKILL.md` file with YAML frontmatter and Markdown instructions. The framework discovers skills through a pluggable backend, injects a compact catalog into the system prompt (~50 tokens/skill), and provides a tool for the LLM to activate skills on demand.
+
+## Creating a Skill
+
+Create a directory with a `SKILL.md` file:
+
+```
+.skills/
+  code-review/
+    SKILL.md
+  data-analysis/
+    SKILL.md
+```
+
+Each `SKILL.md` has YAML frontmatter and a Markdown body:
+
+```markdown
+---
+name: code-review
+description: Reviews code for quality, style, and potential issues.
+---
+You are a code review assistant.
+
+Review the code for:
+- Style issues
+- Potential bugs
+- Performance problems
+```
+
+**Required frontmatter fields:**
+
+- `name` — lowercase alphanumeric with hyphens, 1-64 chars (must match directory name)
+- `description` — 1-1024 chars, helps the LLM decide when to activate
+
+Additional frontmatter keys are passed through as metadata.
+
+## Configuring an Agent
+
+Use the `skills` block DSL to configure skills:
+
+```ruby
+class MyAgent < Riffer::Agent
+  model "openai/gpt-4o"
+  instructions "You are a helpful assistant."
+  skills do
+    backend Riffer::Skills::FilesystemBackend.new(".skills")
+  end
+end
+```
+
+Multiple directories can be scanned (first-path-wins for duplicates):
+
+```ruby
+skills do
+  backend Riffer::Skills::FilesystemBackend.new(".skills", "~/.riffer/skills")
+end
+```
+
+### Dynamic Backend via Proc
+
+```ruby
+skills do
+  backend ->(context) { tenant_backend(context[:tenant_id]) }
+end
+```
+
+### Custom Adapter
+
+The adapter controls how the skill catalog is rendered in the system prompt and which tool the LLM calls to activate a skill. The adapter is auto-selected by provider (Markdown for most, XML for Anthropic). Override with:
+
+```ruby
+skills do
+  backend Riffer::Skills::FilesystemBackend.new(".skills")
+  adapter Riffer::Skills::XmlAdapter
+end
+```
+
+### Activated Skills
+
+Load skill instructions into the system prompt at startup (no tool call needed):
+
+```ruby
+skills do
+  backend Riffer::Skills::FilesystemBackend.new(".skills")
+  activate ["code-review"]
+end
+```
+
+Accepts a Proc for dynamic resolution:
+
+```ruby
+skills do
+  backend Riffer::Skills::FilesystemBackend.new(".skills")
+  activate ->(context) { context[:active_skills] || [] }
+end
+```
+
+## How It Works
+
+1. **Discovery** — At the start of `generate`/`stream`, the backend's `list_skills` returns frontmatter for all available skills.
+2. **Catalog injection** — The adapter formats the catalog and appends it to the system prompt.
+3. **Activation** — When the LLM matches a task to a skill, it calls the `skill_activate` tool with the skill name. The tool returns the full SKILL.md body.
+4. **Execution** — The LLM follows the skill's instructions to complete the task.
+
+## Custom Backends
+
+Implement `Riffer::Skills::Backend` for non-filesystem storage:
+
+```ruby
+class DatabaseBackend < Riffer::Skills::Backend
+  def list_skills
+    # Return Array[Riffer::Skills::Frontmatter]
+  end
+
+  def read_skill(name)
+    # Return String (skill body)
+    # Raise Riffer::ArgumentError if not found
+  end
+end
+```
+
+## Custom Adapters
+
+Subclass `Riffer::Skills::Adapter` to customize how the skill catalog is rendered and which tool the LLM uses to activate skills:
+
+```ruby
+class CustomAdapter < Riffer::Skills::Adapter
+  def render_catalog(skills)
+    # Return String (skill catalog for the system prompt)
+    # Use `activate_tool.name` to reference the activation tool
+  end
+
+  def activate_tool
+    # Return a Riffer::Tool subclass
+    # Defaults to Riffer::Skills::ActivateTool
+    Riffer::Skills::ActivateTool
+  end
+end
+```
+
+The built-in adapters are `Riffer::Skills::MarkdownAdapter` (default) and `Riffer::Skills::XmlAdapter` (used by Anthropic).
+
+## Accessing Skills in Tools
+
+The skills context (`Riffer::Skills::Context`) is available via `context[:skills]` during execution:
+
+```ruby
+class SkillSearchTool < Riffer::Tool
+  identifier "skill_search"
+  description "Searches available skills."
+
+  params do
+    required :query, String
+  end
+
+  def call(context:, query:)
+    skills_context = context[:skills]  # Riffer::Skills::Context
+    matches = skills_context.skills.values.select { |s| s.description.include?(query) }
+    json(matches.map { |s| {name: s.name, description: s.description} })
+  end
+end
+```
--- a/lib/riffer/agent.rb
+++ b/lib/riffer/agent.rb
@@ -142,6 +142,25 @@ class Riffer::Agent
     end
   end
 
+  # Configures skills for this agent via a block DSL.
+  #
+  # Returns the current Riffer::Skills::Config when called without a block.
+  #
+  #   skills do
+  #     backend Riffer::Skills::FilesystemBackend.new(".skills")
+  #     adapter Riffer::Skills::XmlAdapter
+  #     activate ["code-review"]
+  #   end
+  #
+  #: () ?{ () -> void } -> Riffer::Skills::Config?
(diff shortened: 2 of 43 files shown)


docs/10_SKILLS.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+### Activated Skills
+
+Load skill instructions into the system prompt at startup (no tool call needed):
+
+```ruby
+skills do
+  backend Riffer::Skills::FilesystemBackend.new(".skills")
+  activate ["code-review"]
```
Thibaut, line 87: This could have been a follow-up, but somewhat justifies the `skills do ... end` DSL. Another one I have in mind is `allow ["code-review"]` to reduce the set of skills available to use by that agent out of the backend.

lib/riffer/skills/filesystem_backend.rb
```diff
@@ -0,0 +1,79 @@
...
+      next unless File.directory?(path)
+
+      Dir.children(path).sort.each do |dirname|
+        dir = File.join(path, dirname)
+        skill_file = File.join(dir, SKILL_FILENAME)
+        next unless File.directory?(dir) && File.file?(skill_file)
+
+        frontmatter = Riffer::Skills::Frontmatter.parse_frontmatter(File.read(skill_file))
+
+        validate_dirname_matches_name!(dirname, frontmatter.name)
```
Thibaut, line 45: I'm on the fence whether to validate this rule as well as the frontmatter rules at runtime, given they should always be enforced wherever skills are created (but Riffer doesn't see that) and validating adds some runtime overhead (negligible for now but could increase).

lib/riffer/agent.rb
Thibaut, line 489: This is done via `prepare_run` now.

docs/10_SKILLS.md
```diff
@@ -0,0 +1,164 @@
...
+skills do
+  backend Riffer::Skills::FilesystemBackend.new(".skills", "~/.riffer/skills")
+end
+```
+
+### Dynamic Backend via Proc
+
+```ruby
+skills do
+  backend ->(ctx) { tenant_backend(ctx[:tenant_id]) }
```
cloetzi: question(non-blocking): What's the `ctx` here?

docs/10_SKILLS.md
```diff
@@ -0,0 +1,164 @@
...
+    # Return a Riffer::Tool subclass
+    # Defaults to Riffer::Skills::ActivateTool
+    Riffer::Skills::ActivateTool
+  end
+end
+```
+
+The built-in adapters are `Riffer::Skills::MarkdownAdapter` (default) and `Riffer::Skills::XmlAdapter` (used by Anthropic).
+
+## Accessing Skills in Tools
```
cloetzi, line 145: question(non-blocking): Just for my understanding, where is this useful/needed?

lib/riffer/skills/activate_tool.rb
```diff
@@ -0,0 +1,34 @@
...
+#
+# Registered automatically when an agent has skills configured.
+# The LLM calls this when a task matches an available skill's description.
+#
+# See Riffer::Skills::Context.
+class Riffer::Skills::ActivateTool < Riffer::Tool
+  identifier "skill_activate"
+  description "Activates a skill and returns its instructions. " \
+              "Call this when a task matches an available skill's description."
+  timeout 5
```
cloetzi: suggestion(non-blocking): Seems like this could be lower by default.

lib/riffer/skills/activate_tool.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+# rbs_inline: enabled
+
+# Tool for the LLM to activate a skill and receive its full instructions.
+#
+# Registered automatically when an agent has skills configured.
+# The LLM calls this when a task matches an available skill's description.
+#
+# See Riffer::Skills::Context.
+class Riffer::Skills::ActivateTool < Riffer::Tool
```
cloetzi, line 10: thought(non-blocking): This is interesting in the context of distributed tool calls. I'd argue this should always be evaluated inline. We are currently not allowing to set the `tool_runtime` on tools, but I am wondering if we need to expand the DSL? Is not really relevant for this PR, but I wanted to mention it.

lib/riffer/skills/context.rb
```diff
@@ -0,0 +1,75 @@
...
+  # +skills+ - skill catalog indexed by name.
+  # +adapter+ - the adapter used to render skill content.
+  #
+  #: (backend: Riffer::Skills::Backend, skills: Hash[String, Riffer::Skills::Frontmatter], adapter: Riffer::Skills::Adapter) -> void
+  def initialize(backend:, skills:, adapter:)
+    @backend = backend
+    @skills = skills
+    @adapter = adapter
+    @activated = {} #: Hash[String, String]
+    @pre_activated_names = Set.new #: Set[String]
```
cloetzi: nitpick(non-blocking): Is this required? It seems only to be required to read out the `@activated` content, but couldn't this be achieved with `@activated.each` as well?

lib/riffer/skills/filesystem_backend.rb
```diff
@@ -0,0 +1,79 @@
...
+# Built-in backend that reads skills from the filesystem.
+#
+# Scans configured directories for immediate child directories containing
+# +SKILL.md+ files. Directory names must match the skill +name+ field.
+#
+#   backend = Riffer::Skills::FilesystemBackend.new(".skills", "~/.riffer/skills")
+#   backend.list_skills  # => [Riffer::Skills::Frontmatter, ...]
+#   backend.read_skill("code-review")  # => "Full skill instructions..."
+#
+class Riffer::Skills::FilesystemBackend < Riffer::Skills::Backend
```
cloetzi, line 13: thought(non-blocking): I am wondering if this should have a like a `cache_content` flag and if it is set, it reads once for everything and memoizes each step 🤔 It could be tied to the Rails cache classes setting in practice.

lib/riffer/skills/frontmatter.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  #: (String) -> [Hash[Symbol, untyped], String]
+  def self.split_frontmatter(raw) # :nodoc:
+    parts = raw.split(/^---\s*$/, 3)
+
+    if parts.length >= 3 && parts[0].strip.empty?
+      [YAML.safe_load(parts[1])&.transform_keys(&:to_sym) || {}, parts[2].lstrip]
+    else
+      [{}, raw]
```
cloetzi, line 63: nitpick(non-blocking): From robot Claude: When the YAML delimiters are missing or malformed, split_frontmatter returns [{}, raw]. Then parse calls new(name: nil, description: nil, metadata: {}), which raises "name must be a String" — a
  confusing message when the real problem is missing frontmatter.

lib/riffer/skills/xml_adapter.rb
```diff
@@ -0,0 +1,31 @@
...
+  #
+  #: (Array[Riffer::Skills::Frontmatter]) -> String
+  def render_catalog(skills)
+    lines = []
+    lines << "When a user's request matches a skill description below, call the `#{activate_tool.name}` tool with the skill name. After activation, follow the skill's instructions."
+    lines << ""
+    lines << "<available_skills>"
+    skills.each do |skill|
+      lines << "  <skill>"
+      lines << "    <name>#{skill.name}</name>"
```
cloetzi: suggestion(non-blocking): Again, from robot Claude:
>Description is escaped but skill.name is not. The name pattern restricts to [a-z0-9-] so this is safe today, but if the pattern ever relaxes, this becomes an injection vector. Defensive fix:
  CGI.escapeHTML(skill.name).

Seems easy enough to avoid future exploits.

lib/riffer/skills/xml_adapter.rb
```diff
@@ -0,0 +1,31 @@
...
+  # Renders a skill catalog as XML.
+  #
+  # +skills+ - array of Frontmatter objects to render.
+  #
+  #: (Array[Riffer::Skills::Frontmatter]) -> String
+  def render_catalog(skills)
+    lines = []
+    lines << "When a user's request matches a skill description below, call the `#{activate_tool.name}` tool with the skill name. After activation, follow the skill's instructions."
+    lines << ""
+    lines << "<available_skills>"
```
cloetzi, line 21: thought(non-blocking): Have you considered building the XML as a ruby object, then render it? This gives me PHP/HTML jeebies.

cloetzi (commented):
I did a first pass, looks great! All my current comments are non-blocking, but I'd like to do a second pass tomorrow when my mind is fresh.

docs/10_SKILLS.md
Thibaut: `tool_context`. Same as in `model` and others. I'll rename.

docs/10_SKILLS.md
Thibaut: The main use-case I had in mind is the search tool in the example below. But I wouldn't disagree to remove it for now until we actually implement search (the example below is only a naive implementation).

lib/riffer/skills/activate_tool.rb
Thibaut: Great catch! 👏 

I need to do some research and try a few things. I think switching the `tool_runtime` on a per-tool basis could not play with some runtime combinations. Maybe there is a concept of "tool priority", "blocking tool call", or something along those lines.

lib/riffer/skills/context.rb
Thibaut: Last I tried the 🤖 told me it couldn't be before but this wouldn't be the first time it lied to me. I'll try again. It's mainly a risk if the system prompt is recreated after some tools have been activated.

lib/riffer/skills/filesystem_backend.rb
Thibaut: You're right to point out this is relatively native implementation. I didn't want to overengineer it and TBH I think if someone wanted to take this to large-scale production, best they write their own backend.

lib/riffer/skills/xml_adapter.rb
Thibaut: This is just sugar syntax for the 🤖. It's not really XML (it will never be parsed as XML, even by the 🤖). From a perf standpoint, I think much better to do string concat. Even the `CGI.escapeHTML` is technically overkill but I used it to appease the security robots 😅

cloetzi: Ah, yeah, if it is just that, the approach makes sense!

docs/10_SKILLS.md
Thibaut: `context[:skills]` is also used by the `ActivateSkill` tool, which I'm realizing I need to document how to provide a custom one.

tsov:
This looks great! Had a couple of non-blocking questions:                   
                                                                              
1. ~What happens if skills change between turns in a multi-turn conversation? Say a skill gets deleted from disk between generate calls. The old tool result would still be in the message history. Is that something we need to worry about?~ Just had a chat with @cloetzi and sounds like this shouldn't be an issue with how we store messages in the monolith
2. Would it make sense for skill activations to emit stream events? The rest of the framework has them for tool calls, text, etc. Could be useful for logging which skills get used in a run.

Thibaut:
@tsov good call! Added a stream event for skill activation. Decided not to emit events for pre-activated skills since those aren't picked by the LLM and rather act as "extensions of the system prompt" than regular skills.

Thibaut:
This made me catch the issue that pre-activated skills are listed in the catalog, which could result in the LLM activating a pre-activated skill, since it doesn't know what skills were preactivated. Removed pre-activated skills from the catalog.

cloetzi (approved):
As discussed, merging this despite minor issues with `resume` and how we generate in our app. I'll address that in a subsequent PR.

PR #145, opened 2026-02-27
feat: experimental ToolRuntime abstraction

cloetzi:
## Summary

- Introduces `Riffer::Runner` (Sequential/Threaded) as a generic concurrency primitive with `map(items, &block)` interface
- Introduces `Riffer::ToolRuntime` (Inline/Threaded) to decouple tool execution from the agent loop, with `execute` (batch), `call` (single), and `around_tool_execution` class-level callbacks
- Adds `tool_runtime` DSL to agents (per-agent and global config) with symbol, instance, and Proc resolution
- Refactors `Agent#execute_tool_calls` and `execute_pending_tool_calls` to delegate to the runtime
- Removes `execute_tool_call`, `find_tool_class`, `parse_tool_arguments` from Agent (moved to ToolRuntime)
- Config defaults `tool_runtime` to `Riffer::ToolRuntime::Inline.new`
- All marked as **experimental** — may be removed without warning

## Test plan

- [ ] All 980 existing + new tests pass (`bundle exec rake`)
- [ ] New runner tests verify sequential ordering, threaded concurrency, and max_concurrency
- [ ] New tool_runtime tests cover dispatch, error handling, around callbacks, inline/threaded
- [ ] New agent DSL tests cover symbol/lambda/instance resolution, global fallback, per-agent override
- [ ] Config tests verify default and setter

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #154, opened 2026-03-03
feat: auto-derive step offset on resume for max_steps enforcement

cloetzi:
## Summary

- On resume, auto-derive the step offset by counting assistant messages in the conversation history
- `max_steps` is enforced across the full session transparently — no explicit step parameter needed

## Test plan

- [x] In-memory resume hits `max_steps` based on prior assistant messages
- [x] Cross-process resume counts persisted assistant messages for step offset

### DIFF
--- a/lib/riffer/agent.rb
+++ b/lib/riffer/agent.rb
@@ -260,6 +260,8 @@ class Riffer::Agent
   # state from persisted data (useful for cross-process resume).
   #
   # Skips message initialization and before guardrails in both cases.
+  # The step offset is derived automatically from the number of assistant
+  # messages so +max_steps+ is enforced across the full session.
   #
   #: (?messages: Array[Hash[Symbol, untyped] | Riffer::Messages::Base]?, ?tool_context: Hash[Symbol, untyped]?) -> Riffer::Agent::Response
   def resume(messages: nil, tool_context: nil)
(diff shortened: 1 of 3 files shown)


lib/riffer/agent.rb
```diff
@@ -305,7 +307,7 @@ def interrupt!(reason = nil)
 
   #: (?Array[Riffer::Guardrails::Modification], ?resume: bool) -> Riffer::Agent::Response
   def run_generate_loop(all_modifications = [], resume: false)
-    step = 0
+    step = resume ? count_assistant_messages : 0
```
Thibaut, line 310: question (non-blocking): would `count_assistant_messages` work on its own? I assume it would return `0` for a new session.

Thibaut (approved):

lib/riffer/agent.rb
cloetzi: Yes, all it does is `@messages.count { |m| m.is_a?(Riffer::Messages::Assistant) }` and `@messages` is initialized as `[]`.

Thibaut: KK. Then (and sorry I didn't make my suggestion clear 🙈) we can simplify to `step = count_assistant_messages`?

PR #155, opened 2026-03-03
feat: add interrupt! public method for clean loop interrupts

cloetzi:
## Summary

- Add `interrupt!` public method that allows tools or callbacks to cleanly interrupt the agent loop
- Updated docs with interrupt! usage examples

## Test plan

- [x] `interrupt!` stops the agent loop and returns the interrupt event

### DIFF
--- a/docs/03_AGENTS.md
+++ b/docs/03_AGENTS.md
@@ -370,7 +370,6 @@ response = agent.generate('Call the tool')
 response.interrupted?      # => true
 response.interrupt_reason  # => "needs human approval"
 response.content           # => last assistant content before interrupt
-response.step              # => number of LLM calls completed
 ```
 
 **Streaming** — interrupts emit an `Interrupt` event:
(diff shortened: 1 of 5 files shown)


docs/03_AGENTS.md
```diff
@@ -352,17 +352,17 @@ Works with both `generate` and `stream`. Only emits agent-generated messages (As
 
 #### Interrupting the Agent Loop
 
-Callbacks can interrupt the agent loop using Ruby's `throw`/`catch` pattern. This is useful for human-in-the-loop approval, cost limits, or content filtering.
+Callbacks can interrupt the agent loop. This is useful for human-in-the-loop approval, cost limits, or content filtering.
 
-Use `throw :riffer_interrupt` to stop the loop. The response will have `interrupted?` set to `true` and contain the accumulated content up to the point of interruption.
+Use `agent.interrupt!` (or the lower-level `throw :riffer_interrupt`) to stop the loop. The response will have `interrupted?` set to `true` and contain the accumulated content up to the point of interruption.
```
janetxinli, line 357: question(non-blocking): when should the lower-level `:riffer_interrupt` be used instead of `agent.interrupt!`? Should we provide some guidance about that here?

janetxinli (approved):

docs/03_AGENTS.md
cloetzi: The `agent.interrupt!` is mostly syntactic sugar, so it is generally what should be used, but there's no functional difference. The dev can pick what they prefer. 

We might want to consider removing the documentation for `throw` in the future, but we are currently not using it yet, so I don't quite know yet how the interrupt workflows will look in practice.

PR #156, opened 2026-03-03
feat: add experimental ToolRuntime abstraction for tool execution

cloetzi:
## Summary

- Add `ToolRuntime` abstraction with `Inline` (sequential) and `Threaded` (thread pool) implementations
- Add `Runner` abstraction (`Sequential` and `Threaded`) used by tool runtimes for concurrency control
- New `tool_runtime` DSL on agents and global config
- `around_tool_call` callback for wrapping individual tool call execution
- Refactor `execute_tool_calls` / `execute_pending_tool_calls` to delegate to the resolved runtime

## Test plan

- [x] ToolRuntime inline and threaded modes execute tool calls correctly
- [x] Runner sequential and threaded modes work correctly
- [x] `around_tool_call` callbacks compose properly (block, symbol, inheritance)
- [x] DSL resolves symbol, instance, and Proc configurations

### DIFF
--- a/docs/03_AGENTS.md
+++ b/docs/03_AGENTS.md
@@ -219,6 +219,22 @@ Using both `of:` and a block raises `Riffer::ArgumentError`. Using `of:` with a 
 
 Structured output is not compatible with streaming — calling `stream` on an agent with structured output configured raises `Riffer::ArgumentError`.
 
+### tool_runtime (Experimental)
+
+> **Warning:** This feature is experimental and may be removed or changed without warning in a future release.
+
+Configures how tool calls are executed. Defaults to sequential (inline) execution:
+
+```ruby
+class MyAgent < Riffer::Agent
+  model 'openai/gpt-4o'
+  uses_tools [WeatherTool, SearchTool]
+  tool_runtime :threaded
+end
+```
+
+Accepts `:inline`, `:threaded`, a `Riffer::ToolRuntime` instance, or a `Proc`. This setting is **not inherited** by subclasses — each agent class must configure its own tool runtime explicitly. When unset, falls back to `Riffer.config.tool_runtime`. See [Tools — Tool Runtime](04_TOOLS.md#tool-runtime-experimental) for details.
+
 ### guardrail
 
 Registers guardrails for pre/post processing of messages. Pass the guardrail class and any options:
@@ -542,7 +558,7 @@ end
 When an agent receives a response with tool calls:
 
 1. Agent detects `tool_calls` in the assistant message
-2. For each tool call:
+2. The configured tool runtime executes the tool calls (sequentially by default, or concurrently with `:threaded`):
    - Finds the matching tool class
    - Validates arguments against the tool's parameter schema
    - Calls the tool's `call` method with `context` and arguments
--- a/docs/03_AGENTS.md
+++ b/docs/03_AGENTS.md
@@ -233,7 +233,7 @@ class MyAgent < Riffer::Agent
 end
 ```
 
-Accepts `:inline`, `:threaded`, a `Riffer::ToolRuntime` instance, or a `Proc`. This setting is **not inherited** by subclasses — each agent class must configure its own tool runtime explicitly. When unset, falls back to `Riffer.config.tool_runtime`. See [Tools — Tool Runtime](04_TOOLS.md#tool-runtime-experimental) for details.
+Accepts `:inline`, `:threaded`, a `Riffer::ToolRuntime` instance, or a `Proc`. Inherited by subclasses. When unset, falls back to `Riffer.config.tool_runtime`. See [Tools — Tool Runtime](04_TOOLS.md#tool-runtime-experimental) for details.
 
 ### guardrail
 
--- a/docs/03_AGENTS.md
+++ b/docs/03_AGENTS.md
@@ -229,11 +229,11 @@ Configures how tool calls are executed. Defaults to sequential (inline) executio
 class MyAgent < Riffer::Agent
   model 'openai/gpt-4o'
   uses_tools [WeatherTool, SearchTool]
-  tool_runtime :threaded
+  tool_runtime Riffer::ToolRuntime::Threaded
 end
 ```
 
-Accepts `:inline`, `:threaded`, a `Riffer::ToolRuntime` instance, or a `Proc`. Inherited by subclasses. When unset, falls back to `Riffer.config.tool_runtime`. See [Tools — Tool Runtime](04_TOOLS.md#tool-runtime-experimental) for details.
+Accepts a `Riffer::ToolRuntime` subclass, a `Riffer::ToolRuntime` instance, or a `Proc`. Inherited by subclasses. When unset, falls back to `Riffer.config.tool_runtime`. See [Tools — Tool Runtime](04_TOOLS.md#tool-runtime-experimental) for details.
 
 ### guardrail
 
@@ -558,7 +558,7 @@ end
 When an agent receives a response with tool calls:
 
 1. Agent detects `tool_calls` in the assistant message
-2. The configured tool runtime executes the tool calls (sequentially by default, or concurrently with `:threaded`):
+2. The configured tool runtime executes the tool calls (sequentially by default, or concurrently with `Riffer::ToolRuntime::Threaded`):
    - Finds the matching tool class
    - Validates arguments against the tool's parameter schema
    - Calls the tool's `call` method with `context` and arguments
--- a/docs/04_TOOLS.md
+++ b/docs/04_TOOLS.md
@@ -474,16 +474,6 @@ class HttpToolRuntime < Riffer::ToolRuntime
 end
 ```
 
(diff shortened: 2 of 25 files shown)


docs/03_AGENTS.md
```diff
@@ -219,6 +219,22 @@ Using both `of:` and a block raises `Riffer::ArgumentError`. Using `of:` with a
...
+
+```ruby
+class MyAgent < Riffer::Agent
+  model 'openai/gpt-4o'
+  uses_tools [WeatherTool, SearchTool]
+  tool_runtime :threaded
+end
+```
+
+Accepts `:inline`, `:threaded`, a `Riffer::ToolRuntime` instance, or a `Proc`. This setting is **not inherited** by subclasses — each agent class must configure its own tool runtime explicitly. When unset, falls back to `Riffer.config.tool_runtime`. See [Tools — Tool Runtime](04_TOOLS.md#tool-runtime-experimental) for details.
```
Thibaut: question (non-blocking):

> each agent class must configure its own tool runtime explicitly

If an agent class subclasses another class that has `tool_runtime :threaded`, and then doesn't call `tool_runtime`, do we raise an error or fall back to `:inline`? I think the latter would be very unexpected. I don't see a significant downside of inheriting `tool_runtime`. It's the generally-expected behavior for subclassing (especially when we have other DSL methods that inherit like that).

docs/04_TOOLS.md
```diff
@@ -368,6 +368,161 @@ rescue => e
...
+    Riffer::Tools::Response.text(response.body)
+  rescue Riffer::ToolExecutionError => e
+    Riffer::Tools::Response.error(e.message, type: :execution_error)
+  rescue RuntimeError => e
+    Riffer::Tools::Response.error("Error executing tool: #{e.message}", type: :execution_error)
+  end
+end
+```
+
+You can also compose with a custom `Riffer::Runner` for concurrency control:
```
Thibaut: question (blocking): I don't quite grok this use-case and it is the first/only mentions of "runners". Do "runners" need to be part of the public API, or should we keep them internal? I think the latter is probably best for now.

docs/04_TOOLS.md
```diff
@@ -368,6 +368,161 @@ rescue => e
...
+```
+
+### Around-Call Hook
+
+Each tool call is wrapped by the `around_tool_call` instance method, which yields by default. Use the `around_tool_call` class method DSL to define custom behavior via a symbol or block:
+
+```ruby
+# Symbol — delegates to a named instance method (recommended for non-trivial logic):
+class InstrumentedRuntime < Riffer::ToolRuntime
+  around_tool_call :instrument
```
Thibaut: thought (non-blocking): I find it odd/inconsistent that runtime classes both have a DSL available to them as well as the expectation to override certain methods (`dispatch_tool_call`). Since runtimes are a low-level abstraction I would suggest sticking to overriding methods as the only pattern.

lib/riffer/tool_runtime.rb
```diff
@@ -0,0 +1,142 @@
...
+# tool calls are dispatched (e.g., HTTP, gRPC).
+#
+#   runtime = Riffer::ToolRuntime::Inline.new
+#   results = runtime.execute(tool_calls, tools: tools, context: ctx)
+#
+class Riffer::ToolRuntime
+  # +runner+ - the concurrency runner to use for batch execution.
+  #
+  #: (?runner: Riffer::Runner) -> void
+  def initialize(runner: Riffer::Runner::Sequential.new)
```
Thibaut: suggestion (blocking): I think this should be set in `Riffer::ToolRuntime::Inline` and this class should become abstract (`raise NotImplementedError` in intialize). That way a subclass has to be explicit about what runner to run, and subclasses that want to use the sequential runner can (explicitly) subclass `Riffer::ToolRuntime::Inline` if that's the behavior they want.

lib/riffer/runner.rb
```diff
@@ -0,0 +1,24 @@
...
+# Generic concurrency primitive for batch execution.
+#
+# Subclasses implement +map+ to control how items are processed
+# (sequentially, threaded, etc.).
+#
+#   runner = Riffer::Runner::Sequential.new
+#   runner.map([1, 2, 3]) { |n| n * 2 }
+#   # => [2, 4, 6]
+#
+class Riffer::Runner
```
Thibaut, line 13: question (blocking): runner are so small/trivial (one method) that I'd love to see how things look like if we folded them into the runtime. Is a runtime really that different from a runner? If we have to expose runners to the public interface then I think consolidating behind a single class would be beneficial.

Thibaut (commented):
Great work! Some questions/suggestions but very aligned on the idea/approach overall 👍

lib/riffer/runner.rb
cloetzi: Ya, so the idea behind runners is to use them for sub-agents as well. I agree that it shouldn't be a public interface, and I'll remove them from the doc. We might merge it at a later point.

Thibaut: I would leave them out until we actually do sub-agents, then we can assess the abstraction in that context, with much more concrete data to go by.

cloetzi: Hmm, are you OK keeping it in for now? I do expect we'll do a second iteration on runners and tool runtime, and we can tackle it then. 

The main motivation is to introduce the concept and see if it holds up, if not, we'll remove it.

docs/04_TOOLS.md
cloetzi: Yeah, bad doc. I removed it.

docs/04_TOOLS.md
cloetzi: That would simplify a lot of code, let me remove this.

lib/riffer/tool_runtime.rb
cloetzi: [7f4dba4](https://github.com/janeapp/riffer/pull/156/commits/7f4dba43364edcd992b3255aea02e133b5f62c9e)

lib/riffer/runner.rb
Thibaut: How will we see if it holds up?

If it turns out we don't need sub-agents or not that way, I'm worried we'll leave behind a more complicated implementation of runtimes than we have to. Given we are solving the runtime problem for now, I feel it's best to scope the implementation to only that problem and do the simplest thing, but if you feel strongly about it I'm OK to leave it.

docs/04_TOOLS.md
cloetzi: [004fdc5](https://github.com/janeapp/riffer/pull/156/commits/004fdc5cd4c3c89424e6ca1933561653d6d414a2)

docs/04_TOOLS.md
cloetzi: [8334b50](https://github.com/janeapp/riffer/pull/156/commits/8334b509eb4676ddcd91830ef474577f87f54b15)

lib/riffer/runner.rb
cloetzi: We'll see if it holds up when we implement subagents. I know Jake wants it for OSS riffer and I can't see world where we never need it for our use.

I am not sure if I agree with that it is significantly more complicated. As you said, it is a single method right now, and it will be straight forward to re-integrate it back. If we remove it now, my worry is that it is going to be painful to separate the runtime out of the tool runner.

I am happy to admit defeat if this starts getting in the way in subsequent iterations, but right now it is implemented and working, and there's a plan to re-use it for subagents.

docs/03_AGENTS.md
cloetzi: Good point, I initially removed it for simplicity’s sake, but it is not a ton of extra code to remove a potential unexpected behaviour.

[71d85ed](https://github.com/janeapp/riffer/pull/156/commits/71d85eda1594d2e028c05d634280e00900a5882b)

lib/riffer/agent.rb
```diff
@@ -618,16 +618,24 @@ def resolved_tools
...
+      runtime = if config.is_a?(Proc)
+        (config.arity == 0) ? config.call : config.call(@tool_context)
+      else
+        config
+      end
 
-    JSON.parse(arguments, symbolize_names: true)
+      case runtime
+      when :inline then Riffer::ToolRuntime::Inline.new
+      when :threaded then Riffer::ToolRuntime::Threaded.new
```
Thibaut: suggestion (blocking): just realized these make a special case out of the built-in runtimes, and is a pattern we moved away from in the past to have special identifiers for classes. We were able to simplify quite a few things and side-step load-order issues by sticking to classes, so I would do the same here.

```ruby
tool_runtime Riffer::ToolRuntime::Inline
```
is simple enough and points directly to the underlying implementation.

cloetzi: [a96c7d1](https://github.com/janeapp/riffer/pull/156/commits/a96c7d1f3a2a6f6d404b3070e42253f4ab8c6180)

Thibaut (approved):
Thanks for the changes!

PR #158, opened 2026-03-04
feat: add support for dynamic instructions

Thibaut:
- The `instructions` DSL now accepts a Proc for dynamic instructions, matching the existing `model` pattern. The Proc is resolved at generate/stream time and receives `tool_context`.

```ruby
class MyAgent < Riffer::Agent
  model 'openai/gpt-4o'
  instructions ->(ctx) { "You are assisting #{ctx[:name]}" }
end

MyAgent.generate('Hello!', tool_context: { name: 'Jane' })
```

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -16,6 +16,17 @@ agent = EchoAgent.new
 puts agent.generate('Hello world')
 ```
 
+`instructions` also accepts a Proc for dynamic instructions resolved at generate time. The Proc receives the `context` hash:
+
+```ruby
+class PersonalAgent < Riffer::Agent
+  model 'openai/gpt-5-mini'
+  instructions ->(context) { "You are assisting #{context[:name]}" }
+end
+
+PersonalAgent.generate('Hello!', context: { name: 'Jane' })
+```
+
 ### Providers (`lib/riffer/providers/`)
 
 Adapters for LLM APIs. The base class uses a template-method pattern — `generate_text` and `stream_text` orchestrate the flow, delegating to five hook methods each provider implements:
--- a/lib/riffer/agent.rb
+++ b/lib/riffer/agent.rb
@@ -50,11 +50,26 @@ class Riffer::Agent
 
   # Gets or sets the agent instructions.
   #
-  #: (?String?) -> String?
-  def self.instructions(instructions_text = nil)
-    return @instructions if instructions_text.nil?
-    validate_is_string!(instructions_text, "instructions")
-    @instructions = instructions_text
+  # Accepts a static string or a Proc for dynamic instructions.
+  # When a Proc is given, it is called at generate time and receives
+  # the +context+ hash (which may be +nil+).
+  #
+  #   instructions "You are a helpful assistant."
+  #
+  #   instructions -> (context) {
+  #     "You are assisting #{context[:name]}"
+  #   }
+  #
+  #: (?(String | Proc)?) -> (String | Proc)?
+  def self.instructions(instructions_or_proc = nil)
+    return @instructions if instructions_or_proc.nil?
+
+    if instructions_or_proc.is_a?(Proc)
+      @instructions = instructions_or_proc
+    else
+      validate_is_string!(instructions_or_proc, "instructions")
+      @instructions = instructions_or_proc
+    end
   end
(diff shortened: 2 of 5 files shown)


.agents/architecture.md
```diff
@@ -16,6 +16,17 @@ agent = EchoAgent.new
 puts agent.generate('Hello world')
 ```
 
+`instructions` also accepts a Proc for dynamic instructions resolved at generate time. The Proc receives the `tool_context` hash:
+
+```ruby
+class PersonalAgent < Riffer::Agent
+  model 'openai/gpt-5-mini'
+  instructions ->(tool_context) { "You are assisting #{tool_context[:name]}" }
```
Thibaut: I think we should rename `tool_context` to `context`, but will do that separately since it's a breaking change.

cloetzi: thought(blocking): I have some concerns hijacking the tool context for this. With that we will always forward the full context that has been provided for the instructions to each tool call. I could see this being relatively verbose for some more complex agents, and if we are discussing a potential distributed tool invocations, this could lead to unnecessary overhead. I would prefer separating this out into `instructions_context` and `tool_context` and keep the concerns separate.

lib/riffer/agent.rb
```diff
@@ -381,7 +396,8 @@ def track_token_usage(usage)
   #: ((String | Array[Hash[Symbol, untyped] | Riffer::Messages::Base]), ?files: Array[Hash[Symbol, untyped] | Riffer::FilePart]?) -> void
   def initialize_messages(prompt_or_messages, files: nil)
     @messages = []
-    @messages << Riffer::Messages::System.new(@instructions_text) if @instructions_text
+    instructions_text = generate_instructions
+    @messages << Riffer::Messages::System.new(instructions_text) if instructions_text
```
cloetzi, line 400: thought(blocking): This dynamic nature might be tricky to get right - at least with our current use of `riffer`. Currently, riffer is generating this every time as the first system message, no matter what's passed in. That means that the systems instructions could change on first call vs subsequent calls if instruction context changes.
In our current implementation, we are not storing the system message at all, but rely on riffer generating it. As a consequence, this means that we need to store the `instruction_context` in our database (outside the regular message table) to be able to resume a session, but this is exactly what we are trying to avoid.

While we can still add this new feature (I think it is neat), I don't think it is going to be the solution for what we are trying to solve. We still will have to generate a separate system message that holds the context and store it in the database.

.agents/architecture.md
Thibaut: Good thought!

I think separating the context is a slippery slope that will result in a lot of complexity. Multiple methods use this pattern already: `model`, `uses_tools`, `tool_runtime`, and more are on the way (the skills DSL). Guardrails also use `context`.  The client code is going to become complex if every one of these methods ends up with its own context object. 

I don't disagree with the potential pitfall you describe, but I don't think the solution is to create more context objects. In those cases it's inevitable that `tool_context` itself ends up with a litany of objects when it needs to service many tools. IMHO a better solution would be to support per-tool context and/or provide an abstraction that allows the runtime to manipulate the context, but keep a single/simple `context` that is passed around through all the layers of the agent execution.

lib/riffer/agent.rb
Thibaut: > That means that the systems instructions could change on first call vs subsequent calls if instruction context changes.

Wouldn't those "subsequent calls" use `resume` / `resume_stream`, which doesn't re-generate the system prompt? 

> We still will have to generate a separate system message that holds the context and store it in the database.

Why "hold the context"? I think we can simply store the generated instructions to be able to resume off the same system prompt.

I just spotted in Jane that we are actually persisting this system message by manually recreating it through reading `self.class.instructions`. This is a bad pattern and we should let Riffer create the system message (be the source of truth for it). Not sure this requires changes in Riffer, but we can sort that out.

Thibaut: As discussed live, will re-architect `Riffer::Agent`'s public interface to remove the ambiguity around the behavior of "starting with an input" vs. "starting with messages" vs. "resuming with messages" and how/when the system message gets injected.

cloetzi (approved):

PR #159, opened 2026-03-05
refactor!: rename `tool_context` to `context`

Thibaut:
`tool_context` stopped being "context for tools" once we started passing it around for doing dynamic model selection and guardrails. We have even more features in the works that would use this "context" for dynamic behavior, namely dynamic instructions https://github.com/janeapp/riffer/pull/158 and skills https://github.com/janeapp/riffer/pull/151. So I think we're due for renaming this object `context`, to avoid confusion and simplify things. It's a breaking change, but better now than later.

### DIFF
--- a/docs/03_AGENTS.md
+++ b/docs/03_AGENTS.md
@@ -37,12 +37,12 @@ class MyAgent < Riffer::Agent
 end
 ```
 
-When the lambda accepts a parameter, it receives the `tool_context`:
+When the lambda accepts a parameter, it receives the `context`:
 
 ```ruby
 class MyAgent < Riffer::Agent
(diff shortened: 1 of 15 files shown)


docs/03_AGENTS.md
```diff
@@ -283,7 +283,7 @@ response = MyAgent.generate([
 ])
 
 # With tool context
```
cloetzi: issue(non-blocking): `tool context` leftover.

cloetzi (approved):
I think PR title should include the hint for `release please`. Not sure what the tag is to indicate that it is a breaking change in the changelog.

PR #165, opened 2026-03-09
feat: Unify generate/stream API with multi-turn support, remove resume methods

cloetzi:
## Summary

- **Remove `resume` / `resume_stream` / `restore_state`** — these separate methods are replaced by unified behavior in `generate` and `stream`
- **Multi-turn string continuation** — calling `generate("Follow up")` on an agent with existing messages appends the user message and continues the conversation (no system message duplication, pending tool calls executed, step offset preserved)
- **Array input = cross-process resume** — passing an array to a fresh agent uses messages as-is (no system messages prepended). Passing an array to an agent that already has messages raises `Riffer::ArgumentError`
- **Split system messages** — instructions and skills catalog are now two separate system messages instead of one combined message
- **Public `instruction_message` / `skills_message`** — new methods for database persistence workflows where system messages need to be stored independently
- **Per-call state reset** — `context`, tools, tool runtime, model, skills state, and interrupted flag reset on each call. Only message history and cumulative `token_usage` persist across calls. Documentation makes this explicit.

### Input behavior matrix

| Input | Agent state | Behavior |
|---|---|---|
| String | Fresh agent | New conversation (builds system messages + user message) |
| String | Has messages | Continue conversation (appends user message, resume mode) |
| Array | Fresh agent | Cross-process resume (messages used as-is) |
| Array | Has messages | Raises `Riffer::ArgumentError` |

## Test plan

- [x] `bundle exec rake` passes (tests + lint + type checking)
- [ ] Verify multi-turn: `agent.generate("Hello"); agent.generate("Follow up")`
- [ ] Verify cross-process resume: `MyAgent.new.generate(persisted_messages)`
- [ ] Verify guard: `agent.generate("Hello"); agent.generate([...])` raises
- [ ] Verify context must be re-passed: `agent.generate("Hello", context: ctx); agent.generate("More", context: ctx)`

### DIFF
--- a/docs/03_AGENTS.md
+++ b/docs/03_AGENTS.md
@@ -293,6 +293,14 @@ The behavior depends on what you pass and the agent's current state:
 | **Array** | No prior messages | **Restore from persisted data.** Uses the array as-is (no system messages added). Pending tool calls are executed. This is for cross-process resume. |
 | **Array** | Has messages from a prior call | **Raises `Riffer::ArgumentError`.** Use a string to continue, or a new agent instance to start from a persisted array. |
 
+**State reset per call:** Each call to `generate` or `stream` resets `context`, tools, tool runtime, model, skills state, and the interrupted flag. This means `context:` must be passed on every call — it is not carried over from a previous call. The only state that persists across calls is the message history and cumulative `token_usage`.
+
+```ruby
+agent.generate('Hello', context: {user_id: 123})
+agent.generate('Follow up')  # context is nil here — pass it again if needed
+agent.generate('More', context: {user_id: 123})  # context is restored
+```
+
 ```ruby
 # New conversation (class method — recommended for simple calls)
 response = MyAgent.generate('Hello')
--- a/docs/03_AGENTS.md
+++ b/docs/03_AGENTS.md
@@ -282,27 +282,38 @@ See [Guardrails](09_GUARDRAILS.md) for detailed documentation.
 
 ### generate
 
-Generates a response synchronously. Returns a `Riffer::Agent::Response` object:
+Generates a response synchronously. Returns a `Riffer::Agent::Response` object.
+
+The behavior depends on what you pass and the agent's current state:
+
+| Input | Agent state | Behavior |
+|---|---|---|
+| **String** | No prior messages | **New conversation.** Builds system messages (instructions + skills), adds user message, calls the LLM. |
+| **String** | Has messages from a prior call | **Continue conversation.** Appends the user message to the existing history and re-enters the LLM loop. Pending tool calls from a prior interrupt are executed first. |
+| **Array** | No prior messages | **Restore from persisted data.** Uses the array as-is (no system messages added). Pending tool calls are executed. This is for cross-process resume. |
+| **Array** | Has messages from a prior call | **Raises `Riffer::ArgumentError`.** Use a string to continue, or a new agent instance to start from a persisted array. |
 
 ```ruby
-# Class method (recommended for simple calls)
+# New conversation (class method — recommended for simple calls)
 response = MyAgent.generate('Hello')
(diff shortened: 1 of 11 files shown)


lib/riffer/agent.rb
```diff
@@ -297,51 +299,43 @@ def stream(prompt_or_messages, files: nil, context: nil)
...
-
-    restore_state(messages: messages, context: context)
-
-    Enumerator.new do |yielder|
-      run_stream_loop(yielder, resume: true)
-    end
+  # Returns +nil+ when no instructions are configured.
+  #
+  #: (?context: Hash[Symbol, untyped]?) -> Riffer::Messages::System?
+  def instruction_message(context: nil)
```
Thibaut: Just a suggestion in passing before I do a full review later: call this `generate_instruction_message` (and `generate_skills_message`) to make it clear this is constructing a new message each time, not setting it or anything like that.

lib/riffer/agent.rb
```diff
@@ -297,51 +299,43 @@ def stream(prompt_or_messages, files: nil, context: nil)
...
-    restore_state(messages: messages, context: context)
-
-    Enumerator.new do |yielder|
-      run_stream_loop(yielder, resume: true)
-    end
+  # Returns +nil+ when no instructions are configured.
+  #
+  #: (?context: Hash[Symbol, untyped]?) -> Riffer::Messages::System?
+  def generate_instruction_message(context: nil)
+    build_instruction_message(context)
```
tsov: Bug: `context` here is the keyword argument value (defaulting to `nil`), but `build_instruction_message` takes a positional argument with `= @context` as its default. Ruby only applies the positional default when the argument is **omitted entirely** — passing `nil` explicitly means it receives `nil`, not `@context`.

So `generate_instruction_message(context: {name: "Alice"})` will pass `nil` through to `build_instruction_message(nil)`, ignoring the caller's context.

Fix:
```ruby
def generate_instruction_message(context: nil)
  build_instruction_message(context || @context)
end
```

docs/03_AGENTS.md
```diff
@@ -282,27 +282,46 @@ See [Guardrails](09_GUARDRAILS.md) for detailed documentation.
...
+Generates a response synchronously. Returns a `Riffer::Agent::Response` object.
+
+The behavior depends on what you pass and the agent's current state:
+
+| Input | Agent state | Behavior |
+|---|---|---|
+| **String** | No prior messages | **New conversation.** Builds system messages (instructions + skills), adds user message, calls the LLM. |
+| **String** | Has messages from a prior call | **Continue conversation.** Appends the user message to the existing history and re-enters the LLM loop. Pending tool calls from a prior interrupt are executed first. |
+| **Array** | No prior messages | **Restore from persisted data.** Uses the array as-is (no system messages added). Pending tool calls are executed. This is for cross-process resume. |
+| **Array** | Has messages from a prior call | **Raises `Riffer::ArgumentError`.** Use a string to continue, or a new agent instance to start from a persisted array. |
```
bottrall: **Nitpick(non-blocking):** Run prettier on this if you can.

lib/riffer/agent.rb
```diff
@@ -273,7 +274,7 @@ def generate(prompt_or_messages, files: nil, context: nil)
     all_modifications.concat(modifications)
     return build_response("", tripwire: tripwire, modifications: all_modifications) if tripwire
 
-    run_generate_loop(all_modifications)
+    run_generate_loop(all_modifications, resume: continuation)
```
bottrall: **Suggestion(non-blocking):** We could probably remove the `resume` argument from `run_generate_loop`.



It's only used for `execute_pending_tool_calls`, which no-ops efficiently if there's nothing needed to be done.

bottrall (approved):

PR #167, opened 2026-03-16
feat: Add Azure OpenAI provider

bottrall:
## Summary

- Add `Riffer::Providers::AzureOpenAI`, a thin subclass of the OpenAI provider that configures Azure-specific endpoint and API key
- Credentials resolve in order: keyword arguments → `Riffer.config.azure_openai` → `AZURE_OPENAI_API_KEY` / `AZURE_OPENAI_ENDPOINT` env vars
- Register `azure_openai` in the provider repository, add config struct, Zeitwerk inflection, and VCR sensitive data filters
- Add tests covering generate, stream, structured output, tool calling, and token usage with VCR cassettes
- Add `docs_providers/05_AZURE_OPENAI.md` documentation and renumber Mock/Custom provider docs to 06/07


docs_providers/05_AZURE_OPENAI.md
```diff
@@ -0,0 +1,165 @@
...
+```ruby
+agent.stream("Explain Ruby blocks").each do |event|
+  case event
+  when Riffer::StreamEvents::TextDelta
+    print event.content
+  end
+end
+```
+
+## Tool Calling
```
Thibaut, line 120: suggestion (non-blocking): I cannot spot differences between providers on tool calling. We just mention that "provider converts tools to [provider format]" followed by an example that has no provider-specific logic. Should we remove these sections from the docs to keep them lean?

Thibaut (approved):

docs_providers/05_AZURE_OPENAI.md
bottrall: I have plans to do a big pass over the docs in the coming weeks as there's many areas we can improve. I'll note this down for then, but for now lets go with consistency.

PR #169, opened 2026-03-16
feat: add subagent support for multi-agent orchestration

bottrall:
## Summary

- Add multi-agent orchestration where agents can delegate tasks to specialized subagents via `uses_agents`
- New DSL methods on Agent: `description`, `uses_agents`, `agent_runtime`
- New classes: `AgentTool` (schema-only tool factory), `AgentRuntime` with `Inline` and `Threaded` implementations
- Context propagates automatically from the calling agent to subagents
- Safety checks: self-reference prevention, name conflict detection, description requirement
- Detect indirect circular agent delegation at runtime (A → B → A) by tracking the agent call stack via `context[:_agent_stack]`, returning a `:circular_delegation` error instead of a `SystemStackError`
- Refactor: replace mutable `seed_agent_stack` with pure `agent_dispatch_context` — `@context` is no longer mutated after `prepare_run`
- Docs: new `14_AGENT_ORCHESTRATION.md` as the canonical source for subagent docs; other docs link to it with minimal examples to avoid duplication

## Test plan

- [x] `bundle exec rake` passes — 1207 tests, 0 failures, no lint violations, no type errors
- [x] Unit test: circular delegation detected when agent is already in call stack
- [x] Integration test: A → B → A cycle returns error response, not stack overflow
- [x] Context propagation still works (includes `_agent_stack`)


bottrall:
@cloetzi hold off on reviewing. @claude raises a good point that I'll need to address.

bottrall:
@cloetzi good to go 👍🏻

docs/03_AGENTS.md
```diff
@@ -591,18 +645,142 @@ Riffer::Agent.all.each do |agent_class|
...
+- **RuntimeError** — Exceptions are caught and returned as error responses
+
+### Constraints
+
+- An agent cannot use itself as a subagent (raises `Riffer::ArgumentError`)
+- Circular agent delegation is detected at runtime — if Agent A delegates to Agent B which delegates back to Agent A, the cycle is caught and returned as an error response to the calling agent (rather than causing a stack overflow)
+- Subagent tool names must not conflict with regular tool names (raises `Riffer::ArgumentError`)
+- Subagents must have a `description` (raises `Riffer::ArgumentError`)
+
+### Circular Delegation
```
cloetzi: nitpick(non-blocking): That's kinda duplicate information. The bullet point above explains the same but more concise.

docs/07_CONFIGURATION.md
```diff
@@ -92,6 +92,26 @@ end
...
+Riffer.configure do |config|
+  config.agent_runtime = Riffer::AgentRuntime::Threaded
+end
+```
+
+| Value                           | Description                                                                                         |
+| ------------------------------- | --------------------------------------------------------------------------------------------------- |
+| `Riffer::AgentRuntime` subclass | Instantiated automatically (e.g., `Riffer::AgentRuntime::Inline`, `Riffer::AgentRuntime::Threaded`) |
+| `Riffer::AgentRuntime` instance | Custom runtime with specific options                                                                |
+| `Proc`                          | Dynamic resolution                                                                                  |
```
cloetzi: suggestion(non-blocking): The description should probably contain what should be returned in the `proc`.

lib/riffer/agent_runtime.rb
```diff
@@ -0,0 +1,101 @@
...
+        type: :execution_error
+      )
+    elsif response.interrupted?
+      Riffer::Tools::Response.text("Agent was interrupted: #{response.content}")
+    else
+      Riffer::Tools::Response.text(response.content)
+    end
+  rescue RuntimeError => e
+    Riffer::Tools::Response.error("Error executing agent: #{e.message}", type: :execution_error)
+  end
```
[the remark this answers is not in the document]
bottrall: Fixed in c2e7f90. Changed `rescue RuntimeError` to `rescue => e` (implicit `StandardError`), which catches `JSON::ParserError`, `TypeError`, and any other `StandardError` subclass. Kept it simple with a single rescue clause rather than splitting into multiple — any failure in `dispatch_agent_call` should return a structured error response regardless of exception type.

test/riffer/agent_runtime_test.rb
```diff
@@ -0,0 +1,250 @@
...
+      Riffer::Messages::Assistant::ToolCall.new(
+        id: i.to_s, call_id: "cid_#{i}", name: "agent__parallel-agent", arguments: '{"message":"test"}'
+      )
+    end
+
+    results = runtime.execute(tool_calls, agents: {"agent__parallel-agent" => parallel_agent}, context: nil)
+
+    expect(results.length).must_equal 3
+    expect(seen.uniq.length).must_equal 3
+  end
```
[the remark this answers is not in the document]
bottrall: Fixed in c2e7f90. Changed to `must_be :>, 1` as suggested.

lib/riffer/agent_tool.rb
```diff
@@ -0,0 +1,48 @@
...
+    "#{AGENT_PREFIX}#{agent_class.identifier.gsub("/", "__")}"
+  end
+
+  # Builds a Riffer::Tool subclass for the given agent class.
+  #
+  # Raises Riffer::ArgumentError if the agent has no description.
+  #
+  #: (singleton(Riffer::Agent)) -> singleton(Riffer::Tool)
+  def self.build(agent_class)
+    raise Riffer::ArgumentError, "Agent #{agent_class} must have a description to be used as a subagent" unless agent_class.description
```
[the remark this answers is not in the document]
bottrall: Skipping this one. The `description` DSL setter already runs through `validate_is_string!` which rejects non-string values. An empty string is a valid (if unhelpful) description — enforcing content quality at this level would be over-engineering for an internal DSL.

lib/riffer/agent.rb
```diff
@@ -677,6 +744,27 @@ def resolved_tools
...
+  #: () -> Array[singleton(Riffer::Agent)]?
+  def resolve_agents_config
+    config = self.class.uses_agents
+    return nil if config.nil?
+
+    if config.is_a?(Proc)
+      (config.arity == 0) ? config.call : config.call(@context)
+    else
+      config
+    end
```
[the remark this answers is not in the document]
bottrall: Skipping this one. `uses_agents` is internal DSL, not a public API boundary. If someone passes a bad value, the resulting `NoMethodError` is clear enough and points directly at the problem. Adding validation here would be defensive coding against a scenario that only gem authors (not end users) would hit.

lib/riffer/agent.rb
```diff
@@ -696,6 +784,25 @@ def resolve_tool_runtime
...
+      runtime = if config.is_a?(Proc)
+        (config.arity == 0) ? config.call : config.call(@context)
+      else
+        config
+      end
+
+      case runtime
+      when Class then runtime.new
+      when Riffer::AgentRuntime then runtime
+      else raise Riffer::ArgumentError, "Invalid agent_runtime: #{runtime.inspect}"
```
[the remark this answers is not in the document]
bottrall: Skipping this one. The existing `resolve_tool_runtime` method uses the same `when Class then runtime.new` pattern without an inheritance guard. Changing one without the other would be inconsistent. If we decide to add inheritance checks, it should be done for both in a separate pass.

lib/riffer/agent.rb
```diff
@@ -548,17 +586,14 @@ def has_tool_calls?(response)
...
 
-    results.each do |tool_call, result|
-      add_message(Riffer::Messages::Tool.new(
-        result.content,
-        tool_call_id: tool_call.id,
-        name: tool_call.name,
-        error: result.error_message,
-        error_type: result.error_type
-      ))
+    if agent_names.empty?
```
cloetzi: suggestion(non-blocking): Claude pointed out that this is duplicate with the code on line 616, given the new complexity, I think that could be extracted into a helper method.

lib/riffer/agent.rb
```diff
@@ -578,17 +613,14 @@ def execute_pending_tool_calls
...
+      dispatch_tool_results(resolve_tool_runtime.execute(pending, tools: resolved_tools, context: @context))
+    else
+      agent_calls, tool_calls = pending.partition { |tc| agent_names.include?(tc.name) }
```
cloetzi: thought(non-blocking): Wouldn't a split based on base-class be more clean here?

This made me think more about @Thibaut's thought of sub-agents are just tools, I guess we could acheive all the same functionality by:
1. Agent classes can be just added to the `Tool` list as is (assuming they have a description).
2. We automatically convert them to `AgentTool` as we do now if we detect the tool inherits from the Agent chain.
3. We do this exact split here based on base-class.

This would satisfy both "requirements":
- Everything is a tool.
- Agents can have a different runtime than tools.

One downside of the runtime separation is that parallel execution of sub-agents and tools is not straight forward, and I am not quite sure yet how we'd handle this in the context of Sidekiq ping-pong, but I don't think that's relevant right now. Also, we'll be able to find a solution.

lib/riffer/agent.rb
```diff
@@ -677,6 +744,27 @@ def resolved_tools
...
+  def resolved_agent_map
+    @resolved_agent_map ||= begin
+      agent_classes = resolve_agents_config
+      return {} unless agent_classes&.any?
+      agent_classes.to_h { |ac| [Riffer::AgentTool.identifier_for(ac), ac] }
+    end
+  end
+
+  #: () -> Array[singleton(Riffer::Agent)]?
+  def resolve_agents_config
```
cloetzi, line 804: suggestion(non-blocking): I think this could be memoized as well alongside our memoize zoo.

lib/riffer/agent.rb
```diff
@@ -142,6 +154,32 @@ def self.tool_runtime(value = nil)
     end
   end
 
+  # Gets or sets the subagents used by this agent.
+  #
+  #: (?(Array[singleton(Riffer::Agent)] | Proc)?) -> (Array[singleton(Riffer::Agent)] | Proc)?
+  def self.uses_agents(agents_or_lambda = nil)
```
cloetzi, line 171: suggestion(non-blocking): I think this would benefit from additional validation. Claude pointed out that this would raise during runtime if garbage is passed in (e.g. single agent class over Array of agent classes).

cloetzi (commented):
Overall looks good, left some minor suggestions. Also, as you said, we should align with @Thibaut before merging. I left a comment around how we could possibly compromise.

lib/riffer/agent.rb
bottrall: We're going to have to talk about this more. But I'm not sold on this "everything is a tool" idea. Technically speaking, yes, everything an LLM interacts with has to be a tool from the providers API perspective. But this isn't how it's framed in any other library I know of and it's not how any product (for example claude code) the average developer uses represents them either. Which is why I think there's value in presenting them as two different concepts, it's more inline with the rest of the industry.

docs/03_AGENTS.md
bottrall: Addressed in 98e517b — removed the duplicate subsection.

docs/07_CONFIGURATION.md
bottrall: Addressed in 98e517b — updated both Proc rows to: "Receives context, must return a runtime subclass or instance".

lib/riffer/agent.rb
bottrall: Addressed in e130606 — extracted a `dispatch_calls` private method that both callers delegate to.

lib/riffer/agent.rb
bottrall: Addressed in e130606 — `resolve_agents_config` is now memoized with `@resolved_agents_config`.

lib/riffer/agent.rb
bottrall: Wasn't sold on this one, simply for the consistency argument. We're not validating on `uses_tools` and I don't want to add that to the scope of this PR. It raises a bigger question of "should we by default add validation to all publicly documented methods?". That was a stance I took early on in this project but we've drifted away from that standard.

lib/riffer/agent.rb
cloetzi: Ya, I agree, I am not sold either. I was mainly coming from the perspective that _if_ we decide to go that route, we can still support separate runtimes for tools and agents with a relatively clean implementation.

PR #178, opened 2026-03-25
feat: add fibers tool runtime using async gem

bottrall:
## Summary

- Add `Riffer::ToolRuntime::Fibers` — a new built-in tool runtime that executes tool calls concurrently using fibers via the `async` gem
- Add `Riffer::Runner::Fibers` — the underlying runner that uses `Async::Barrier` for concurrency and optional `Async::Semaphore` for limiting max concurrency
- No concurrency limit by default (fibers are lightweight); users can opt in via `max_concurrency:`
- Uses `Riffer::Helpers::Dependencies#depends_on` for lazy loading the `async` gem with clear error messages if missing
- Pin `async < 2.38` and `io-event < 1.14` to maintain Ruby 3.2 compatibility
- Fix missing `context:` keyword on `Fibers#map` to match the updated runner interface
- Fix RDoc formatting to match conventions from #182: labeled list params, `#--` stop directives, `<tt>` for multi-word inline code

## Test plan

- [x] `Riffer::Runner::Fibers` tests: order preservation, concurrent execution, max_concurrency enforcement, empty items, exception propagation
- [x] `Riffer::ToolRuntime::Fibers` test: concurrent tool call execution verified via fiber identity tracking
- [x] All existing tests pass (`bundle exec rake` — 1170 runs, 1795 assertions, 0 failures)
- [x] Type checking passes (`bundle exec steep check`)
- [x] RBS signatures generated

### DIFF
--- a/lib/riffer/runner/fibers.rb
+++ b/lib/riffer/runner/fibers.rb
@@ -4,7 +4,7 @@
 # Processes items concurrently using fibers via the +async+ gem.
 #
 # All items run as fibers simultaneously by default. When
-# +max_concurrency+ is set, an +Async::Semaphore+ limits how many
+# +max_concurrency+ is set, an <tt>Async::Semaphore</tt> limits how many
 # fibers execute at once.
 #
 # If multiple fibers raise, only the first exception is re-raised
@@ -16,9 +16,10 @@
 class Riffer::Runner::Fibers < Riffer::Runner
   include Riffer::Helpers::Dependencies
 
-  # +max_concurrency+ - maximum number of fibers to run simultaneously.
+  # [max_concurrency] maximum number of fibers to run simultaneously.
   #   When +nil+, all fibers run without limit.
   #
+  #--
   #: (?max_concurrency: Integer?) -> void
   def initialize(max_concurrency: nil)
     depends_on "async"
@@ -26,6 +27,7 @@ class Riffer::Runner::Fibers < Riffer::Runner
     @max_concurrency = max_concurrency
   end
 
+  #--
   #: (Array[untyped], context: Hash[Symbol, untyped]?) { (untyped) -> untyped } -> Array[untyped]
   def map(items, context:, &block)
     return [] if items.empty?
--- a/lib/riffer/runner/fibers.rb
+++ b/lib/riffer/runner/fibers.rb
@@ -26,8 +26,8 @@ class Riffer::Runner::Fibers < Riffer::Runner
     @max_concurrency = max_concurrency
   end
 
-  #: (Array[untyped]) { (untyped) -> untyped } -> Array[untyped]
-  def map(items, &block)
+  #: (Array[untyped], context: Hash[Symbol, untyped]?) { (untyped) -> untyped } -> Array[untyped]
+  def map(items, context:, &block)
     return [] if items.empty?
 
     results = Array.new(items.size)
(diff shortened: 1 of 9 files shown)


riffer.gemspec
```diff
@@ -44,6 +44,8 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "anthropic", "~> 1.25.0"
   spec.add_development_dependency "aws-sdk-bedrockruntime", "~> 1.42"
   spec.add_development_dependency "openai", "~> 0.56.0"
+  spec.add_development_dependency "async", "~> 2.25", "< 2.38"
+  spec.add_development_dependency "io-event", "< 1.14"
```
Thibaut, line 48: question (non-blocking): why maintain Ruby 3.2 compatibility? This has the hallmarks of something we will forget to revisit/update later 😅

lib/riffer/runner/fibers.rb
```diff
@@ -0,0 +1,60 @@
...
+class Riffer::Runner::Fibers < Riffer::Runner
+  include Riffer::Helpers::Dependencies
+
+  # +max_concurrency+ - maximum number of fibers to run simultaneously.
+  #   When +nil+, all fibers run without limit.
+  #
+  #: (?max_concurrency: Integer?) -> void
+  def initialize(max_concurrency: nil)
+    depends_on "async"
+    require "async/semaphore" if max_concurrency
```
Thibaut, line 26: issue (blocking): I think these should be done at load time. `tool_runtime` takes a Proc and digging into `Gem` and `Bundler` at runtime is, I think, very suboptimal. It could spike memory usage or generally do some slow stuff we wouldn't want to run every time. I think the "trick" would be to not require this file in Riffer by default. Let the client require it if they want it, then the gem dependency gets resolved at load time.

Thibaut (commented):
Overall :+1: :+1:

riffer.gemspec
bottrall: I've got a reminder in my calendar. Support for 3.2 will be dropped on April 1st. At which point dependabots PRs to update these dependencies will start passing CI 😄

lib/riffer/runner/fibers.rb
bottrall: I see where you're coming from. But at the same time this is how riffer already works for providers. We do the exact same thing every time an llm provider is invoked. My understanding is that this is a common pattern for gems and that calling `require` on a gem that is already loaded is akin to a no-op. So yes there _may_ be a performance hit the first time, but I'm skeptical of how much of a problem this is in practice.

Thibaut: The `require` is cheap. I was more concerned about the dependency search we do in `depends_on`. It's probably insignificant, but in my experience loading/checking gem dependencies at load time is an anti-pattern.

Many gems (for example, `active_support`) don't auto-load all their code as a way to let the client optimize what to load, and this also allows for load-time dependency checking. It's arguable a little more dev-friendly to find out "you can't use this without that" at load time than run time.

In a Rails app, how this could work is we drop a `require "riffer/fibers"` (or `require "riffer/async"`) in `config/initializers/riffer.rb` and by then the `async` dependency would already been loaded in (via Bundler, or in the same initializer). Same for the providers.

This isn't the pattern already in place and so out-of-scope of this PR, but just thought to mention it in case you see value in it. Nothing worth prioritizing but maybe something to consider for later.

Thibaut (approved):

PR #180, opened 2026-03-25
MCP Support

kepheusikarius:
## Summary

This PR adds first-class [Model Context Protocol](https://modelcontextprotocol.io) support so Riffer agents can use tools exposed by remote MCP servers without hand-written tool classes per endpoint.

It delivers:

- Global server registration (`Riffer::Mcp.register`) with async `tools/list` discovery on a background thread  
- Agent opt-in via `use_mcp` (tag-based matching; not inherited, same idea as `uses_tools`)  
- Thin HTTP client wrapper around the `mcp` gem (`Riffer::Mcp::Client`) plus dynamic `Riffer::Tool` subclasses from MCP definitions (`ToolFactory`)  
- Optional per-run `tools/call` headers via `Riffer.config.mcp.credentials`, with `AuthenticatedTool` wrappers when that proc is set  
- Configurable behaviour while discovery is still running (`on_pending`: `:ignore`, `:wait`, `:raise`; `wait_timeout`)  
- Duplicate tool name detection across static tools + all MCP sources  
- Docs (`docs/10_MCP.md`), configuration section in `docs/07_CONFIGURATION.md`, agent DSL note in `docs/03_AGENTS.md`, README link, architecture notes in `.agents/architecture.md`  
- Tests under `test/riffer/mcp/` and agent/config coverage  
- Runtime deps: `mcp` (~> 0.8), `faraday` (>= 1.0) for HTTP transport  

---

## Architecture

```mermaid
flowchart LR
  A["Application"] --> B["Riffer::Mcp.register"]
  B --> C["Registry"]
  C --> D["Registration"]
  D --> E["Background thread: tools/list"]
  E --> F["ToolFactory → Riffer::Tool subclasses"]
  F --> G["AgentFactory → container Agent"]
  H["Riffer::Agent subclass"] --> I["use_mcp :tag"]
  I --> J["resolved_tools"]
  J --> K["uses_tools + MCP tools"]
  K --> L{"credentials proc?"}
  L -->|no| M["Shared Client from discovery_headers"]
  L -->|yes| N["AuthenticatedTool → Client per call"]
  M --> O["MCP tools/call"]
  N --> O
```

---

## Discovery and pending behaviour

```mermaid
flowchart TD
  A["Agent resolves tools"] --> B["For each use_mcp tag"]
  B --> C["Registry.find_by_tags"]
  C --> D{"Registration ready?"}
  D -->|yes| E["Include tools"]
  D -->|no| F{"on_pending"}
  F -->|:ignore| G["Skip server"]
  F -->|:wait| H["wait_until_ready! or re-raise discovery error"]
  F -->|:raise| I["NotReadyError or re-raise discovery error"]
  H --> E
```

---

## Public API

Register once (returns immediately; discovery continues async):

```ruby
Riffer::Mcp.register(
  name: "github",
  tags: [:github],
  endpoint: "https://mcp.github.com",
  discovery_headers: -> { {"Authorization" => "Bearer #{ENV["GITHUB_TOKEN"]}"} }
)
```

Opt an agent in by tag (multiple `use_mcp` lines accumulate; MCP tools append after `uses_tools`):

```ruby
class ResearchAgent < Riffer::Agent
  model "openai/gpt-4o"
  use_mcp :github
  use_mcp :jira, on_pending: :wait
end
```

Optional session credentials for `tools/call` (omit a server for a run when the proc returns `nil` at resolve time; deny at call time → `CredentialsDeniedError`):

```ruby
Riffer.configure do |config|
  config.mcp.on_pending = :wait
  config.mcp.wait_timeout = 30
  config.mcp.credentials = lambda do |manifest:, matched_tags:, context:|
    {"Authorization" => "Bearer #{token_for(context)}"}
  end
end
```

Introspection / lifecycle:

```ruby
Riffer::Mcp.registrations["github"] # => Registration
Riffer::Mcp.unregister("github")
```

---

## Key types (`lib/riffer/mcp/`)

| Piece | Role |
|--------|------|
| `Manifest` | Name, tags, endpoint, `discovery_headers`, optional `credentials_scope` hint |
| `Registry` | Thread-safe global map name → `Registration`; `find_by_tags` |
| `Registration` | Discovery thread, `ready?`, `wait_until_ready!`, `tools`, `agent`, `discovery_error` |
| `Client` | `tools_list` / `tools_call` via `mcp` gem HTTP transport |
| `ToolFactory` | Anonymous `Riffer::Tool` classes delegating to client |
| `AgentFactory` | Anonymous `Riffer::Agent` container for a server’s tools |
| `AuthenticatedTool` | Wraps tools when `config.mcp.credentials` is set |

---

## Error model

| Class | When |
|--------|------|
| `Riffer::Mcp::NotReadyError` | `on_pending: :raise` while discovery still in flight |
| `Riffer::Mcp::TimeoutError` | `on_pending: :wait` exceeded `wait_timeout` |
| `Riffer::Mcp::CredentialsDeniedError` | `credentials` returned `nil` during `tools/call` |
| Original exception | Re-raised from `discovery_error` when `:wait` / `:raise` applies |

All extend `Riffer::Mcp::Error` → `Riffer::Error`.

---

## Limitations (v1)

- Tool results are normalized to **joined text** from MCP `content`; non-text parts are not surfaced.  
- With `credentials` set, tool calls may use a **new HTTP client per invocation** (no pooling in this release).  

Details: [`docs/14_MCP.md`](docs/14_MCP.md).

---

## Documentation

- [`docs/14_MCP.md`](docs/14_MCP.md) — full MCP guide  
- [`docs/07_CONFIGURATION.md`](docs/07_CONFIGURATION.md) — `config.mcp`  
- [`docs/03_AGENTS.md`](docs/03_AGENTS.md) — `use_mcp`  
- [`README.md`](README.md) — doc index link  
- [`.agents/architecture.md`](.agents/architecture.md) — MCP integration overview  

---

## Generated artifacts

- Regenerated / updated inline RBS under `sig/generated` for `Agent` and MCP-related surfaces (run `bundle exec rake rbs:generate` if types drift).

---

## Validation

```text
bundle exec rake
```

(Adjust the line above with your branch’s actual test/assertion counts after a local run.)

---

## Notes

- MCP is integrated into **tool resolution** for `Riffer::Agent`; it does not change core text/streaming LLM APIs beyond merging MCP tools into `resolved_tools`.  
- Tags are **application-defined**; multi-tenant and user scoping belong in `context` + the `credentials` proc, not on the manifest.  

## Summary by CodeRabbit

* **New Features**
  * MCP integration: agents can opt-in to discover and use tools from external MCP servers by tag, with per-call credentials support and configurable discovery behavior (global and per-use pending strategies, defaulting to ignore).

* **Documentation**
  * New MCP integration guide and updates to agent/configuration docs; README links to MCP docs; example model key updated to openai/gpt-5-mini.

* **Tests**
  * Comprehensive MCP-focused test coverage added.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -132,9 +132,35 @@ Built-in runtimes:
 
 Context flow: `Agent#execute_tool_calls` → `ToolRuntime#execute(tool_calls, tools:, context:)` → `Runner#map(tool_calls, context:) { dispatch }` → `Tool#call(context:, **args)`
 
+### MCP Integration (`lib/riffer/mcp/`)
+
+Register third-party MCP servers globally; agents opt-in by tag via `use_mcp`. Tags are application-defined (manifests may list several; any overlap with `use_mcp` opts in—see `docs/14_MCP.md`).
+
+```ruby
+Riffer::Mcp.register(
+  name: "github",
+  tags: [:github],
+  endpoint: "https://mcp.github.com",
+  discovery_headers: -> { {"Authorization" => "Bearer #{ENV['GITHUB_TOKEN']}"} }
+)
+
+# Optional: per-run tools/call headers (see docs/14_MCP.md)
+Riffer.configure { |c| c.mcp.credentials = ->(manifest:, matched_tags:, context:) { ... } }
+
+class ResearchAgent < Riffer::Agent
+  model "openai/gpt-5-mini"
+  use_mcp :github                      # picks up any :github-tagged registration
+  use_mcp :search, on_pending: :wait   # per-call override
+end
+```
+
+Key types: `Manifest` (`discovery_headers`, optional `credentials_scope` hint), `Registry` (thread-safe store), `Registration` (spawns discovery thread → `ToolFactory`), `Client` (wraps `mcp` gem), `AuthenticatedTool` (wraps MCP tools when `credentials` proc is set).
+
+**on_pending strategies** (global default `:ignore`): `:ignore` skips the server; `:wait` blocks until ready, re-raises failed discovery immediately, or times out; `:raise` re-raises failed discovery or `NotReadyError` while still pending.
+
 ## Key Patterns
 
-- Model config accepts a `provider/model` string (e.g., `openai/gpt-4`) or a Proc/lambda that returns one
+- Model config accepts a `provider/model` string (e.g., `openai/gpt-5-mini`) or a Proc/lambda that returns one
 - Configuration via `Riffer.configure { |c| c.openai.api_key = "..." }`
 - Providers use `depends_on` helper for runtime dependency checking
 - Zeitwerk for autoloading - file structure must match module/class names
@@ -199,6 +225,14 @@ lib/
       reasoning_done.rb  # Reasoning done event
       web_search_status.rb # Web search status event
       web_search_done.rb   # Web search done event
+    mcp.rb               # MCP public API + error classes
+    mcp/
+      manifest.rb        # Server config value object
+      registry.rb        # Thread-safe global store
+      registration.rb    # Per-server state + discovery thread
+      client.rb          # Thin mcp gem wrapper
+      authenticated_tool.rb  # Wraps MCP tools when credentials proc is configured
+      tool_factory.rb    # Generates Riffer::Tool subclasses from MCP tools
 test/
   test_helper.rb         # Minitest configuration with VCR
   riffer_test.rb         # Main module tests
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -52,6 +52,12 @@ GEM
     dotenv (3.2.0)
     drb (2.2.3)
     erb (6.0.4)
+    faraday (2.14.1)
+      faraday-net_http (>= 2.0, < 3.5)
+      json
+      logger
+    faraday-net_http (3.4.2)
+      net-http (~> 0.5)
     ffi (1.17.4)
     ffi (1.17.4-aarch64-linux-gnu)
     ffi (1.17.4-aarch64-linux-musl)
@@ -94,6 +100,9 @@ GEM
       reline (>= 0.4.2)
     jmespath (1.6.2)
     json (2.19.4)
+    json-schema (6.2.0)
+      addressable (~> 2.8)
+      bigdecimal (>= 3.1, < 5)
     language_server-protocol (3.17.0.5)
     lint_roller (1.1.0)
     listen (3.10.0)
@@ -102,12 +111,16 @@ GEM
       rb-inotify (~> 0.9, >= 0.9.10)
     logger (1.7.0)
     lumberjack (1.4.2)
+    mcp (0.8.0)
+      json-schema (>= 4.1)
     method_source (1.1.0)
     metrics (0.15.0)
     minitest (6.0.5)
       drb (~> 2.0)
       prism (~> 1.5)
     nenv (0.3.0)
+    net-http (0.9.1)
+      uri (>= 0.11.1)
     notiffany (0.1.3)
       nenv (~> 0.1)
       shellany (~> 0.0)
@@ -237,10 +250,12 @@ DEPENDENCIES
   async (~> 2.25, < 2.40)
   aws-sdk-bedrockruntime (~> 1.42)
   dotenv
+  faraday (>= 1.0)
   guard
   guard-shell
   io-event (< 1.16)
   irb
+  mcp (~> 0.8)
   minitest (~> 6.0)
   openai (~> 0.59.0)
   rake (~> 13.0)
--- a/docs/10_CONFIGURATION.md
+++ b/docs/10_CONFIGURATION.md
@@ -38,16 +38,13 @@ For provider credentials and setup, see the individual [Provider guides](provide
 
 Optional settings for [MCP server integrations](14_MCP.md):
 
-| Option          | Description                                                                                                     |
-| --------------- | --------------------------------------------------------------------------------------------------------------- |
-| `on_pending`    | `:ignore` (default), `:wait`, or `:raise` when tool discovery is not finished yet                               |
-| `wait_timeout`  | Seconds to wait when `on_pending` is `:wait` (default `10`)                                                     |
-| `credentials`   | Optional `Proc` for per-run `tools/call` HTTP headers: `->(manifest:, matched_tags:, context:) { Hash or nil }` |
+| Option             | Description                                                                                                     |
+| ------------------ | --------------------------------------------------------------------------------------------------------------- |
+| `credentials`      | Optional `Proc` for per-run `tools/call` HTTP headers: `->(manifest:, matched_tags:, context:) { Hash or nil }` |
+| `discovery_runner` | `Riffer::Runner` instance for tool discovery (default `Runner::Sequential.new`)                                 |
 
 ```ruby
 Riffer.configure do |config|
-  config.mcp.on_pending = :wait
-  config.mcp.wait_timeout = 30
   config.mcp.credentials = lambda do |manifest:, matched_tags:, context:|
     {"Authorization" => "Bearer #{token_for(context)}"}
   end
--- a/docs/10_CONFIGURATION.md
+++ b/docs/10_CONFIGURATION.md
@@ -34,6 +34,28 @@ Riffer.config.anthropic.api_key
 
 For provider credentials and setup, see the individual [Provider guides](providers/).
 
+### MCP (Model Context Protocol)
+
+Optional settings for [MCP server integrations](14_MCP.md):
+
+| Option          | Description                                                                                                     |
+| --------------- | --------------------------------------------------------------------------------------------------------------- |
+| `on_pending`    | `:ignore` (default), `:wait`, or `:raise` when tool discovery is not finished yet                               |
+| `wait_timeout`  | Seconds to wait when `on_pending` is `:wait` (default `10`)                                                     |
+| `credentials`   | Optional `Proc` for per-run `tools/call` HTTP headers: `->(manifest:, matched_tags:, context:) { Hash or nil }` |
+
+```ruby
+Riffer.configure do |config|
+  config.mcp.on_pending = :wait
+  config.mcp.wait_timeout = 30
+  config.mcp.credentials = lambda do |manifest:, matched_tags:, context:|
+    {"Authorization" => "Bearer #{token_for(context)}"}
+  end
+end
+```
+
+See [MCP](14_MCP.md) for registration, tags, and agent `use_mcp`.
+
 ### Tool Runtime (Experimental)
 
 > **Warning:** This feature is experimental and may be removed or changed without warning in a future release.
@@ -46,11 +68,11 @@ Riffer.configure do |config|
 end
 ```
 
-| Value                          | Description                                                                                       |
-| ------------------------------ | ------------------------------------------------------------------------------------------------- |
-| `Riffer::ToolRuntime` subclass | Instantiated automatically (e.g., `Riffer::ToolRuntime::Inline`, `Riffer::ToolRuntime::Threaded`) |
-| `Riffer::ToolRuntime` instance | Custom runtime with specific options                                                              |
-| `Proc`                         | Dynamic resolution                                                                                |
+| Value                           | Description                                                                                        |
+| ------------------------------- | -------------------------------------------------------------------------------------------------- |
+| `Riffer::ToolRuntime` subclass  | Instantiated automatically (e.g., `Riffer::ToolRuntime::Inline`, `Riffer::ToolRuntime::Threaded`)  |
+| `Riffer::ToolRuntime` instance  | Custom runtime with specific options                                                               |
+| `Proc`                          | Dynamic resolution                                                                                 |
 
 Per-agent configuration overrides this global default. See [Advanced Tool Configuration — Tool Runtime](07_TOOL_ADVANCED.md#tool-runtime-experimental) for details.
 
--- a/docs/14_MCP.md
+++ b/docs/14_MCP.md
@@ -1,0 +1,167 @@
+# MCP
+
+Riffer can consume third-party [Model Context Protocol](https://modelcontextprotocol.io) (MCP) servers as tool sources. Tools are discovered automatically at registration time — no handwritten Ruby per tool required.
+
+## Overview
+
+1. Register an MCP server globally with `Riffer::Mcp.register`.
+2. Opt an agent into that server's tools using the `use_mcp` DSL.
+3. The agent picks up the tools and calls them like any other Riffer tool.
+
+## Registering a Server
+
+```ruby
+Riffer::Mcp.register(
+  name: "github",        # unique identifier
+  tags: [:github],       # agents opt-in by tag
+  endpoint: "https://mcp.github.com",
+  discovery_headers: -> { {"Authorization" => "Bearer #{ENV['GITHUB_TOKEN']}"} }
+)
+```
+
+`register` returns immediately. Tool discovery runs in a background thread. Use `on_pending` (see below) to control agent behaviour while discovery is in progress.
+
+### Discovery headers
+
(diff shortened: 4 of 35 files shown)


kepheusikarius:
@coderabbitai review

lib/riffer/mcp.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  # Raised when +Riffer.config.mcp.credentials+ returns +nil+ during +tools/call+
+  # after the server's tools were included for this run.
+  class CredentialsDeniedError < Error; end
+
+  # Registers an MCP server and starts async tool discovery.
+  #
+  # Returns immediately. Pass a +Manifest+ instance or a hash with the same keys.
+  #
+  #: ((Hash[Symbol, untyped] | Riffer::Mcp::Manifest)) -> Riffer::Mcp::Registration
```
bottrall, line 35: **issue(blocking):** missing `#--` RDoc stop directive before every `#:` annotation.

`.agents/rdoc.md`:

> Always add `#--` (RDoc stop directive) on the line before a standalone `#:` type annotation. Without it, RDoc treats `#:` as a label-list marker and corrupts the preceding comment into a `<pre>` block.

Same comment for every other lib file in this PR.

lib/riffer/mcp/client.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+# rbs_inline: enabled
+
+require "mcp"
```
bottrall: suggestion (non-blocking): load mcp lazily via depends_on to match the provider pattern.

Same deal with faraday.

.agents/architecture.md
```diff
@@ -130,6 +130,32 @@ Built-in runtimes:
 
 Context flow: `Agent#execute_tool_calls` → `ToolRuntime#execute(tool_calls, tools:, context:)` → `Runner#map(tool_calls, context:) { dispatch }` → `Tool#call(context:, **args)`
 
+### MCP Integration (`lib/riffer/mcp/`)
+
+Register third-party MCP servers globally; agents opt-in by tag via `use_mcp`. Tags are application-defined (manifests may list several; any overlap with `use_mcp` opts in—see `docs/10_MCP.md`).
```
bottrall: **issue(blocking):** broken link `docs/10_MCP.md` should be `docs/14_MCP.md`.


```suggestion
Register third-party MCP servers globally; agents opt-in by tag via `use_mcp`. Tags are application-defined (manifests may list several; any overlap with `use_mcp` opts in—see `docs/14_MCP.md`).
```

lib/riffer/mcp/client.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  # Calls a tool on the MCP server and returns joined text content from the response.
+  #
+  #: (String, ?Hash[untyped, untyped]) -> String
+  def tools_call(name, arguments = {})
+    tool = MCP::Client::Tool.new(name: name, description: nil, input_schema: nil)
+    response = @client.call_tool(tool: tool, arguments: arguments)
+
+    if response["error"]
```
bottrall, line 55: **nitpick(blocking):** string-keyed hash access deviates from the symbol-key convention.

`.agents/code-style.md`:

> Use **symbol keys** for all internal hashes … normalize to symbol keys at the boundary instead.

lib/riffer/mcp/authenticated_tool.rb
```diff
@@ -0,0 +1,45 @@
...
+          return inner.new.call(context: context, **kwargs)
+        end
+
+        headers = cred.call(manifest: man, matched_tags: tags, context: context)
+        if headers.nil?
+          raise Riffer::Mcp::CredentialsDeniedError,
+            "MCP credentials returned nil for server '#{man.name}' during tools/call"
+        end
+
+        client = Riffer::Mcp::Client.new(endpoint: man.endpoint, headers: headers)
```
bottrall: **thought (non-blocking):** building a fresh `Client` + Faraday connection per tool call will churn under load.

Each invocation builds a new `Riffer::Mcp::Client`, which builds a new `MCP::Client::HTTP` transport, which builds a new Faraday connection. Fine at low volume; under concurrent tool use (threaded runtime) or high throughput it'll show up in tail latency and connection churn.

Not necessarily blocking, but worth either a `# TODO: reuse client when headers are equal` or a tiny per-headers client cache.

.agents/architecture.md
```diff
@@ -130,6 +130,32 @@ Built-in runtimes:
...
+
+```ruby
+Riffer::Mcp.register(
+  name: "github",
+  tags: [:github],
+  endpoint: "https://mcp.github.com",
+  discovery_headers: -> { {"Authorization" => "Bearer #{ENV['GITHUB_TOKEN']}"} }
+)
+
+# Optional: per-run tools/call headers (see docs/10_MCP.md)
```
bottrall: **issue(blocking):** same broken link as line 135.


```suggestion
# Optional: per-run tools/call headers (see docs/14_MCP.md)
```

.agents/architecture.md
```diff
@@ -130,6 +130,32 @@ Built-in runtimes:
...
+Riffer.configure { |c| c.mcp.credentials = ->(manifest:, matched_tags:, context:) { ... } }
+
+class ResearchAgent < Riffer::Agent
+  model "openai/gpt-4o"
```
bottrall: **Nitpick(non-blocking):** Use more up to date models in examples.



We recently did a big cleanup here https://github.com/janeapp/riffer/pull/168.

Same comment for other examples in the docs.

docs/10_CONFIGURATION.md
```diff
@@ -34,6 +34,28 @@ Riffer.config.anthropic.api_key
...
 
+### MCP (Model Context Protocol)
+
+Optional settings for [MCP server integrations](14_MCP.md):
+
+| Option | Description |
+| ------ | ----------- |
+| `on_pending` | `:ignore` (default), `:wait`, or `:raise` when tool discovery is not finished yet |
+| `wait_timeout` | Seconds to wait when `on_pending` is `:wait` (default `10`) |
+| `credentials` | Optional `Proc` for per-run `tools/call` HTTP headers: `->(manifest:, matched_tags:, context:) { Hash or nil }` |
```
bottrall: **Nitpick(non-blocking):** Format table with prettier.



When I or someone with prettier installed in their editor inevitably updates this file it will cause a bunch of whitespace churn 😅 Worth doing the same on the other md files touched.

lib/riffer/mcp/agent_factory.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+# rbs_inline: enabled
+
+# Generates an anonymous Riffer::Agent subclass for an MCP server.
+#
+# The generated agent has no model, instructions, or system prompt — it is
+# a container for the server's tools, exposed via the registration's +agent+
+# attribute.
+#
+module Riffer::Mcp::AgentFactory
```
bottrall: **Question(blocking):** I'm not fully understanding what this factory and the tool factory are for?



From what I can tell it's just so we can reuse the methods defined on the agent in the discovery thread? If so I think that's a good indication that the MCP implementation is too coupled to agents and is probably worth separating out.

lib/riffer/mcp/registration.rb
```diff
@@ -0,0 +1,104 @@
...
+        @ready = true
+      end
+    rescue => e
+      @mutex.synchronize do
+        next if @cancelled
+        @discovery_error = e
+      end
+      # Leave @ready = false — callers apply the on_pending strategy
+    end
+  end
```
bottrall: **Suggestion(blocking):** This goes against the pattern we've been developing of allowing configurable concurrency solutions.



See [here](https://docs.riffer.ai/07_TOOL_ADVANCED_md.html#tool-runtime-experimental) for our tool runtime example. It would be nice to follow this pattern as spawning threads in some situations could cause unwanted side effects.

Gemfile.lock
```diff
@@ -269,5 +284,116 @@ DEPENDENCIES
...
+  rake (13.3.1) sha256=8c9e89d09f66a26a01264e7e3480ec0607f0c497a861ef16063604b1b08eb19c
+  rb-fsevent (0.11.2) sha256=43900b972e7301d6570f64b850a5aa67833ee7d87b458ee92805d56b7318aefe
+  rb-inotify (0.11.1) sha256=a0a700441239b0ff18eb65e3866236cd78613d6b9f78fea1f9ac47a85e47be6e
+  rbs (3.10.4) sha256=b17d7c4be4bb31a11a3b529830f0aa206a807ca42f2e7921a3027dfc6b7e5ce8
+  rbs-inline (0.13.0) sha256=aba6e48c2d1b75276e8557376164f4b4ba96dd0204e779a5cee2af442bd30442
+  rdoc (7.2.0) sha256=8650f76cd4009c3b54955eb5d7e3a075c60a57276766ebf36f9085e8c9f23192
+  regexp_parser (2.11.3) sha256=ca13f381a173b7a93450e53459075c9b76a10433caadcb2f1180f2c741fc55a4
+  reline (0.6.3) sha256=1198b04973565b36ec0f11542ab3f5cfeeec34823f4e54cebde90968092b1835
+  rexml (3.4.4) sha256=19e0a2c3425dfbf2d4fc1189747bdb2f849b6c5e74180401b15734bc97b5d142
+  riffer (0.23.0)
```
bottrall: **Issue(blocking):** We can't introduce checksums as it breaks release-please.



See https://github.com/googleapis/release-please/issues/2720

riffer.gemspec
```diff
@@ -38,6 +38,8 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   # Runtime dependencies
+  spec.add_dependency "faraday", ">= 1.0"
+  spec.add_dependency "mcp", "~> 0.8"
```
bottrall: **Issue(blocking):** Add to dev dependencies.



A stance we've taken in this project is that all dependencies are optional with the exception of zeitwork. That way consumers only install what they need for each feature.

docs/14_MCP.md
```diff
@@ -0,0 +1,160 @@
+# MCP
+
+Riffer can consume third-party [Model Context Protocol](https://modelcontextprotocol.io) (MCP) servers as tool sources. Tools are discovered automatically at registration time — no handwritten Ruby per tool required.
```
bottrall, line 3: **Thought(non-blocking):** We may want to put more thought into the auto discovery.



This is a fast tracked way of bloating the context window and many other products/libraries have had to do work so that mcp tools can be progressively disclosed to agents.

bottrall (changes requested):
Thanks @kepheusikarius.

.agents/architecture.md
kepheusikarius: Doh, thought I got all these. Thanks!

docs/10_CONFIGURATION.md
kepheusikarius: Can do. I much prefer this style exactly because it doesn't care about whitespace, but agree that consistency is preferable.

lib/riffer/mcp/agent_factory.rb
kepheusikarius: I think this is a leftover from a previous iteration as it's not used. Will remove.

lib/riffer/mcp/client.rb
kepheusikarius: I thought about this one some more and I feel like string keys makes sense here precisely because the tool call is on the boundary. Marshalling to symbols just to parse the API response seems unnecessary.

`.agents/code-style.md` (emphasis mine):
> String keys are only used at **serialization boundaries** (JSON Schema output, **external API payloads**)

That said, if y'all feel strongly about it, I'll update it.

lib/riffer/mcp/authenticated_tool.rb
kepheusikarius: Yep, fair point.

I'll leave this as a TODO for now. The cache itself is relatively straight-forward, but having a bad client in cache could lead to a lot of other issues, so I think it's worth diving into cache invalidation separately.

docs/14_MCP.md
```diff
@@ -0,0 +1,167 @@
...
+
+**Call time:** Authenticated tool wrappers invoke the proc again for each execution. If it returns **`nil`**, `Riffer::Mcp::CredentialsDeniedError` is raised.
+
+If **`credentials` is unset**, discovery and `tools/call` share one client built from `discovery_headers` (same behaviour as a single static token for both list and call).
+
+## Tags
+
+Tags are entirely up to your application; Riffer does not define a canonical vocabulary. Each server may declare multiple tags; an agent includes every registration that shares **any** tag passed to `use_mcp`.
+
+When MCPs are registered, assign **stable bucket tags** so agent classes can opt in without listing every server name—for example `tags: [:connectors, :github]` with `use_mcp :connectors` for all enabled connectors, or `use_mcp :github` for that integration only.
```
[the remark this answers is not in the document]
kepheusikarius: It's a code snippet in context, so the casing is correct.

docs/14_MCP.md
kepheusikarius: Added a note to `docs/14_MCP.md` to follow up. Definitely open to further discussions/suggestions on how to approach this. I suspect this'll also be true as teams add new tools manually (though MCP will hasten the issue), so a more holistic approach may be warranted.

kepheusikarius:
Thanks again for the feedback! Ready for another round (pending RBS fixes).

docs/14_MCP.md
bottrall: Context bloat via your own tools versus context bloat from MCP tools are two different issues IMO & we have ideas on how to solve the former downstream. MCP is a bit more foundational though and the current approach doesn't really allow consumers to solve it themselves.

We should follow the best practices laid out in the [MCP client docs](https://modelcontextprotocol.io/docs/develop/clients/client-best-practices#progressive-tool-discovery).

kepheusikarius: Fair point. So what's the path forward here? Are we reclassifying this from **thought (non-blocking)** to **suggestion (blocking)** and addressing it here or taking another approach?

lib/riffer/mcp/registration.rb
bottrall: Hey @kepheusikarius, bumping this as it's the only blocking feedback I left that hasn't been addressed.

kepheusikarius: Sorry @bottrall, I thought I addressed this in [`Riffer::Mcp::Config`](https://github.com/janeapp/riffer/pull/180/changes#diff-4e25b831d47feb2929ed3a2ce66e74c9d065ddcd5d2eda7aa048722e0b64182dR112) with the `discovery_thread_factory`. If that's not what you were looking for, let's chat more.

docs/14_MCP.md
bottrall: Happy to leave it as is for now if you think it adds too much extra scope to this PR. But I would like us to address this in the near to short term if possible.

lib/riffer/mcp/registration.rb
bottrall: If possible we should try to reuse the runner primitive. So a consumer can declare what runner is used. That and the default should be inline, as we've done for everything else.

kepheusikarius: K, will dig in. Thanks for clarifying!

docs/14_MCP.md
kepheusikarius: Let's make this a follow-up. This PR is already longer than I'd like, and should be additive work.

lib/riffer/mcp/registration.rb
kepheusikarius: I replaced the async discovery handling with `Riffer::Runner::Sequential`, so things are a bit simpler, but it's still configurable in case someone wants to override it. Please see https://github.com/janeapp/riffer/pull/180/commits/114ad2c46ec950a4e01abe6ca311e665139a5e7b for details.

bottrall (approved):

PR #181, opened 2026-03-25
feat: provide agent context to runners

bottrall:
## Summary

- Add required `context:` keyword argument to `Runner#map` so runner implementations can access the agent context for job serialization
- `ToolRuntime#execute` now forwards context through to the runner
- Built-in `Sequential` and `Threaded` runners accept context (no behavior change)
- Architecture docs updated with Runner and ToolRuntime sections

## Test plan

- [x] All existing tests pass (`bundle exec rake` — 1164 tests, 0 failures)
- [x] Tests verify all runners require `context:` keyword
- [x] Test verifies custom runner subclass can access context in `map`
- [x] RBS type signatures regenerated and type checking passes (`steep check`)

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -107,6 +107,29 @@ MyAgent.new.stream(persisted_messages)  # cross-process resume
 
 On resume, `execute_pending_tool_calls` detects tool calls from the last assistant message that lack corresponding tool result messages and executes them before entering the LLM loop. This handles the case where an interrupt fired mid-way through tool execution.
 
+### Runner (`lib/riffer/runner.rb`)
+
+Concurrency primitive for batch execution. Subclasses implement `#map(items, context: nil, &block)` to control how items are processed. The `context` keyword carries the agent's context hash, enabling runners that need it for job serialization or routing.
+
+Built-in runners:
+- `Sequential` — processes items in the current thread via `Array#map`
+- `Threaded` — processes items concurrently using a thread pool with configurable `max_concurrency`
+
+```ruby
+runner = Riffer::Runner::Threaded.new(max_concurrency: 3)
+runner.map(items, context: ctx) { |item| process(item) }
+```
+
+### ToolRuntime (`lib/riffer/tool_runtime.rb`)
+
+Composes with a Runner to execute tool calls. Provides `#execute` as the public entry point and `#around_tool_call` as a hook for instrumentation. Passes the agent context through to the runner.
+
+Built-in runtimes:
+- `Inline` — uses `Runner::Sequential` (default)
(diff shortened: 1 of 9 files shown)


lib/riffer/runner/sequential.rb
```diff
@@ -6,8 +6,8 @@
 # This is the default runner used when no concurrency is needed.
 #
 class Riffer::Runner::Sequential < Riffer::Runner
-  #: (Array[untyped]) { (untyped) -> untyped } -> Array[untyped]
-  def map(items, &block)
+  #: (Array[untyped], ?context: Hash[Symbol, untyped]?) { (untyped) -> untyped } -> Array[untyped]
+  def map(items, context: nil, &block)
```
Thibaut: question (non-blocking): why make it optional? I think this could result in more complexity (nil checks) or bugs. As it stands, we always pass `context`, so I would leave it as required until such time that we have a good reason not to (unless there already exists a good reason)

.agents/architecture.md
```diff
@@ -107,6 +107,29 @@ MyAgent.new.stream(persisted_messages)  # cross-process resume
 
 On resume, `execute_pending_tool_calls` detects tool calls from the last assistant message that lack corresponding tool result messages and executes them before entering the LLM loop. This handles the case where an interrupt fired mid-way through tool execution.
 
+### Runner (`lib/riffer/runner.rb`)
+
+Concurrency primitive for batch execution. Subclasses implement `#map(items, context: nil, &block)` to control how items are processed. The `context` keyword carries the agent's context hash, enabling runners that need it for job serialization or routing.
+
+Built-in runners:
```
Thibaut, line 114: question (non-blocking): is there a case to be made that runners are a premature and potentially unnecessary abstraction? Would it not simplify things for now to build the runner directly (inline) in the runtime? It doesn't seem that we'd want to mix & match runtimes and runners. If so (and runners/runtimes will always be 1:1 in the short term), then I think it may be worth consolidating/simplifying.

Thibaut (approved):
Some thoughts to ponder upon. I trust in whatever you decide 😄

.agents/architecture.md
bottrall: It's not premature in the sense that there is already a [PR](https://github.com/janeapp/riffer/pull/169) up that makes use of this abstraction for agent orchestration. Which as we've all discussed (and I think we agree on?) running tools, agent orchestration and workflow orchestration benefit from being able to define different runtime characteristics.

Thibaut: Gotcha :+1: Hadn't made the connection in that PR.

PR #185, opened 2026-03-26
refactor: simplify dependency helper

bottrall:
## Summary

- Simplify `Riffer::Helpers::Dependencies#depends_on` to just `require` + friendly error message
- Remove `VersionError` class, `req:` parameter, `gem()` call, and all Bundler version-checking logic
- Use `depends_on` for `async/semaphore` in `Riffer::Runner::Fibers` for consistency
- Update tests and regenerate RBS types

## Test plan

- [x] `bundle exec rake` passes (1165 tests, 0 failures, lint clean, types clean)

## Release Notes

* **Refactor**
  * Streamlined dependency declaration by removing bundler and version validation checks.
  * Simplified method signature for declaring optional dependencies.

* **Tests**
  * Updated tests to reflect new dependency handling behavior.

### DIFF
--- a/lib/riffer/helpers/dependencies.rb
+++ b/lib/riffer/helpers/dependencies.rb
@@ -8,41 +8,14 @@ module Riffer::Helpers::Dependencies
   # Raised when a required gem cannot be loaded.
   class LoadError < ::LoadError; end
 
-  # Raised when a gem version requirement is not satisfied.
-  class VersionError < ScriptError; end
-
-  # Declares a dependency on a gem.
+  # Requires a gem by name, raising a helpful error if it is not installed.
   #
-  # Verifies the gem is installed and satisfies version requirements,
-  # then requires it.
-  #
-  # Raises LoadError if the gem is not installed.
-  # Raises VersionError if the gem version does not satisfy requirements.
+  # Raises LoadError if the gem cannot be required.
   #
   #--
-  #: (String, ?req: (bool | String)) -> true
-  def depends_on(gem_name, req: true)
-    gem(gem_name)
-
-    return true unless defined?(Bundler)
-
-    gem_version = Gem.loaded_specs[gem_name].version
-    gem_dependency = Bundler.load.dependencies.find { |g| g.name == gem_name }
-    gem_requirement = gem_dependency&.requirement
-
-    unless gem_requirement
-      raise VersionError, "The #{gem_name} gem is installed but not specified in your Bundler dependencies (e.g., Gemfile)."
-    end
-
-    unless gem_requirement.satisfied_by?(gem_version)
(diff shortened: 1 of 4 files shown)


lib/riffer/helpers/dependencies.rb
```diff
@@ -8,41 +8,14 @@ module Riffer::Helpers::Dependencies
...
-    end
-
-    lib_name = gem_name if req == true
-    lib_name = req if req.is_a?(String)
-
-    require(lib_name) if lib_name
-
+  #: (String) -> true
+  def depends_on(gem_name)
+    require gem_name
```
Thibaut, line 18: 👌

lib/riffer/helpers/dependencies.rb
```diff
@@ -8,41 +8,14 @@ module Riffer::Helpers::Dependencies
...
-      raise VersionError, "The #{gem_name} gem is installed, but version #{gem_requirement} is required. You have #{gem_version}."
-    end
-
-    lib_name = gem_name if req == true
-    lib_name = req if req.is_a?(String)
-
-    require(lib_name) if lib_name
-
+  #: (String) -> true
+  def depends_on(gem_name)
```
Thibaut, line 17: suggestion (non-blocking): should we simply use `require` then instead of `depends_on`? The only difference is the error message, but the default one is probably fine no?

Thibaut (approved):

lib/riffer/helpers/dependencies.rb
bottrall: Good question. Currently a developer would get the following:
```
Riffer::Helpers::Dependencies::LoadError: Could not load nonexistent_gem_xyz_12345. Please ensure that the nonexistent_gem_xyz_12345 gem is installed.
    lib/riffer/helpers/dependencies.rb:21:in 'Riffer::Helpers::Dependencies#depends_on'
```

Without this helper they would get:
```
LoadError: cannot load such file -- nonexistent_gem_xyz_12345
    /Users/jakebottrall/.rbenv/versions/3.4.8/lib/ruby/3.4.0/bundled_gems.rb:82:in 'Kernel.require'
    /Users/jakebottrall/.rbenv/versions/3.4.8/lib/ruby/3.4.0/bundled_gems.rb:82:in 'block (2 levels) in Kernel#replace_require'
    /Users/jakebottrall/.rbenv/versions/3.4.8/lib/ruby/gems/3.4.0/gems/zeitwerk-2.7.5/lib/zeitwerk/core_ext/kernel.rb:34:in 'Kernel#require'
    lib/riffer/helpers/dependencies.rb:18:in 'Riffer::Helpers::Dependencies#depends_on'
```

I think our error message is a bit more friendly and helpful. It also standardizes the best practice of including gems in the project. For those reasons and considering it's little to no burden to maintain, I think it's worth keeping.

PR #189, opened 2026-03-31
feat: add timestamp to messages

cloetzi:
## Summary

- Adds a `timestamp` attribute to `Messages::Base` (defaults to `Time.now`, overridable)
- Serialized as ISO 8601 with millisecond precision in `to_h`
- `Converter` round-trips timestamps from hashes (parses ISO 8601 strings and passes through `Time` objects)
- All subclasses (System, User, Assistant, Tool) forward `timestamp:` to base

## Test plan

- [x] All 1183 existing tests pass (0 failures, 0 errors)
- [x] New tests for default timestamp, custom timestamp, `to_h` serialization, and converter round-trip on each message type
- [x] StandardRB lint clean (no new offenses)

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -65,6 +65,8 @@ Typed message objects that extend `Riffer::Messages::Base`:
 - `Assistant` - AI responses
 - `Tool` - tool execution results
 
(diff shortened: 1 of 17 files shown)


cloetzi:
We had the discussion, this is not a riffer concern.

PR #190, opened 2026-04-01
feat(amazon_bedrock): support S3 URI file sources

bottrall:
## Summary
- Add support for S3 URI (`s3://`) file sources in the Amazon Bedrock provider, in addition to existing base64 data sources
- S3 URIs are passed as `s3_location` in the Bedrock API request, enabling direct references to files in S3 buckets
- Document filenames are derived from the S3 URI path when not explicitly provided
- Non-S3 URLs now raise `Riffer::ArgumentError` with a clear message

## Test plan
- Added VCR cassette tests for S3 URI images and documents via both `generate_text` and `stream_text`
- Added test for unsupported URL source raising `ArgumentError`

## Summary by CodeRabbit

* **New Features**
  * Support for supplying files via S3 URIs to the Amazon Bedrock integration for images and documents; payloads now accept either embedded file bytes or S3 references.

* **Tests**
  * Added end-to-end and streaming tests for S3 file inputs, a negative test for unsupported/non-S3 URLs, recorded fixtures for S3 interactions, and expanded test redaction for S3 URI placeholders.


lib/riffer/providers/amazon_bedrock.rb
```diff
@@ -287,15 +287,21 @@ def append_tool_result(conversation_messages, message)
...
+    else
+      raise Riffer::ArgumentError, "Amazon Bedrock only supports S3 URI or base64 data file sources"
+    end
 
     if file.image?
-      {image: {format: format, source: {bytes: bytes}}}
+      {image: {format: format, source: source}}
     else
-      {document: {format: format, name: file.filename, source: {bytes: bytes}}}
+      name = file.filename || ::File.basename(URI.parse(file.url).path, ".*")
```
cloetzi: suggestion(blocking): Edge-case bug found by CC:
>If a user provides a base64 document without a filename (and no URL), file.url is nil, and URI.parse(nil) will raise a TypeError. The old code just passed file.filename (which could also be nil, but wouldn't crash). This is a regression for base64 documents without filenames.

I don't think we have any nil protection anywhere, so it might be worth addressing.

lib/riffer/providers/amazon_bedrock.rb
```diff
@@ -287,15 +287,21 @@ def append_tool_result(conversation_messages, message)
   #--
   #: (Riffer::FilePart) -> Hash[Symbol, untyped]
   def convert_file_part_to_bedrock_format(file)
-    raise Riffer::ArgumentError, "Amazon Bedrock does not support URL file sources; provide base64 data instead" if file.url? && file.data.nil?
-
     format = bedrock_format(file.media_type)
-    bytes = Base64.decode64(file.data)
+
+    source = if file.data
+      {bytes: Base64.decode64(file.data)}
```
cloetzi, line 295: question(non-blocking): I comment from CC made check the `require` for `base64` and `uri`. It looks like we are quite inconsistent in regard to requiring gems. What is the loading pattern we are using in riffer? Is it autoloading gems? In either case, we should either use `require` consistent, or not use it at all if we don't have to and clean up the existing ones.

cloetzi (changes requested):
question(non-blocking): Have you decided not to document the s3 URI feature and its limitations?

lib/riffer/providers/amazon_bedrock.rb
bottrall: Sorry Claude slipped that in when it was fixing tests without me noticing. Addressed here https://github.com/janeapp/riffer/pull/190/commits/d2c26cbd86f5d022eebd3921948e4ebb27498452.

We previously made the decision to not guard against filename being nil, as the bedrock client raises the appropriate error message & we don't necessarily want to get into the business of accounting for every error that can occur within the providers.

The s3 URi is a special case here in my mind because the riffer api makes it seem like any url will work & for any other provider it will.

lib/riffer/providers/amazon_bedrock.rb
bottrall: https://github.com/janeapp/riffer/pull/190/commits/704d2662a0da2b383960543ae37bf33b8b043505

I'd prefer to be explicit for now, as we're fairly consistent with that.

lib/riffer/providers/amazon_bedrock.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 # rbs_inline: enabled
 
+require "base64"
```
cloetzi, line 4: suggestion(non-blocking): Should also require `uri`.

cloetzi (approved):

lib/riffer/providers/amazon_bedrock.rb
bottrall: This file doesn't explicitly use `uri` though? It uses `FilePart` which explicitly uses it and imports it.

PR #193, opened 2026-04-02
refactor: remove redundant `id` from ToolCall struct

bottrall:
## Summary
- Remove the `id` field from `Messages::Assistant::ToolCall`, keeping only `call_id` for tool result matching across all providers
- The `id` was originally added because OpenAI distinguishes item IDs (`fc_*`) from call IDs (`call_*`), but in practice only `call_id` is needed — Anthropic and Bedrock already duplicated the same value into both fields
- Update all providers (OpenAI, Anthropic, Amazon Bedrock, Mock), the agent tool execution loop, format converters, RBS signatures, tests, and VCR cassettes

### DIFF
--- a/lib/riffer/agent.rb
+++ b/lib/riffer/agent.rb
@@ -502,7 +502,6 @@ class Riffer::Agent
             current_tool_call[:name] ||= event.name
           when Riffer::StreamEvents::ToolCallDone
             accumulated_tool_calls << Riffer::Messages::Assistant::ToolCall.new(
-              id: event.item_id,
               call_id: event.call_id,
               name: event.name,
               arguments: event.arguments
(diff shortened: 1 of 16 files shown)


lib/riffer/agent.rb
```diff
@@ -590,7 +589,7 @@ def execute_tool_calls(response)
     results.each do |tool_call, result|
       add_message(Riffer::Messages::Tool.new(
         result.content,
-        tool_call_id: tool_call.id,
+        tool_call_id: tool_call.call_id,
```
bottrall, line 592: From what I can tell this was actually a dormant bug. However, it would only effect the openai provider because in our other providers `id` and `call_id` are the same.

cloetzi (approved):

PR #199, opened 2026-04-07
feat: add Gemini provider

bottrall:
## Summary

- Add Google Gemini as a new LLM provider adapter (`Riffer::Providers::Gemini`)
- Implements all hook methods: `generate_text`, `stream_text`, structured output, tool calling, file handling (base64 images), and token usage tracking
- Uses the Gemini REST API directly via `Net::HTTP` (stdlib) — no external gem dependency required
- True incremental streaming via `Net::HTTP#read_body` chunked reading with SSE frame parsing
- Strips `additionalProperties` from JSON schemas as Gemini doesn't support them
- Generates client-side `call_id` values for tool calls since Gemini doesn't provide them
- Safely handles nil tool call arguments via `encode_tool_arguments` helper
- Passes API key via `x-goog-api-key` header (not URL query parameter) to prevent credential exposure in logs
- Validates model names against `VALID_MODEL_PATTERN` before URL interpolation
- Configurable `open_timeout` and `read_timeout` with sensible defaults

## Files

- `lib/riffer/providers/gemini.rb` — Provider implementation
- `lib/riffer/config.rb` — Gemini config struct
- `lib/riffer/providers/repository.rb` — Registry entry
- `test/riffer/providers/gemini_test.rb` — 39 tests with VCR cassettes
- `docs/providers/08_GEMINI.md` — Provider documentation

## Summary by CodeRabbit

* **New Features**
  * Added Google Gemini provider: text generation, streaming, structured JSON output, tool calling, and inline file/image support.
  * Global and per-agent Gemini API key and timeout configuration exposed.

* **Documentation**
  * Comprehensive Gemini provider guide with setup, supported models, options, examples, limitations.

* **Tests**
  * New provider test suite and numerous recorded fixtures for generation, streaming, tool-calling, file handling, and usage. 

* **Chores**
  * VCR sensitive-data filter for Gemini API key.


lib/riffer/providers/gemini.rb
```diff
@@ -0,0 +1,307 @@
...
+    request = Net::HTTP::Post.new(uri)
+    request["Content-Type"] = "application/json"
+    request.body = body.to_json
+    Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) { |http| http.request(request) }
+  end
+
+  #--
+  #: (String, String) -> String
+  def api_path(model, method)
+    "v1beta/models/#{model}:#{method}?key=#{@api_key}"
```
cloetzi: suggestion(non-blocking): Does their API support a header for the API key?

lib/riffer/providers/gemini.rb
```diff
@@ -0,0 +1,307 @@
...
+  def execute_stream(params, yielder)
+    model = params[:model]
+    body = params.except(:model)
+
+    uri = URI("#{BASE_URI}/#{api_path(model, "streamGenerateContent")}&alt=sse")
+    request = Net::HTTP::Post.new(uri)
+    request["Content-Type"] = "application/json"
+    request.body = body.to_json
+
+    full_text = ""
```
cloetzi: suggestion(blocking): I think the way this is done is quite inefficient because each time you call `+=`, you create a full new string. You can apply the same pattern as `buffer` uses:
```ruby
  full_text = +""
  # ...
  full_text << part[:text]
```

lib/riffer/providers/gemini.rb
```diff
@@ -0,0 +1,307 @@
...
+        parsed = JSON.parse(json_str, symbolize_names: true)
+        parts = parsed.dig(:candidates, 0, :content, :parts)
+
+        parts&.each do |part|
+          if part[:text]
+            full_text += part[:text]
+            yielder << Riffer::StreamEvents::TextDelta.new(part[:text])
+          elsif part[:functionCall]
+            fc = part[:functionCall]
+            call_id = "gemini_call_#{SecureRandom.hex(12)}"
```
cloetzi, line 144: question(non-blocking): Is gemini not having its own tool id?

lib/riffer/providers/gemini.rb
```diff
@@ -0,0 +1,307 @@
...
+            arguments = encode_tool_arguments(fc[:args])
+            yielder << Riffer::StreamEvents::ToolCallDone.new(
```
cloetzi, line 146: thought(non-blocking): The `ToolCallDone` name is a bit confusing, I am wondering if an alias like `ToolCallRequestDone` would be beneficial?

lib/riffer/providers/gemini.rb
cloetzi: suggestion(non-blocking): Can we validate the model string somehow with a regex, e.g. ensuring there's no slash or any other weird characters that would allow path escape somehow? While the risk is kinda negligible, cycode is right and a small extra guard might mitigate future embarrassment 😄

lib/riffer/providers/gemini.rb
```diff
@@ -0,0 +1,307 @@
...
+            token_usage: Riffer::TokenUsage.new(
+              input_tokens: usage[:promptTokenCount] || 0,
+              output_tokens: usage[:candidatesTokenCount] || 0
+            )
+          )
+        end
+      end
+    end
+
+    Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) do |http|
```
cloetzi: suggestion(non-blocking): Claude flagged that there are no timeouts, I am wondering if we can configure some reasonable defaults some in the gemini provider config (e.g. 60s for read, 10s for open)

cloetzi (changes requested):
Just one blocking comment regarding the string handling, otherwise just suggestions/questions.

lib/riffer/providers/gemini.rb
bottrall: No. Gemini's API returns `functionCall` objects with only name and args. No call ID. Their API is looser than previous providers.

lib/riffer/providers/gemini.rb
bottrall: Agreed, however that's a bit out of scope for this PR.

cloetzi (approved):

PR #201, opened 2026-04-09
feat: normalize consecutive same-role messages before provider serialization

bottrall:
## Summary

- Adds `merge_consecutive_messages` to `Riffer::Providers::Base` that merges consecutive same-role messages by concatenating content blocks and combining auxiliary data (files, tool_calls) into a single message
- Runs after validation but before `build_request_params` in both `generate_text` and `stream_text`, ensuring all four providers receive identical normalized message lists
- Tool messages are never merged since each has a unique `tool_call_id`
- Prevents behavioral divergence when the same model is accessed through different providers (e.g. Anthropic direct vs Bedrock) — context tools inject consecutive user messages that some providers reject or auto-merge differently

## Test plan

- 8 new tests covering: alternating passthrough, single message passthrough, consecutive user merge, file combining, system merge, assistant merge with tool_calls, tool non-merge, mixed sequences, and the context message + user message scenario

## Summary by CodeRabbit

* **New Features**
  * Consecutive messages from the same role are merged before sending to providers: System/User/Assistant contents are concatenated with a blank line; User file lists and Assistant tool calls are combined; Tool messages are never merged. In-memory message list remains unchanged for logging/debugging.

* **Documentation**
  * Added a “Consecutive Message Merging” doc with rules and a concrete example.

* **Tests**
  * Added tests covering merge semantics and edge cases.

### DIFF
--- a/lib/riffer/providers/base.rb
+++ b/lib/riffer/providers/base.rb
@@ -126,25 +126,10 @@ class Riffer::Providers::Base
   #--
   #: (Array[Riffer::Messages::Base]) -> Array[Riffer::Messages::Base]
   def merge_consecutive_messages(messages)
-    messages.chunk { |msg| msg.class }.flat_map do |klass, group|
-      next group if klass == Riffer::Messages::Tool || group.size == 1
+    messages.chunk { |msg| msg.role }.flat_map do |role, group|
+      next group if role == :tool || group.size == 1
 
-      case klass.name
-      when "Riffer::Messages::System"
-        Riffer::Messages::System.new(group.map(&:content).join("\n\n"))
-      when "Riffer::Messages::User"
-        Riffer::Messages::User.new(
-          group.map(&:content).join("\n\n"),
-          files: group.flat_map(&:files)
-        )
-      when "Riffer::Messages::Assistant"
-        Riffer::Messages::Assistant.new(
-          group.map(&:content).join("\n\n"),
-          tool_calls: group.flat_map(&:tool_calls)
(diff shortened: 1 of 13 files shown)


lib/riffer/providers/base.rb
```diff
@@ -121,6 +123,31 @@ def parse_tool_arguments(arguments)
     arguments.is_a?(String) ? JSON.parse(arguments) : arguments
   end
 
+  #--
+  #: (Array[Riffer::Messages::Base]) -> Array[Riffer::Messages::Base]
+  def merge_consecutive_messages(messages)
+    messages.chunk { |msg| msg.class }.flat_map do |klass, group|
+      next group if klass == Riffer::Messages::Tool || group.size == 1
+
+      case klass.name
```
Thibaut: suggestion (non-blocking): the robot below has a bit of a point. Using `msg.role` instead of `msg.class` would let us abstract the class, such that for example our `role: "context` messages (IIRC) in Jane would automatically get merged together by this. 

To encapsulate the class-specific logic for merging messages (the join string, files/tool calls…) we could override `#+` on `Riffer::Message::System` &co, then the code here would all simplify down to something like `group.inject(group.first, &:+)` (or maybe even `group.sum` but that's too "clever" 😝)

Thibaut (approved):

lib/riffer/providers/base.rb
bottrall: Good call 👍🏻

PR #210, opened 2026-04-17
feat: add optional message id generation

cloetzi:
## Summary

Adds a global `Riffer.config.message_id_strategy` setting with three supported values:

- `:none` (default) — no ids, existing behavior preserved
- `:uuid` — UUIDv4 via `SecureRandom.uuid`
- `:uuidv7` — time-ordered UUIDv7 via `SecureRandom.uuid_v7` (Ruby 3.3+ built-in)

When the strategy is not `:none`, each `Riffer::Messages::Base` instance gets an `id` auto-populated at construction time. The `id` is exposed via `#id`, included in `#to_h` when present (and omitted when nil), and preserved when passed explicitly.

Seeded messages passed to `Riffer::Agent#generate` must carry their own `:id` when the strategy is enabled — Riffer never fabricates identifiers for pre-existing history. Missing seed ids raise `Riffer::ArgumentError` with the offending index.

## Changes

- `lib/riffer/config.rb` — new `message_id_strategy` attribute + validated setter
- `lib/riffer/messages/base.rb` — `id` attr, `id:` kwarg, auto-generation, conditional serialization
- `lib/riffer/messages/{user,assistant,tool}.rb` — forward `id:` via `super`, include in `to_h`
- `lib/riffer/messages/converter.rb` — propagate `hash[:id]` into every role branch
- `lib/riffer/agent.rb` — `validate_seed_ids!` called from `initialize_messages`
- Tests for config, base/subclass behavior, converter propagation, and agent seed validation
- Regenerated RBS signatures

Provider serialization is unaffected: providers manually build API-request hashes from message attributes and do not spread `message.to_h`, so the new `id` field does not leak to external APIs.

## Test plan

- [x] `bundle exec rake test` — 1265 tests, 1958 assertions, 0 failures
- [x] `bundle exec rake standard` — clean
- [x] `bundle exec steep check` — no type errors
- [x] `bundle exec rake rbs:generate` — signatures regenerated
- [ ] Verify consumer (downstream app) can opt in with `Riffer.configure { |c| c.message_id_strategy = :uuidv7 }` and round-trip seeded history with ids
- [ ] Confirm mock provider cassettes (if any) still match after the `:id` key addition

## Summary by CodeRabbit

* **New Features**
  * Configurable message ID strategy with optional auto-generation (:none [default], :uuid, :uuidv7).
  * Messages now support optional per-message IDs and enforce seeded-message ID requirements when strategy is enabled.

* **Documentation**
  * Added configuration and per-message ID docs and clarified docs guidance for public-facing options.
  * Minor formatting improvements to several docs.

* **Tests**
  * Added tests covering ID generation, preservation, seeding validation, and converter behavior.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -112,6 +112,7 @@ On resume, `execute_pending_tool_calls` detects tool calls from the last assista
 Concurrency primitive for batch execution. Subclasses implement `#map(items, context: nil, &block)` to control how items are processed. The `context` keyword carries the agent's context hash, enabling runners that need it for job serialization or routing.
 
 Built-in runners:
+
 - `Sequential` — processes items in the current thread via `Array#map`
 - `Threaded` — processes items concurrently using a thread pool with configurable `max_concurrency`
 
@@ -125,6 +126,7 @@ runner.map(items, context: ctx) { |item| process(item) }
 Composes with a Runner to execute tool calls. Provides `#execute` as the public entry point and `#around_tool_call` as a hook for instrumentation. Passes the agent context through to the runner.
 
 Built-in runtimes:
+
 - `Inline` — uses `Runner::Sequential` (default)
 - `Threaded` — uses `Runner::Threaded`
 
--- a/docs/08_MESSAGES.md
+++ b/docs/08_MESSAGES.md
@@ -252,12 +252,12 @@ Without this step, the same model can receive different input depending on the p
 
 ### Merge rules
 
-| Message type | Content | Auxiliary data | Merged? |
-|--------------|---------|----------------|---------|
-| `System` | Joined with `"\n\n"` | — | Yes |
-| `User` | Joined with `"\n\n"` | `files` arrays concatenated | Yes |
-| `Assistant` | Joined with `"\n\n"` | `tool_calls` arrays concatenated | Yes |
-| `Tool` | — | — | Never (each has a unique `tool_call_id`) |
+| Message type | Content              | Auxiliary data                   | Merged?                                  |
+| ------------ | -------------------- | -------------------------------- | ---------------------------------------- |
+| `System`     | Joined with `"\n\n"` | —                                | Yes                                      |
+| `User`       | Joined with `"\n\n"` | `files` arrays concatenated      | Yes                                      |
+| `Assistant`  | Joined with `"\n\n"` | `tool_calls` arrays concatenated | Yes                                      |
+| `Tool`       | —                    | —                                | Never (each has a unique `tool_call_id`) |
 
 ### Example
 
--- a/docs/08_MESSAGES.md
+++ b/docs/08_MESSAGES.md
@@ -278,20 +278,50 @@ agent.generate(messages)
 
 Merging happens at serialization time only. The agent's `messages` array still contains the original separate messages for logging, evals, and debugging.
 
+## IDs
+
+Every message carries an optional `id` attribute. By default ids are disabled (`message.id` returns `nil` and `:id` is omitted from `to_h`). Enable them globally by setting `Riffer.config.message_id_strategy`:
+
+```ruby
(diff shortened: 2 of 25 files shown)


.agents/architecture.md
```diff
@@ -112,6 +112,7 @@ On resume, `execute_pending_tool_calls` detects tool calls from the last assista
 Concurrency primitive for batch execution. Subclasses implement `#map(items, context: nil, &block)` to control how items are processed. The `context` keyword carries the agent's context hash, enabling runners that need it for job serialization or routing.
 
 Built-in runners:
+
```
cloetzi, line 115: I ran prettier against all mds.

lib/riffer/agent.rb
```diff
@@ -448,6 +449,20 @@ def initialize_messages(prompt_or_messages, files: nil)
...
+  def validate_seed_ids!(items)
+    strategy = Riffer.config.message_id_strategy
+    return if strategy == :none
+
+    items.each_with_index do |item, idx|
+      raw_id = item.is_a?(Hash) ? item[:id] : item.id
+      next unless raw_id.nil?
+      raise Riffer::ArgumentError,
+        "seeded message at index #{idx} is missing :id (required when Riffer.config.message_id_strategy = #{strategy.inspect})"
+    end
```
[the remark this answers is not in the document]
cloetzi: Addressed in [c9106e3](https://github.com/janeapp/riffer/pull/210/commits/c9106e38b1ebeb3e288d6b02ba59e648358c44f1).

lib/riffer/agent.rb
```diff
@@ -448,6 +449,24 @@ def initialize_messages(prompt_or_messages, files: nil)
...
   end
 
+  #--
+  #: (Array[Hash[Symbol, untyped] | Riffer::Messages::Base]) -> void
+  def validate_seed_ids!(items)
+    strategy = Riffer.config.message_id_strategy
+    return if strategy == :none
+
+    items.each_with_index do |item, idx|
+      raw_id = case item
```
cloetzi, line 459: I am not fully sold on this approach, initially I just naively extracted the id and it would have raised a no method error if there was another type passed. The robots suggested something like this, so I obeyed their wishes. Seems fine, just a bit odd to kind of passing through a non error (or raising here).

docs/08_MESSAGES.md
```diff
@@ -278,20 +278,50 @@ agent.generate(messages)
 
 Merging happens at serialization time only. The agent's `messages` array still contains the original separate messages for logging, evals, and debugging.
 
+## IDs
+
+Every message carries an optional `id` attribute. By default ids are disabled (`message.id` returns `nil` and `:id` is omitted from `to_h`). Enable them globally by setting `Riffer.config.message_id_strategy`:
```
Thibaut, line 283: suggestion (blocking): simplify things (all the `unless id.nil?` & whatnot) by making IDs always on and not able to be disabled. "If you don't need it, don't use it". Is there a harm to always generating IDs?

lib/riffer/agent.rb
Thibaut: What's the risk if we skip this validation altogether? (assuming IDs are always present, and we generate them as a fallback)

docs/08_MESSAGES.md
cloetzi: It would require to always generate an ID to seed a message history with IDs, and it becomes a breaking change. It also is overhead that might not be required. We could always auto-generate when no ID is present, but I would find that fairly unexpected and might to lead to collision in existing systems.

lib/riffer/agent.rb
cloetzi: It just might raise a NoMethod error. As said above, I'd prefer not to assume an ID is always present.

bottrall (approved):

Thibaut (approved):

PR #230, opened 2026-04-29
feat!: class-level tool resolution; lift activation tool to global config

cloetzi:
## Summary

- Adds `Riffer::Agent.resolved_tool_classes(context:)` — class-level companion to instance `#resolved_tools` that resolves the LLM-facing tool list without hitting the skills backend or running `prepare_run`. Lets callers compute the tool schema for an agent without instantiating it.
- Lifts the activation tool out of the adapter into global config (`Riffer.config.skill_activate_tool`) with a per-agent override inside the `skills` DSL (`skill_activate_tool MyTool`). Adapters now own only catalog rendering — the part that's actually provider-specific.
- Adapter takes `skill_activate_tool:` at construction; `render_catalog(skills)` is back to a single positional argument.
- Conflict check between agent tools and the activation tool moves out of `#resolved_tools` into the new class method, so it fires before the run loop instead of mid-call.

## Motivation

**Class-level tool resolution.** Out-of-process tool runtimes (e.g. Jane's Sidekiq runtime) need the LLM-facing tool list without instantiating the agent or paying for backend I/O. Today that requires duplicating the resolution logic in calling code (`compute_resolved_tool_classes`). This PR collapses that duplication: the call site becomes one line.

**Cleanly separating "which adapter" from "which activation tool".** The adapter selection is auto per-provider via `provider_class.skills_adapter` (Markdown for most, XML for Anthropic), with the existing per-agent override in the `skills` block. That behavior is preserved unchanged. The activation tool, by contrast, was speculatively coupled to the adapter — neither built-in adapter customizes it. Lifting the tool to global config matches how `tool_runtime` works (global default + per-agent override) and removes the need to re-derive the provider class just to find a default activation tool. Two orthogonal concerns that are now properly separated: adapter handles provider-specific catalog rendering; activation tool is a global concern that any agent can override per-instance.

## Breaking changes

- `Riffer::Skills::Adapter#activate_tool` removed. Configure via `Riffer.config.skill_activate_tool` or per-agent `skills do; skill_activate_tool MyTool; end`.
- `Riffer::Skills::Adapter#render_catalog(skills)` no longer takes an `activate_tool:` kwarg. The activation tool is set at construction: `Adapter.new(skill_activate_tool: ...)`. Custom adapters that override `#initialize` must call `super`.
- `Riffer::Skills::Context.new` no longer accepts a per-cycle activate tool — it lives on the adapter.

Both built-in adapters (`MarkdownAdapter`, `XmlAdapter`) inherit the new initializer; the agent passes the tool through automatically. Only downstream code that overrode `Adapter#activate_tool` or `#render_catalog`, or that constructed `Skills::Context` directly, needs updating.

The `bump-minor-pre-major` flag is added to release-please config in this PR so the breaking change bumps `0.24.x → 0.25.0` rather than `1.0.0`.

## Empty-skills-backend behavior (accepted trade-off)

`resolved_tool_classes` and instance `#resolved_tools` both delegate to the same logic — the tool list reflects class-level intent. When a `skills` block is configured but the backend yields no skills for this cycle (no `.skills/` directory, per-tenant skills with none for the current user, conditional Proc backend returning empty, etc.), the LLM still sees `skill_activate` in the schema even though there's no catalog in the system prompt.

Consequences:

- ~80 tokens/request for the tool definition.
- If the LLM calls it speculatively, the runtime guard in `Riffer::Skills::ActivateTool` returns `"Skills not configured"` — graceful error, recoverable in the loop.
- ~1 wasted round-trip per hallucinated call.

We accept this in favor of consistency between the class and instance views — earlier versions of this PR trimmed the activation tool at runtime when the backend was empty, but the divergence between "what `Agent.resolved_tool_classes` reports" and "what the run loop actually sends to the LLM" is worse than the modest token overhead. Downstream code that wants to filter the activation tool when its backend is empty can do so explicitly: `tool_classes.reject { |c| c == Riffer.config.skill_activate_tool }` — one line at the call site.

If this becomes a real source of pain, the right escape valve is probably an opt-in `read_backend: true` flag on the class method that does run the backend — not silent runtime divergence.

## Test plan

- [x] `bundle exec rake` passes (1282 runs, 1990 assertions, 0 failures, 0 errors)
- [x] `bundle exec rake standard` passes
- [x] `bundle exec steep check` passes
- [x] RBS regenerated (`bundle exec rake rbs:generate`)
- [x] New tests cover `.resolved_tool_classes`: behavior with/without a skills block, no-backend-I/O guarantee, per-agent activation tool override, class-level conflict raising
- [ ] Manual: integrate against Jane's `compute_resolved_tool_classes` and confirm the duplication collapses

## Summary by CodeRabbit

* **New Features**
  * App-wide defaults for skill activation tool and skills backend, with per-agent overrides.
  * Class-level tool resolution updated to include skill activation tooling and detect name conflicts.

* **Documentation**
  * Expanded guidance for configuring default activation tool and backend, plus adapter usage examples.

* **Tests**
  * Expanded integration and adapter tests covering configurable activation tooling, backend fallbacks, and class-level resolution.

### DIFF
--- a/docs/10_CONFIGURATION.md
+++ b/docs/10_CONFIGURATION.md
@@ -54,6 +54,18 @@ end
 
 Per-agent configuration overrides this global default. See [Advanced Tool Configuration — Tool Runtime](07_TOOL_ADVANCED.md#tool-runtime-experimental) for details.
 
+### Skill Activation Tool
+
+Override the tool the LLM calls to activate a skill. Defaults to `Riffer::Skills::ActivateTool`:
+
+```ruby
+Riffer.configure do |config|
+  config.skill_activate_tool = MyCustomActivateTool
+end
+```
+
+Per-agent override is available inside the `skills` block via `skill_activate_tool MyCustomActivateTool`. See [Skills — Custom Activation Tool](13_SKILLS.md#custom-activation-tool).
+
 ### Message ID Strategy
 
 Opt in to stable identifiers on every message for logging, persistence, or replay:
--- a/docs/10_CONFIGURATION.md
+++ b/docs/10_CONFIGURATION.md
@@ -70,6 +70,18 @@ end
 
 Per-agent override is available inside the `skills` block via `activate_tool MyCustomActivateTool`. See [Skills — Custom Activation Tool](13_SKILLS.md#custom-activation-tool).
 
+#### Default backend
+
+Set an app-wide default skills backend. Used by any agent that declares a `skills` block without specifying its own `backend`:
+
+```ruby
+Riffer.configure do |config|
+  config.skills.default_backend = Riffer::Skills::FilesystemBackend.new(".skills")
+end
+```
+
+Accepts a `Riffer::Skills::Backend` instance or a `Proc` that receives `context` and returns a backend. Defaults to `nil` — agents that don't set their own backend get no skills, matching pre-existing behavior. Per-agent backends override this default.
+
 ### Message ID Strategy
 
 Opt in to stable identifiers on every message for logging, persistence, or replay:
--- a/docs/10_CONFIGURATION.md
+++ b/docs/10_CONFIGURATION.md
@@ -64,7 +64,7 @@ Riffer.configure do |config|
 end
 ```
 
-Per-agent override is available inside the `skills` block via `skill_activate_tool MyCustomActivateTool`. See [Skills — Custom Activation Tool](13_SKILLS.md#custom-activation-tool).
+Per-agent override is available inside the `skills` block via `activate_tool MyCustomActivateTool`. See [Skills — Custom Activation Tool](13_SKILLS.md#custom-activation-tool).
 
 ### Message ID Strategy
 
--- a/docs/10_CONFIGURATION.md
+++ b/docs/10_CONFIGURATION.md
@@ -54,13 +54,17 @@ end
 
 Per-agent configuration overrides this global default. See [Advanced Tool Configuration — Tool Runtime](07_TOOL_ADVANCED.md#tool-runtime-experimental) for details.
 
-### Skill Activation Tool
+### Skills
+
+Skills-related global configuration lives under `config.skills`.
+
+#### Default activation tool
 
 Override the tool the LLM calls to activate a skill. Defaults to `Riffer::Skills::ActivateTool`:
 
 ```ruby
 Riffer.configure do |config|
-  config.skill_activate_tool = MyCustomActivateTool
+  config.skills.default_activate_tool = MyCustomActivateTool
 end
 ```
 
--- a/docs/13_SKILLS.md
+++ b/docs/13_SKILLS.md
@@ -125,24 +125,39 @@ end
 
 ## Custom Adapters
 
-Subclass `Riffer::Skills::Adapter` to customize how the skill catalog is rendered and which tool the LLM uses to activate skills:
+Subclass `Riffer::Skills::Adapter` to customize how the skill catalog is rendered in the system prompt:
 
 ```ruby
 class CustomAdapter < Riffer::Skills::Adapter
   def render_catalog(skills)
     # Return String (skill catalog for the system prompt)
-    # Use `activate_tool.name` to reference the activation tool
-  end
-
-  def activate_tool
-    # Return a Riffer::Tool subclass
-    # Defaults to Riffer::Skills::ActivateTool
-    Riffer::Skills::ActivateTool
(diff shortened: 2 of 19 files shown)


docs/13_SKILLS.md
```diff
@@ -125,24 +125,39 @@ end
...
+class MyAgent < Riffer::Agent
+  skills do
+    backend Riffer::Skills::FilesystemBackend.new(".skills")
+    skill_activate_tool MyCustomActivateTool
   end
 end
 ```
 
-The built-in adapters are `Riffer::Skills::MarkdownAdapter` (default) and `Riffer::Skills::XmlAdapter` (used by Anthropic).
+The default is `Riffer::Skills::ActivateTool`. Custom tools must subclass `Riffer::Tool` and accept a `name:` parameter that resolves to a skill body via `context[:skills]`.
```
bottrall: **Question(non-blocking):** Would it make more sense to suggest inheriting from `Riffer::Skills::ActivateTool`? That way they don't need to worry about changing the parameters.

bottrall (approved):

docs/13_SKILLS.md
cloetzi: Good call — updated in 54ef426. The doc now recommends subclassing `Riffer::Skills::ActivateTool` by default (so identifier/description/params/timeout come along for free) and only drops to subclassing `Riffer::Tool` directly if a different parameter shape is needed. Example shows wrapping the default with `super`.

docs/13_SKILLS.md
```diff
@@ -125,24 +125,39 @@ end
...
-    Riffer::Skills::ActivateTool
+```ruby
+# Global default
+Riffer.config.skill_activate_tool = MyCustomActivateTool
+
+# Per-agent override
+class MyAgent < Riffer::Agent
+  skills do
+    backend Riffer::Skills::FilesystemBackend.new(".skills")
+    skill_activate_tool MyCustomActivateTool
```
Thibaut: suggestion (non-blocking): `activate_tool` since we are already in `skills do end`

docs/10_CONFIGURATION.md
```diff
@@ -54,6 +54,18 @@ end
 
 Per-agent configuration overrides this global default. See [Advanced Tool Configuration — Tool Runtime](07_TOOL_ADVANCED.md#tool-runtime-experimental) for details.
 
+### Skill Activation Tool
+
+Override the tool the LLM calls to activate a skill. Defaults to `Riffer::Skills::ActivateTool`:
+
+```ruby
+Riffer.configure do |config|
+  config.skill_activate_tool = MyCustomActivateTool
```
Thibaut: suggestion (non-blocking): how about `config.skills.default_activate_tool` and `config.skills.default_backend` for maximum flexibility?

Thibaut (approved):

docs/13_SKILLS.md
cloetzi: Done in 66c82ea — `Skills::Config` now exposes `#activate_tool` (matching the existing prefix-free `backend`/`adapter`/`activate` DSL methods). The global `Riffer.config.skills.default_activate_tool` keeps the qualifier since it lives at config root, where the namespace makes the prefix meaningful.

docs/10_CONFIGURATION.md
cloetzi: Adopted in two commits:

- 43ca0aa: moved `Riffer.config.skill_activate_tool` → `Riffer.config.skills.default_activate_tool`. New `Riffer::Config::Skills` sub-config object groups skills-related globals.
- 372080e: added `Riffer.config.skills.default_backend` — app-wide default skills backend that agents fall back to when they declare a `skills` block without their own `backend`. Purely additive (defaults to nil); per-agent backends take precedence.

lib/riffer/config.rb
```diff
@@ -61,6 +61,24 @@ def tool_runtime=(value)
     @tool_runtime = value
   end
 
+  # Global skill activation tool class.
+  #
+  # The tool class the LLM calls to activate a skill. Defaults to
+  # <tt>Riffer::Skills::ActivateTool</tt>. Per-agent override is available
+  # via <tt>skills do; activate_tool ...; end</tt>.
```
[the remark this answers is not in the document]
cloetzi: Auto-resolved by 66c82ea — the per-agent DSL is now `activate_tool` (matching pre-rename name), so the docstring is correct as written. The global config also moved to `Riffer.config.skills.default_activate_tool` in 43ca0aa.

Thibaut (approved):
Solid!

PR #235, opened 2026-05-01
fix: close streaming HTTP connections when consumer raises

cloetzi:
## Summary

Anthropic and OpenAI streaming providers leak the underlying HTTP socket when the consumer fiber is cancelled or raises mid-stream. The SDKs both expose an idempotent `BaseStream#close` that fires the Net::HTTP cleanup path, but it is never invoked unless iteration completes naturally — so on `Async::Stop` or any handler raise, the producer fiber stays suspended and the socket sits in `CLOSE_WAIT` until GC. This was observed downstream in Maestro bench, with `CLOSE_WAITs` piling up against `api.anthropic.com`.

Wraps `stream.each` in `begin / ensure stream.close` for both providers. The SDK's `close` is documented (and verified) to be idempotent and a no-op after EOF.

### Provider matrix

| Provider | Status | Reason |
|---|---|---|
| **anthropic** | Fixed | SDK has `BaseStream#close` (idempotent, EOF-safe), no auto-close on raise. |
| **open_ai** | Fixed | Same `BaseStream` shape as Anthropic. |
| **azure_open_ai** | Fixed transitively | Inherits `execute_stream` from `OpenAI`. |
| **amazon_bedrock** | Already safe | Block-form `converse_stream(...) { |stream| ... }` is wrapped in Seahorse's `session_for` ensure (`session.finish` runs on raise). |
| **gemini** | Already safe | `Net::HTTP.start(...) { |http| ... }` runs `http.finish` in an ensure when the block exits. |
| **mock** | Skipped | No external resource. |

## Test plan

- [x] `bundle exec rake` (1308 runs, 0 failures, lint clean, steep clean)
- [x] New unit tests assert `stream.close` runs once when iteration raises mid-stream (Anthropic + OpenAI)
- [x] New unit tests assert `stream.close` runs exactly once on the happy path (no double-close after EOF)
- [x] No behavior change on happy path — same events delivered in same order

## Summary by CodeRabbit

* **Bug Fixes**
  * Streaming connections to AI providers are now deterministically closed after use, both on normal completion and when errors or cancellations occur, preventing resource leaks.

* **Behavior**
  * Token usage is now reported only when available and emitted reliably at stream end, improving accuracy of usage reporting.

* **Tests**
  * Added tests to verify stream resources are closed exactly once across success and failure scenarios.

### DIFF
--- a/lib/riffer/providers/anthropic.rb
+++ b/lib/riffer/providers/anthropic.rb
@@ -152,28 +152,35 @@ class Riffer::Providers::Anthropic < Riffer::Providers::Base
     )
     current_state[:stream] = stream
 
-    stream.each do |event|
-      case event
-      when Anthropic::Models::RawContentBlockStartEvent
-        handle_raw_content_block_start(event, state: current_state)
-      when Anthropic::Models::RawContentBlockDeltaEvent
-        handle_raw_content_block_delta(event, state: current_state)
-      when Anthropic::Streaming::TextEvent
-        handle_text_event(event, state: current_state, yielder: yielder)
-      when Anthropic::Streaming::ThinkingEvent
-        handle_thinking_event(event, state: current_state, yielder: yielder)
-      when Anthropic::Streaming::InputJsonEvent
-        handle_input_json_event(event, state: current_state, yielder: yielder)
-      when Anthropic::Streaming::ContentBlockStopEvent
-        block_type = event.content_block&.type.to_s
-        handle_content_block_stop_text(event, state: current_state, yielder: yielder) if block_type == "text" && current_state[:text]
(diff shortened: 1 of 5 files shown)


lib/riffer/providers/anthropic.rb
```diff
@@ -150,30 +150,36 @@ def execute_stream(params, yielder)
       **params,
       request_options: {extra_headers: {"accept-encoding" => "identity"}}
     )
-    current_state[:stream] = stream
```
cloetzi, line 153: This was technically not required, but it really bothered me to not be sure what side effect closing stream has if it is part of a hash. This was only used in one place and was easy to remove.

bottrall (approved):

PR #244, opened 2026-05-06
feat: accept StructuredOutput on Agent.structured_output and add from_json_schema

cloetzi:
## Summary
- `Agent.structured_output` now accepts a `Riffer::StructuredOutput` in addition to a `Riffer::Params` or block DSL — `resolve_structured_output` no longer double-wraps when one is already passed.
- New `Riffer::StructuredOutput.from_json_schema(schema)` constructor for pre-built JSON Schemas. The schema is forwarded to the provider verbatim; client-side `parse_and_validate` only checks that the response parses as JSON, since the provider enforces the schema server-side.
- Together these remove the need to subclass `Riffer::Params` and reach for `instance_variable_set` to feed Agent a raw JSON Schema.

## Why
The setter previously required a `Riffer::Params`, but the actual runtime contract used by providers and `Agent#validate_structured_output` is `Riffer::StructuredOutput` (`json_schema(strict:)` + `parse_and_validate`). `Params` is the DSL builder, not the runtime type — surfacing `StructuredOutput` as a public input makes that explicit and unblocks consumers whose schemas are generated externally.

## Notes
- Backwards compatible: existing block DSL and `Params`-instance call sites are unchanged. The argument-error message string is updated to reflect the broader accepted set; one test asserting the old message is updated.
- `RawJsonSchema` (the adapter behind `from_json_schema`) is a `private_constant` — not part of the public surface.
- RBS sigs regenerated via `bundle exec rake rbs:generate`.

## Test plan
- [x] `bundle exec rake` — 1396 runs, 0 failures, lint clean, Steep clean
- [x] New tests cover `from_json_schema` (schema returned verbatim, JSON parse path, no client-side schema validation, parse-error surface) and Agent accepting a `StructuredOutput` (setter + end-to-end `#generate`)
- [ ] Manual: smoke-test against a real provider with a pre-built schema (left for the consumer)

## Summary by CodeRabbit

* **New Features**
  * Expanded structured_output configuration to accept StructuredOutput instances alongside parameter objects
  * Added support for initializing structured output from pre-built JSON Schemas

* **Documentation**
  * Updated guides with examples and instructions for using pre-built JSON Schemas with structured output

### DIFF
--- a/lib/riffer/structured_output.rb
+++ b/lib/riffer/structured_output.rb
@@ -21,6 +21,29 @@ class Riffer::StructuredOutput
     @params = params
   end
 
+  # Builds a StructuredOutput from a pre-built JSON Schema, bypassing the
+  # Params DSL.
+  #
+  # Useful when the schema is generated externally (e.g. from a Pydantic-style
+  # model, dumped JSON, or another tool). The returned object exposes the same
+  # +json_schema+ and +parse_and_validate+ surface as a Params-backed
+  # StructuredOutput. Client-side validation only checks that the LLM produced
+  # parseable JSON whose top level is an object — schema enforcement is left
+  # to the LLM provider, which receives the raw schema verbatim.
+  #
+  #   schema = JSON.parse(File.read("sentiment.schema.json"), symbolize_names: true)
+  #   so = Riffer::StructuredOutput.from_json_schema(schema)
+  #   class SentimentAgent < Riffer::Agent
+  #     model "openai/gpt-5-mini"
+  #     structured_output so
+  #   end
+  #
+  #--
+  #: (Hash[Symbol, untyped]) -> Riffer::StructuredOutput
+  def self.from_json_schema(schema)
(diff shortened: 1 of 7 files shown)


lib/riffer/structured_output.rb
```diff
@@ -44,4 +67,31 @@ def parse_and_validate(json_string)
...
     Result.new(error: "Validation error: #{e.message}")
   end
+
+  # Adapter used by StructuredOutput.from_json_schema to make a raw JSON
+  # Schema look like a Riffer::Params for the two methods StructuredOutput
+  # calls on it. Validation is intentionally a pass-through: the LLM provider
+  # already enforces the schema server-side, and bringing in a JSON Schema
+  # validator on the client would be a heavier dependency than this entry
+  # point warrants.
+  class RawJsonSchema
```
cloetzi, line 77: It feels like a hack to use this proxy class, but none of the alternatives seemed better.

PR #247, opened 2026-05-08
feat!: forward assistant_message to ToolRuntime hooks

cloetzi:
## Summary

- Adds an `assistant_message:` kwarg to `Riffer::ToolRuntime#execute`, `#around_tool_call`, and `#dispatch_tool_call`. The agent threads the originating `Riffer::Messages::Assistant` through both runtime call sites (`execute_tool_calls` and `execute_pending_tool_calls`).
- Custom runtimes can now reach the assistant turn — accompanying text, sibling tool calls, etc. — when wrapping or dispatching individual tool calls. The kwarg is **not** forwarded to `Tool#call`; tools that need it must opt in via a `dispatch_tool_call` override.
- `pending_tool_calls` now returns `[assistant, calls]` so `execute_pending_tool_calls` can forward the message without a second scan over `@messages`.

## Breaking change

Custom subclasses of `Riffer::ToolRuntime` that override `around_tool_call` or `dispatch_tool_call` **must** accept the new `assistant_message:` kwarg (or `**kwargs`). Existing overrides without it will raise `ArgumentError: unknown keyword: :assistant_message` at runtime — Ruby does not silently absorb unknown keyword arguments. Direct callers of `runtime.execute(...)` are unaffected; the kwarg defaults to `nil`.

Migration:

```ruby
# Before
def around_tool_call(tool_call, context:)
  ...
end

# After
def around_tool_call(tool_call, context:, assistant_message: nil)
  ...
end
```

## Test plan

- [x] `bundle exec rake` passes (1392 runs, 0 failures, steep clean, standard clean)
- [x] New tests cover: assistant_message threading into `around_tool_call`, threading into `dispatch_tool_call`, and the default-`nil` case for direct `execute` callers
- [x] Existing override tests updated to accept `**kwargs`, demonstrating the migration path

## Summary by CodeRabbit

* **Documentation**
  * Updated custom runtime documentation to reflect new `assistant_message:` parameter in tool execution hook signatures.

* **Chores**
  * Internal changes to forward assistant message through tool execution pipeline. Custom runtime implementations must update hook signatures to accept `assistant_message:` (or `**kwargs`) to prevent errors.

### DIFF
--- a/docs/07_TOOL_ADVANCED.md
+++ b/docs/07_TOOL_ADVANCED.md
@@ -215,7 +215,7 @@ Create a custom runtime by subclassing `Riffer::ToolRuntime` and overriding the 
 class HttpToolRuntime < Riffer::ToolRuntime
   private
 
-  def dispatch_tool_call(tool_call, tools:, context:)
+  def dispatch_tool_call(tool_call, tools:, context:, assistant_message: nil)
     # Dispatch tool execution to an external service
     response = HttpClient.post("/tools/execute", {
       name: tool_call.name,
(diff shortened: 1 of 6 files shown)


bottrall (approved):
Only thought I have is whether or not there would be value in providing the entire message array, as opposed to just the last assistant message. But this looks fine for now.

PR #253, opened 2026-05-13
MCP progressive discovery

kepheusikarius:
## Summary
This addresses concerns from #180. `use_mcp` now defaults to progressive discovery. Instead of injecting every tool schema up front, the agent receives two meta-tools — `riffer__mcp_search` and `riffer__mcp_call` — and the LLM discovers tools on demand.

## What changed
- `use_mcp :tag` defaults to progressive mode; opt out with `progressive: false` for small, stable servers
- `riffer__mcp_search` — filters the tool list by name or description substring, returns matching names, descriptions, and schemas
- `riffer__mcp_call` — invokes a tool by exact name with JSON-encoded arguments
- Multiple progressive `use_mcp` calls share one search/call pair across all registrations
- `riffer__mcp_search` and `riffer__mcp_call` are reserved; mixing them with a static tool raises `Riffer::ArgumentError`

## Design notes

- **Two focused tools over a single dispatch tool** — A single tool with an `action` param (considered and rejected) requires nullable fields for strict-mode providers (OpenAI). Two tools each have simple, required-only schemas.
- **`query` is required in the search schema** — This keeps the schema strict-mode clean without `anyOf` nullable types. An empty string returns all tools; the implementation treats `nil` and `""` equivalently.
- **`arguments` is a JSON-encoded string** — Strict-mode providers prohibit open-ended object types. Using a string sidesteps this while passing arbitrary args through to inner tools.

## Known tradeoffs

- A broad or empty `query` against a large tool union returns all schemas, which can be expensive. Documented in `docs/14_MCP.md`. This may be mitigated in the future by returning names/descriptions only by default, with schemas available on demand, when needed.

## Summary by CodeRabbit

* **New Features**
  * Progressive MCP discovery enabled by default with mcp_search and mcp_call tools; per-registration progressive flag supported.

* **Documentation**
  * Expanded MCP docs covering progressive workflow, credential callback behavior, token-cost guidance, and reserved tool identifier guidelines.

* **Tests**
  * Added and updated tests for progressive/non‑progressive flows, credential handling, search/call tools, and validation behavior.


lib/riffer/mcp/discovery_tool_factory.rb
```diff
@@ -0,0 +1,142 @@
...
+# of MCP tools to the LLM.
+#
+# Instead of loading all tool schemas up front, an agent configured with
+# +use_mcp :tag+ receives +riffer__mcp_search+ to find tools by name or
+# description, and +riffer__mcp_call+ to invoke them by name with
+# JSON-encoded arguments.
+#
+# See Riffer::Agent.use_mcp and docs/14_MCP.md.
+#
+module Riffer::Mcp::DiscoveryToolFactory
```
bottrall: **Suggestion(blocking):** I think this is leaning too heavily into pattern created in the previous PR with `Riffer::Mcp::ToolFactory`.



There's a need for a tool factory in regards to the mcp tools, because we have to define them at runtime. However we know the below tools ahead of time, so we can should be able to define them like any other tool. See `lib/riffer/skills/activate_tool.rb` for reference of utilizing context to provide the tools list to the tools.

bottrall (commented):
Thanks @kepheusikarius.

lib/riffer/mcp/discovery_tool_factory.rb
kepheusikarius: Fair point. My original reasoning ruled out passing tools during construction, but it didn't consider call-time context. I'll update to match `activate_tool`, so tools are read from `context[:mcp_progressive_tools]`.

lib/riffer/mcp/call_tool.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    target.new.call(context: context, **parsed)
+  end
+
+  private
+
+  #: (String?) -> (Hash[Symbol, untyped] | Riffer::Tools::Response)
+  def parse_arguments(arguments)
+    return {} if arguments.nil? || arguments.strip.empty?
+
```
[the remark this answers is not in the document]
kepheusikarius: Tool calls are made with `call_with_validation`. Tests were added asserting this path is safe.

lib/riffer/mcp/search_tool.rb
```diff
@@ -0,0 +1,55 @@
...
+    text(format_matches(matches))
+  end
+
+  private
+
+  #: (Array[singleton(Riffer::Tool)], String) -> Array[singleton(Riffer::Tool)]
+  def filter(tools, query)
+    return tools if query.strip.empty?
+    q = query.downcase
+    tools.select { |t| t.name.downcase.include?(q) || t.description.to_s.downcase.include?(q) }
```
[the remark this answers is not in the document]
kepheusikarius: Tool calls are made with `call_with_validation`. Tests were added asserting this path is safe.

lib/riffer/agent.rb
```diff
@@ -657,23 +661,31 @@ def resolve_mcp_tool_classes
...
-  #
-  #: (Array[Hash[Symbol, untyped]]) -> Hash[Riffer::Mcp::Registration, Array[Symbol]]
-  def gather_mcp_registrations_with_tags(configs)
-    by_reg = {}
+    reg_tags = {} #: Hash[Riffer::Mcp::Registration, Array[Symbol]]
+    progressive_reg_tags = {} #: Hash[Riffer::Mcp::Registration, Array[Symbol]]
+
     configs.each do |cfg|
+      target = cfg[:progressive] ? progressive_reg_tags : reg_tags
       Riffer::Mcp::Registry.find_by_tags(cfg[:tags]).each do |reg|
```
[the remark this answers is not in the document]
kepheusikarius: This is enforce in the config. Adding defensive coding here means the default value is in two places.

kepheusikarius:
@coderabbitai review

kepheusikarius:
@bottrall Can I get another pass on this one? If it's still desirable, I'd love to get this in before I have to rebase again 😅

lib/riffer/agent/context.rb
```diff
@@ -1,13 +1,31 @@
...
+# write — +skills+ must be a +Riffer::Skills::Context+ (or nil);
+# +token_usage+ must be a +Riffer::Providers::TokenUsage+ (or nil);
+# +mcp_progressive_tools+ must be an Array of +Riffer::Tool+ subclasses
+# (or nil).
+#
+#   context = Riffer::Agent::Context.new(user_id: 42)
+#   context[:user_id]    # => 42
+#   context.skills       # => nil
+#   context.token_usage  # => nil
+#
```
bottrall: **Suggestion(blocking):** Please revise this comment based on the latest guidance in `.agents/code-style.md`.



I just did a massive cleanup of all the comments in https://github.com/janeapp/riffer/pull/297, to get rid of overly verbose ai generated comments like this, so I would love to avoid them creeping back into the repo 😅

Same comment applies to a bunch of other comments in this PR.

lib/riffer/mcp/call_tool.rb
```diff
@@ -0,0 +1,62 @@
...
+# Reads the per-agent set of auth-wrapped MCP tool classes from
+# +context[:mcp_progressive_tools]+ (populated by Riffer::Agent when a
+# progressive +use_mcp+ matches at least one registration).
+#
+# Arguments are passed as a JSON-encoded string because strict-mode
+# providers prohibit open-ended object schemas.
+#
+# See Riffer::Agent.use_mcp and docs/14_MCP.md.
+class Riffer::Mcp::CallTool < Riffer::Tool
+  IDENTIFIER = "mcp_call" #: String
```
bottrall: **Nitpick(non-blocking):** Explicit rbs inline isn't necessary here, `String` will be inferred.

lib/riffer/agent.rb
```diff
@@ -111,12 +111,52 @@ def self.uses_tools(value = nil)
...
+  # When +uses_tools+ is a Proc, +context+ is forwarded to it.
+  #
+  # Each returned tool class is validated via +validate_as_tool!+.
+  #
+  # Raises Riffer::ArgumentError on tool name conflicts with the skill
+  # activation tool, or when a tool class fails +validate_as_tool!+.
+  #
+  #--
+  #: (?context: Hash[Symbol, untyped]?, ?config: Riffer::Agent::Config) -> Array[singleton(Riffer::Tool)]
+  def self.resolved_tool_classes(context: nil, config: self.config)
```
bottrall: **Suggestion(blocking):** Remove this method.



Recent work has made it easier and more ergonomic to grab the resolved tools of an agent by doing this:
```rb
tools = MyAgent.new(context: my_context).tools
```
This is because we moved the config resolution to agent instantiation, as opposed to it previously being bound to generation.

Therefore this is no longer necessary.

bottrall (changes requested):
Thanks @kepheusikarius and sorry wouldn't normally block on commenting, but since we only just fixed up all the comments I think its worth being a little stricter for the foreseeable future.

One other thing that might be worth having a conversation about is the pattern for progressive discovery. I didn't think too much about it in the last review, but I got thinking in this review about what patterns for progressive discovery of MCP are being adopted by the wider industry. From what I can tell there are two patterns emerging:
- Pattern A — dispatch/proxy meta-tool (what the PR does)
- Pattern B — discover → inject native tools

I asked clanker claude to do some research and figure out where each framework is landing. I haven't fact checked it so take it with a grain of salt:

> ### Pattern A — dispatch/proxy meta-tool
> 
> This is essentially the **MCP gateway/router** family. They expose a few generic tools and route by a `tool_name` argument — functionally identical to the PR's `mcp_search` + `mcp_call`, just usually running as a separate proxy process rather than in-process:
> 
> - **[RaiAnsar/mcp-gateway](https://github.com/RaiAnsar/mcp-gateway)** — the closest analogue. A lazy-loading proxy exposing `gw_discover` (search) + `gw` (dispatch by tool name) + `gw_batch`. Claims ~95% context-token reduction by loading on demand. `gw(tool_name, args)` *is* `mcp_call`.
> - **[ContextForge MCP Gateway](https://medium.com/@crivetimihai/mcp-gateway-the-missing-proxy-for-ai-tools-2b16d3b018d5)** (IBM/Mihai Criveti) — registry + proxy aggregating many servers behind a unified dispatch surface.
> - **[Microsoft mcp-gateway](https://github.com/microsoft/mcp-gateway)** and **[block-town MCP Gateway](https://glama.ai/mcp/servers/block-town/mcp-gateway)** — reverse-proxy/routing layers in the same shape.
> - **[LiteLLM MCP](https://docs.litellm.ai/docs/mcp)** and **[Docker MCP Gateway](https://docs.docker.com/ai/mcp-catalog-and-toolkit/mcp-gateway/)** — aggregate N servers, present a unified catalog, route calls to the owning backend.
> 
> Note these share the PR's exact tradeoff: the inner tool's arguments get passed *through* the dispatch tool (as a nested object/string), so the provider doesn't enforce the real tool's schema. The whole gateway category is Pattern A.
> 
> ### Pattern B — discover → inject native tools
> 
> This is the **model-platform / agent-framework** family. Discovery returns real tool definitions that the model then calls natively:
> 
> - **[Anthropic Tool Search Tool](https://www.anthropic.com/engineering/advanced-tool-use)** — mark long-tail tools `defer_loading: true`; the model sees only the search tool + critical tools; searching loads 3–5 real defs that are then called as normal `tool_use` blocks. No dispatch wrapper.
> - **Claude Code's MCP tool search** (Jan 2026) — same mechanism applied to MCP; this environment's own `ToolSearch` works this way (search by name → fetch schema → call natively).
> - **[LlamaIndex `ObjectIndex` + `ObjectRetriever`](https://developers.llamaindex.ai/python/examples/agent/openai_agent_retrieval/)** — index your `Tool` objects, retrieve the relevant ones at query time, hand them to the agent as first-class tools. The canonical "retrieval-augmented agent."
> - **[LangChain tool retrieval](https://python.langchain.com/api_reference/community/retrievers/langchain_community.retrievers.llama_index.LlamaIndexRetriever.html)** — retrieve relevant tools from a vector store and `bind` them to the model for that call.
> - **[Hermes Agent](https://www.marktechpost.com/2026/05/29/hermes-agent-ships-tool-search-for-mcp-anthropic-evals-show-49-to-74-accuracy-gain-on-opus-4/)** — ships Anthropic-style tool search for MCP, citing the 49%→74% Opus-4 eval gain.
> 
> ## The pattern that's worth calling out
> 
> The interesting tell is that **Pattern A is almost always a *separate gateway process*, not a library's native agent abstraction.** Gateways reach for dispatch because they sit *outside* the model loop — they can't mutate the caller's tool list, so a proxy tool is the only lever they have. When a framework *owns* its own agent loop (LlamaIndex, LangChain, Anthropic's platform), it does Pattern B, because it *can* inject tools and gets native schema validation for free.
> 
> Riffer owns its agent loop. So it's in the Pattern-B position architecturally but has implemented the Pattern-A design that gateways use out of necessity. That's the crux for your review: the PR adopts the workaround that exists specifically for components that *can't* do what Riffer can.
> 
> Sources:
> - [RaiAnsar/mcp-gateway (lazy-loading dispatch proxy)](https://github.com/RaiAnsar/mcp-gateway)
> - [ContextForge MCP Gateway](https://medium.com/@crivetimihai/mcp-gateway-the-missing-proxy-for-ai-tools-2b16d3b018d5)
> - [Anthropic — advanced tool use / Tool Search Tool](https://www.anthropic.com/engineering/advanced-tool-use)
> - [LlamaIndex — Retrieval-Augmented Agents (ObjectIndex/ObjectRetriever)](https://developers.llamaindex.ai/python/examples/agent/openai_agent_retrieval/)
> - [Hermes Agent ships Tool Search for MCP](https://www.marktechpost.com/2026/05/29/hermes-agent-ships-tool-search-for-mcp-anthropic-evals-show-49-to-74-accuracy-gain-on-opus-4/)

There would be added complexity by moving to "discover → inject native tools" as we would need some sort of mechanism for changing a tool list mid session and also a way of expressing that in the message history. Food for thought, but we can also use "dispatch/proxy meta-tool" as a stepping stone if we think it's too large of a lift.

kepheusikarius:
> One other thing that might be worth having a conversation about is the pattern for progressive discovery.

Great call out. Didn't seem like a huge lift, so I added it [in this commit](https://github.com/janeapp/riffer/pull/253/changes/d791c6cd9bf77640ece9c2578198f9cc8f6de89a). Certainly feels a little more straight-forward.

If we feel it's complicating the run, we can drop that commit and go back to dispatch/proxy.

bottrall (approved):
Awesome, thanks @kepheusikarius!

PR #265, opened 2026-05-20
refactor: extract Riffer::Session and rework Agent constructor (CL2-776)

bottrall:
## Summary

- Extract conversation state from `Riffer::Agent` into a new `Riffer::Session` class (`lib/riffer/session.rb`) with a uniform mutation API — `add`, `set`, `unset`, `remove`, `update(id:|tool_call_id:, **attrs)` — plus `Enumerable`, the `on_message` callback list, and the two invariant queries (`orphaned_tool_call_ids`, `pending_tool_calls`).
- Rework `Agent.new` to accept `session:` and `context:`. Without a session it eagerly seeds `[sys, skills].compact` using the constructor-time `context`. With a session it uses it as-is and applies `experimental_history_healing` at init when the flag is on.
- Simplify `Agent#generate` / `Agent#stream` to `(prompt = nil, files: nil)` — no more `prompt_or_messages` discriminated union, no more per-call `context:`. With a prompt, the new user message is silently appended; without one, the loop runs against the current session (resume case). Class-method `Agent.generate` / `Agent.stream` keep `context:` as a convenience that forwards into `new(context:)`.
- Delete `seed_session`, `seed_session_from_array`, `continue_session`, `start_session`, `validate_seed_ids!`, `prepare_run`, `clear_resolved_model`, `extract_final_response`, and all message-API methods on Agent — every caller goes through `agent.session.*` now.
- Cross-process resume: `Riffer::Session.new(messages: persisted)` → `Agent.new(session: session, context: ctx)` → `agent.generate`. The discriminated-union `agent.generate(array_or_string)` path is gone.
- Test sweep: deleted obsolete describe blocks (array-of-messages, array-of-hashes, seed-id-validation), rewrote `resume via generate -> with persisted messages` to use seeded `Riffer::Session`, swept ~110 `agent.messages` / `agent.on_message` references to `agent.session.*`, repointed ~14 `context:` call sites to constructor-context, added a `test/riffer/session_test.rb` covering the new API surface (40+ specs including `#set`/`#unset` callback contracts and Enumerable behavior).
- Docs sweep across `04_AGENT_LIFECYCLE.md`, `08_MESSAGES.md`, `10_CONFIGURATION.md`, `03_AGENTS.md`, `06_TOOLS.md`, `07_TOOL_ADVANCED.md`, `providers/06_MOCK_PROVIDER.md`.

**Breaking changes for consumers (maestro, jane):**

- `agent.messages` / `agent.add_message` / `agent.message_by_id` / `agent.tool_message_for` / `agent.last_assistant` / `agent.orphaned_tool_call_ids` / `agent.replace_assistant_content` / `agent.remove_message` / `agent.replace_tool_result` / `agent.on_message` → all moved to `agent.session.*`. `replace_assistant_content(id:, content:)` is now `session.update(id:, content:)` (with explicit `session.remove(id:)` if you want the old empty-content cascade-to-remove behavior). `replace_tool_result(tool_call_id:, ...)` is now `session.update(tool_call_id:, ...)`.
- `agent.generate(persisted_array, context: ctx)` → `Agent.new(session: Riffer::Session.new(messages: persisted_array), context: ctx).generate`.
- `agent.generate("prompt", context: ctx)` (per-call context) → `Agent.new(context: ctx).generate("prompt")` (context fixed at init). Class-method `MyAgent.generate("prompt", context: ctx)` still works.

The CL2-776 Jira DoD's "Maestro contract forwarders" clause does not match this PR — to be updated separately. Maestro sweep and Jira update are tracked as out-of-band closeout items.

## Test plan

- [ ] Update maestro to call `agent.session.*` instead of `agent.*` for the message API, and construct sessions via `Riffer::Session.new(messages: ...)` for cross-process resume. Coordinate the riffer release with maestro's update.
- [ ] Update Jira CL2-776 DoD to reflect the no-forwarders direction and Session-based resume.
- [ ] Check CL2-779 (Phase 3: Extract Riffer::Run) for any wording that assumed the Phase 1 forwarders would survive.

CI covers: full test suite (1441 tests, `bundle exec rake test`), standard lint (`bundle exec rake standard`), Steep type check, RBS generation.

## Summary by CodeRabbit

* **New Features**
  * Session-backed conversation model centralizes message history, pending tool-call tracking, and resume behavior.

* **Refactor**
  * Agents now hold context for their lifetime at initialization; per-call context removed.
  * Callbacks and history access are session-centric (register callbacks on the session; read/update/remove history via session APIs).
  * Supplying a prompt appends a user message; omitting it continues the current session.

* **Documentation**
  * Guides and examples updated for session-centric APIs, interrupt/resume, tools/context, and history mutation patterns.

* **Tests**
  * Test suites updated to assert session-based behavior and examples.

[Review Change Stack](https://app.coderabbit.ai/change-stack/janeapp/riffer/pull/265?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/04_AGENT_LIFECYCLE.md
+++ b/docs/04_AGENT_LIFECYCLE.md
@@ -211,16 +211,16 @@ agent.stream.each do |event|
 end
 ```
 
-### Building System Messages for Persistence
+### Reading System Messages for Persistence
 
-Use `generate_instruction_message` and `generate_skills_message` to generate system messages independently. This is useful for database persistence workflows where you need to store and later reconstruct message histories.
+Read the agent's instruction and skills system messages from `agent.instruction_message` and `agent.skills_message`. Both are built once at `Agent.new` time using the constructor `context:` and cached — they reflect the agent's configured `instructions` and `skills` DSL output. Useful for database persistence workflows where you need to store and later reconstruct message histories.
 
-Both methods return a `Riffer::Messages::System` or `nil` (when unconfigured). They accept an optional `context:` keyword that overrides the agent's init context for a one-off evaluation.
+Both return `Riffer::Messages::System` or `nil` (when unconfigured / empty).
 
 ```ruby
 agent = MyAgent.new(context: ctx)
-sys = agent.generate_instruction_message              # => Riffer::Messages::System or nil
-skills = agent.generate_skills_message                # => Riffer::Messages::System or nil
+sys = agent.instruction_message     # => Riffer::Messages::System or nil
+skills = agent.skills_message       # => Riffer::Messages::System or nil
 
 # Store in DB, then later resume in a new process:
 session = Riffer::Session.new(messages: [sys, skills, user_msg].compact)
--- a/docs/04_AGENT_LIFECYCLE.md
+++ b/docs/04_AGENT_LIFECYCLE.md
@@ -1,69 +1,64 @@
 # Agent Lifecycle
 
+## Construction
+
+### Agent.new
+
+```ruby
+Agent.new(session: nil, context: nil)
+```
+
+- **`session:`** — an existing `Riffer::Session`. When given, the agent uses it as-is (no system/skills seeding). Typical use case: cross-process resume from persisted history. With `Riffer.config.experimental_history_healing` on, a provided session is healed at construction time so the `tool_use` ↔ `tool_result` invariant holds before the next inference call.
+- **`context:`** — a `Hash` carried for the lifetime of the agent. Used to evaluate Proc-based `instructions`, `model`, `uses_tools`, and skill activation at construction time, and threaded through tool execution and guardrails on every `generate`/`stream` call.
+
+When `session:` is omitted, the agent constructs a fresh session and seeds it with `[instruction_message, skills_message].compact` eagerly. To swap context, construct a new agent — context is fixed for the lifetime of an agent instance.
+
 ## Instance Methods
 
 ### generate
 
 Generates a response synchronously. Returns a `Riffer::Agent::Response` object.
 
-The behavior depends on what you pass and the agent's current state:
-
-| Input      | Agent state                    | Behavior                                                                                                                                                              |
-| ---------- | ------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| **String** | No prior messages              | **New conversation.** Builds system messages (instructions + skills), adds user message, calls the LLM.                                                               |
-| **String** | Has messages from a prior call | **Continue conversation.** Appends the user message to the existing history and re-enters the LLM loop. Pending tool calls from a prior interrupt are executed first. |
-| **Array**  | No prior messages              | **Restore from persisted data.** Uses the array as-is (no system messages added). Pending tool calls are executed. This is for cross-process resume.                  |
-| **Array**  | Has messages from a prior call | **Raises `Riffer::ArgumentError`.** Use a string to continue, or a new agent instance to start from a persisted array.                                                |
-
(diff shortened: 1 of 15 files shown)


lib/riffer/session.rb
```diff
@@ -0,0 +1,239 @@
...
+  # entries from +tool_calls+, every +Riffer::Messages::Tool+ result whose
+  # +tool_call_id+ matches a dropped call is removed atomically — keeping the
+  # +tool_use+ ↔ +tool_result+ invariant intact.
+  #
+  # Raises Riffer::ArgumentError when neither or both lookup keys are
+  # provided, or when no message matches.
+  #
+  #--
+  #: (?id: String?, ?tool_call_id: String?, **untyped) -> Riffer::Messages::Base
+  def update(id: nil, tool_call_id: nil, **attrs)
```
cloetzi, line 129: thought(non-blocking): I am wondering if we should split up the update methods into `update_assistant(id:)` vs `update_user(id:)` vs `update_tool_call(tool_call_id:)`. The method does a lot and it would make calling update more precise, as in, whatever updates, should know the type. 

Genuinly on the fence if it is better or worse.

docs/04_AGENT_LIFECYCLE.md
```diff
@@ -176,61 +180,59 @@ end
...
 
-Both methods return a `Riffer::Messages::System` or `nil` (when unconfigured). They accept an optional `context:` keyword, just like `generate`.
+Both methods return a `Riffer::Messages::System` or `nil` (when unconfigured). They accept an optional `context:` keyword that overrides the agent's init context for a one-off evaluation.
 
 ```ruby
-agent = MyAgent.new
-sys = agent.generate_instruction_message(context: ctx)     # => Riffer::Messages::System or nil
-skills = agent.generate_skills_message(context: ctx)        # => Riffer::Messages::System or nil
+agent = MyAgent.new(context: ctx)
+sys = agent.generate_instruction_message              # => Riffer::Messages::System or nil
```
cloetzi: thought(non-blocking): Does this approach still make sense? At this point, the agent class instance already generated and stored the system message in the session. If we follow that, this should be `get_system_message` and `get_skill_message` or something and not regenerate it. In practice it is the same, but kinda weird that we are prompting the instance for generation.

lib/riffer/agent.rb
```diff
@@ -982,13 +794,6 @@ def parse_model_string!(model_string)
...
-  #--
-  #: () -> void
-  def clear_resolved_model
-    @resolved_model = nil
-    @provider_instance = nil if @model_config.is_a?(Proc)
-  end
-
   #--
   #: (?Hash[Symbol, untyped]?) -> String?
   def generate_instructions(context = @context)
```
cloetzi: issue(non-blocking): Not technically an active bug, but we should no longer accept context here. This is now called on an instance that has context set, so there is no reason for overriding. As mentioned in the other comment, I'd also rename it.

The same applies to `generate_skills_message` (not in diff)

cloetzi (approved):
Nothing strictly blocking, but I'd suggest removing the context from the two `generate_*` methods.

I might have missed some minor things, but we might as well merge and release a patch if needed. Many more refactorings to come 🙈

docs/04_AGENT_LIFECYCLE.md
bottrall: Good call, revised 👍🏻

lib/riffer/session.rb
bottrall: I see the argument, but I'm leaning towards keeping the generic `update`. If you know the id, you should know what type of message it is.

lib/riffer/agent.rb
bottrall: Good call: https://github.com/janeapp/riffer/pull/265/commits/15a5f127f7b361de7fda413685ac20c2e2548e32

PR #266, opened 2026-05-21
refactor: extract Riffer::Session::Healer pure module (CL2-777)

bottrall:
## Problem

`Riffer::Agent` has grown into a 1000+ LOC class and the history-healing concern is tangled inside it — three private methods (`heal_orphan_tool_calls`, `apply_seed_healing`, `heal_seeded_history`), a placeholder constant, and three call sites each gated by an `if Riffer.config.experimental_history_healing` check that Agent shouldn't have to know about. The healing logic is pure (takes messages, returns messages) but is hidden behind Agent's instance state, which makes it awkward to test in isolation and noisy to read alongside the orchestration logic that's Agent's real job.

This is Phase 2 of the Agent breakup. Phase 1 (#265) extracted `Riffer::Session` as the conversation handle; this phase extracts the healing.

## Solution

Lifts the healing logic into a new stateless `Riffer::Session::Healer` module at `lib/riffer/session/healer.rb` with two entry points — `Healer.heal_orphans(messages) -> [messages, healed_ids]` and `Healer.heal_seeded_session(messages) -> messages` — both pure transforms. The `HEALING_PLACEHOLDER` constant moves with them.

The `experimental_history_healing` flag check moves inside Healer: when the flag is off, each entry point returns its input unchanged (same array reference, no allocation). Agent no longer reads the flag at any call site. The three call sites (init, generate-loop interrupt, stream-loop interrupt) now call Healer unconditionally.

One subtlety worth noting: in the flag-off path, Agent now calls `@session.set(messages)` with the same array it just passed in — a benign identity reassignment, since `Session#set` is silent (no callbacks). No observable behavior change. Net -96 LOC out of Agent (101 deletions, 5 insertions).

The new `test/riffer/session/healer_test.rb` covers the pure logic directly. Two transformation-logic cases in `test/riffer/agent_test.rb` (orphan-strip and parentless-strip at init time) became redundant with the unit tests and were dropped; the interrupt, max-steps, off-default contract, and resume-boundary integration tests are all kept since they exercise Agent's wiring that the unit tests can't reach.

## Proof

CI is sufficient. `bundle exec rake` is green locally (1453 tests, 0 failures, Steep clean, StandardRB clean). The new Healer unit tests cover flag on/off, placeholder placement, `error_type: :interrupted`, call_id ordering, resume-boundary preservation, and input-not-mutated invariants. The retained integration tests in `test/riffer/agent_test.rb` exercise the full wiring through `Agent#generate`, `Response#healed_tool_call_ids`, and the streaming `Interrupt` event — all unchanged and passing.

No user-visible behavior change, no config surface change, no migration concerns.

## Summary by CodeRabbit

* **Refactor**
  * Improved message history management with centralized healing logic for consistent handling of interrupted tool interactions and malformed message structures.

* **Bug Fixes**
  * Agent initialization now automatically removes malformed messages. Interrupt handling more reliably restores tool call states and conversation context when resuming conversations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/janeapp/riffer/pull/266?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/lib/riffer/agent.rb
+++ b/lib/riffer/agent.rb
@@ -26,13 +26,6 @@ class Riffer::Agent
   DEFAULT_MAX_STEPS = 16 #: Integer
   INTERRUPT_MAX_STEPS = :max_steps #: Symbol
 
-  # Placeholder used to fill orphan +tool_use+ blocks when
-  # +Riffer.config.experimental_history_healing+ is enabled and an
-  # interrupt fires mid-tool-use.
-  HEALING_PLACEHOLDER = ->(_tool_call) {
-    Riffer::Tools::Response.error("Tool call interrupted before completion.", type: :interrupted)
-  } #: ^(Riffer::Messages::Assistant::ToolCall) -> Riffer::Tools::Response
-
   # Gets or sets the agent identifier.
   #
   #--
@@ -377,7 +370,7 @@ class Riffer::Agent
     @skills_message = build_skills_message
 
     @session = session || Riffer::Session.new(messages: [@instruction_message, @skills_message].compact)
-    apply_seed_healing if session && Riffer.config.experimental_history_healing
+    @session.set(Riffer::Session::Healer.heal_seeded_session(@session.messages)) if session
   end
 
   # Generates a response from the agent.
@@ -455,46 +448,6 @@ class Riffer::Agent
 
   private
 
-  # Fills any orphaned +tool_use+ in history with the +HEALING_PLACEHOLDER+
-  # response. Each placeholder Tool message is inserted immediately after
-  # its parent assistant message. Returns the array of call_ids that were
-  # filled, in order; +[]+ when there are no orphans.
-  #
-  # Bypasses +on_message+ — placeholders are not inference output.
-  # Consumers learn that healing happened via the structured
-  # +Riffer::Agent::Response#healed_tool_call_ids+ field (and the streaming
-  # +Interrupt+ event's matching field).
-  #
-  #--
-  #: () -> Array[String]
-  def heal_orphan_tool_calls
-    result_ids = @session.messages.filter_map { |m| m.tool_call_id if m.is_a?(Riffer::Messages::Tool) }
-    healed = [] #: Array[String]
-    new_messages = [] #: Array[Riffer::Messages::Base]
-
-    @session.messages.each do |m|
-      new_messages << m
-      next unless m.is_a?(Riffer::Messages::Assistant) && !m.tool_calls.empty?
-
-      m.tool_calls.each do |tc|
-        next if result_ids.include?(tc.call_id)
-
-        response = HEALING_PLACEHOLDER.call(tc)
-        new_messages << Riffer::Messages::Tool.new(
-          response.content,
-          tool_call_id: tc.call_id,
-          name: tc.name,
-          error: response.error_message,
-          error_type: response.error_type
-        )
-        healed << tc.call_id
-      end
-    end
-
-    @session.set(new_messages)
-    healed
-  end
-
   #--
   #: (?Array[Riffer::Guardrails::Modification]) -> Riffer::Agent::Response
   def run_generate_loop(all_modifications = [])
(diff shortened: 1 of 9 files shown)


lib/riffer/agent.rb
```diff
@@ -377,7 +370,7 @@ def initialize(session: nil, context: nil)
     @skills_message = build_skills_message
 
     @session = session || Riffer::Session.new(messages: [@instruction_message, @skills_message].compact)
-    apply_seed_healing if session && Riffer.config.experimental_history_healing
+    @session.set(Riffer::Session::Healer.heal_seeded_session(@session.messages)) if session
```
bottrall: One thing I'm on the fence about is if the agent class should be aware of `Riffer::Session::Healer` at all, or if this should be handled entirely in `Riffer::Session` 🤔

cloetzi: Initially I thought, yea, it should move to session, but on second thought, I think keeping it in the agent is cleaner. The healing happens on agent initialization which makes conceptually sense. If you move to the session, then it becomes this weird question if it is the responsibility of the initializer of the session to call healing, or if it magically does it - which could not be desired under some circumstances.

lib/riffer/session/healer.rb
```diff
@@ -0,0 +1,111 @@
...
+# function returns its input unchanged.
+#
+# Two seams:
+#
+# - +heal_orphans+ — fills orphan +tool_use+ blocks with placeholder
+#   results. Used on interrupt (caller-issued or +max_steps+).
+# - +heal_seeded_session+ — repairs a caller-provided seed so it is
+#   well-formed before the next inference call. Used at construction
+#   time when +Riffer::Agent.new(session:)+ receives a session.
+module Riffer::Session::Healer
```
cloetzi: suggestion(non-blocking): I am starting to have regrets using the word `heal` in the methods initially, would you be open to rename that to `Riffer::Session::Repair`?

lib/riffer/session/healer.rb
```diff
@@ -0,0 +1,111 @@
...
+  # Placeholder used to fill orphan +tool_use+ blocks. Emitted as the
+  # +Riffer::Tools::Response+ body for each healed call_id.
+  HEALING_PLACEHOLDER = ->(_tool_call) {
```
cloetzi: suggestion(non-blocking): `ORPHAN_PLACEHOLDER`

lib/riffer/session/healer.rb
```diff
@@ -0,0 +1,111 @@
...
+  # inserted immediately after its parent assistant message. Returns
+  # +[new_messages, healed_call_ids]+; +healed_call_ids+ is empty when
+  # there are no orphans.
+  #
+  # No-op when +Riffer.config.experimental_history_healing+ is off:
+  # returns +[messages, []]+ with the same array reference.
+  #
+  #--
+  #: (Array[Riffer::Messages::Base]) -> [Array[Riffer::Messages::Base], Array[String]]
+  def self.heal_orphans(messages)
```
cloetzi: suggestion(non-blocking): `Riffer::Session::Repair.fill_orphans`

lib/riffer/session/healer.rb
```diff
@@ -0,0 +1,111 @@
...
+  # tail is purely Tool results (or empty) — are preserved. They get
+  # swept up by +execute_pending_tool_calls+ at the start of the next
+  # generate/stream call.
+  #
+  # No-op when +Riffer.config.experimental_history_healing+ is off:
+  # returns +messages+ unchanged.
+  #
+  #--
+  #: (Array[Riffer::Messages::Base]) -> Array[Riffer::Messages::Base]
+  def self.heal_seeded_session(messages)
```
cloetzi: suggestion(non-blocking): `Riffer::Session::Repair.prune_orphans`

cloetzi (approved):
Abstraction looks good, but I don't like the naming. Consider it nit, but see if it resonates 😄

lib/riffer/session/healer.rb
bottrall: Done — renamed to `Riffer::Session::Repair` in 1eeaf92.

lib/riffer/session/healer.rb
bottrall: Done — `ORPHAN_PLACEHOLDER` in 1eeaf92.

lib/riffer/session/healer.rb
bottrall: Done — `Repair.fill_orphans` in 1eeaf92.

lib/riffer/session/healer.rb
bottrall: Done — `Repair.prune_orphans` in 1eeaf92.

PR #267, opened 2026-05-21
refactor: extract Riffer::Agent::Config (CL2-778)

bottrall:
## Problem

`Riffer::Agent` has accumulated ~12 class-level ivars driven by its DSL (`model`, `instructions`, `provider_options`, `model_options`, `structured_output`, `max_steps`, `uses_tools`, `use_mcp`, `tool_runtime`, `skills`, `guardrail`, `identifier`). Instance methods reach into class-level state via `self.class.foo` throughout `#generate` / `#stream`, which tightly couples instances to their class and forces tests and observability layers to lean on `instance_variable_set` and `send(:provider_instance)` to swap or inspect configuration.

Validations and coercions were also scattered: `model` and `instructions` validated in the Agent DSL, `guardrail` validated phase + class on Agent, `structured_output` validated `is_a?(Riffer::Params)` on Agent, `identifier` coerced via `.to_s` on Agent. Nothing owned the invariants in one place.

This is Phase 2.5 of the Agent breakup — Phase 1 (#265) extracted `Riffer::Session`, Phase 2 (#266) extracted `Session::Healer`, and the upcoming Phase 3 will introduce `Run` to snapshot per-call resolved state.

## Solution

Introduces `Riffer::Agent::Config` — a single mutable configuration object holding every class-level DSL setting. Each `Riffer::Agent` subclass owns one Config via `MyAgent.config`. DSL methods read it and mutate it in place; instance methods read from an instance `@config` ivar that defaults to `self.class.config` but can be replaced via a new data-construction path: `Agent.new(config: explicit_config)`. When an explicit Config is supplied, the class-level configuration is ignored for that instance.

Validation invariants now live on Config:

- `identifier=` coerces non-nil values to String.
- `model=` / `instructions=` reject non-String, non-Proc, and empty-string values.
- `structured_output=` rejects non-`Riffer::Params`, non-nil values.
- `tool_runtime=` rejects values that aren't a ToolRuntime subclass, instance, or Proc.
- `add_guardrail` validates phase (`:before` / `:after` / `:around`) and that the class descends from `Riffer::Guardrail`.

`tool_runtime` is materially simpler. The `tool_runtime` field on Config defaults to `Riffer.config.tool_runtime` at construction time, so the per-instance resolution chain collapses to a Proc resolve + post-resolve Class→Instance step. The cross-subclass inheritance walk is dropped — `tool_runtime` now starts fresh on each subclass like every other DSL setting (it had been the lone exception). Marked experimental in docs, so this is a deliberate API simplification rather than a breaking change of stable behavior. Four inheritance tests removed.

`Riffer::Helpers::Validations` is deleted — its sole consumer was the model / instructions setters, now inlined as a private `validate_string_or_proc!` method on Config. The helper module, its rbs sig, and its dedicated test file all go.

`Riffer::Agent::DEFAULT_MAX_STEPS` moves to `Riffer::Agent::Config::DEFAULT_MAX_STEPS` (kept the constant name; updated three callers). `INTERRUPT_MAX_STEPS` stays on Agent — it's the interrupt-reason symbol, not configuration.

Test coverage was rebalanced: `test/riffer/agent/config_test.rb` (new) owns every Config-side invariant; `test/riffer/agent_test.rb` keeps the Agent-level smoke tests (DSL→Config delegation), integration paths (`#generate` / `#stream` / class-name `identifier` default / block-DSL parsing for `structured_output` / `**options` spreading for `guardrail`), and `#initialize` behavior. Net change to the test suite: 12 duplicate tests removed from `agent_test.rb`, equivalent coverage added in `config_test.rb`, plus 2 regression tests for the `tool_runtime:` constructor flag and the global-default snapshot. New `Agent.new(config:)` block covers the data path.

## Proof

CI is sufficient. Locally `bundle exec rake` is green — 1468 tests, 0 failures, Steep clean, StandardRB clean. `bundle exec rake rbs:generate` produces `sig/generated/riffer/agent/config.rbs` with full type info from rbs-inline (no hand-written sig needed).

No user-visible behavior change except the documented `tool_runtime` subclass-inheritance drop. DSL surface, error messages, and `Agent.new` / `#generate` / `#stream` semantics are all preserved.

## Summary by CodeRabbit

* **New Features**
  * Agents can accept a custom configuration at instantiation to override class-level settings; per-class configs are isolated.

* **Documentation**
  * Guides clarified config behavior and defaulting for tool runtime and class-level vs instance-level overrides.

* **Tests**
  * Added comprehensive tests covering the new configuration behavior.

* **Chores**
  * Removed legacy validations helper and its tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/janeapp/riffer/pull/267?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


docs/03_AGENTS.md
```diff
@@ -263,7 +263,7 @@ class MyAgent < Riffer::Agent
 end
 ```
 
-Accepts a `Riffer::ToolRuntime` subclass, a `Riffer::ToolRuntime` instance, or a `Proc`. Inherited by subclasses. When unset, falls back to `Riffer.config.tool_runtime`. See [Tools — Tool Runtime](07_TOOL_ADVANCED.md#tool-runtime-experimental) for details.
```
bottrall, line 266: I decided to remove the logic to ensure subclass inheritance because it added complexity + diverges from the behavior of every other DSL. Which do not get inherited by subclasses.

lib/riffer/helpers/validations.rb
bottrall, line 1: Only used within the config ivars. Therefore felt it didn't deserve its own helper module.

lib/riffer/agent.rb
```diff
@@ -101,35 +91,30 @@ def self.model_options(options = nil)
...
+    config.structured_output = params if params
+    config.structured_output
   end
 
   # Gets or sets the maximum number of LLM call steps in the tool-use loop.
   #
-  # Defaults to DEFAULT_MAX_STEPS (16). Set to +Float::INFINITY+ for
-  # unlimited steps.
+  # Defaults to Riffer::Agent::Config::DEFAULT_MAX_STEPS (16). Set to
+  # +Float::INFINITY+ for unlimited steps.
```
cloetzi, line 104: thought(non-blocking): This doesn't translate to protobuf 😅 I had to impelemnt `-1 == infinity` as a workaround. Given that we are breaking stuff left and right, would it make sense to transition that to a more agnostic value? I am not quite sure what it would be tough except `-1`.

lib/riffer/agent/config.rb
```diff
@@ -0,0 +1,173 @@
...
+  # [phase] +:before+, +:after+, or +:around+. +:around+ appends to both
+  #   +:before+ and +:after+.
+  # [klass] the Riffer::Guardrail subclass to register.
+  # [options] options forwarded to the guardrail at runtime.
+  #
+  # Raises Riffer::ArgumentError on an invalid phase or non-Guardrail class.
+  #
+  #--
+  #: (Symbol, klass: singleton(Riffer::Guardrail), ?options: Hash[Symbol, untyped]) -> void
+  def add_guardrail(phase, klass:, options: {})
```
cloetzi, line 140: thought(non-blocking): I do wonder if we should extract this out of the config class. In practice not much would change, but it might deserve its own class definition. The motivation behind is that for clean transfer for remote execution, ideally config is cleanly serializable. Currently, guardrails and the tool runtime are not "wire-compatible". We can either handle it when we add the serialization method, or we could treat them separate. 

What's the plan for serializing and rebuilding methods?

cloetzi (approved):
Nothing preventing from merging this, but some open question around how we are going to serialize the config class in the future.

lib/riffer/agent.rb
bottrall: Agreed we should transition to something more agnostic. However lets do that as a dedicated followup instead of tacking onto this refactor.

lib/riffer/agent/config.rb
bottrall: Agreed, guardrails and `tool_runtime` aren't wire-compatible. Two things I'd add:

The wire-unfriendly boundary is actually a bit wider than just those two. `tools_config` carries Tool class refs (or a Proc), `skills_config` references backend / adapter / activate_tool classes, and any of `model` / `instructions` / `tools_config` / `tool_runtime` can hold a Proc. So if we split, I think we'd want to capture all of those code-reference fields on one side and keep `Config` as pure data (`identifier`, `model`/`instructions` as String, `provider_options`, `model_options`, `max_steps`, `mcp_configs`).

No concrete plan documented yet for serialization. Phase 3 of the current refactor introduces `Riffer::Run` to snapshot per-call resolved values. That gets us closer to wire-friendly for the Proc-resolving fields (Run holds Strings, not lambdas), but it doesn't address the static class-ref problem on guardrails / tool_runtime / tools_config / skills_config.

For now lets defer until we have a concrete serialization design. Both because the right split is easier to see once that design exists, and because Phase 3 will reshape what's read from Config vs Run at resolution time, which will inform where the boundary should sit.

Sound good?

PR #268, opened 2026-05-22
refactor!: extract Riffer::Agent::Run, eagerly resolve per-agent state

bottrall:
## Problem

`Riffer::Agent` conflated three concerns: class-level DSL, the conversation handle (session/messages), and the per-call generation engine (LLM loop, tool dispatch, guardrails, healing). The loop methods accumulated ad-hoc per-call ivars (`@structured_output`, `@provider_instance`, `@resolved_tools`, `@resolved_tool_runtime`) and reached for global `Riffer.config.X` mid-loop. Downstream code worked around the missing surface with `instance_variable_set` hacks. Phase 3 of the refactor plan was to extract a `Riffer::Run` for per-run state and shrink Agent to a DSL + facade.

This PR delivers that — and goes further, settling into a final shape after several iterations: skills/provider/tool resolution are owned by Agent (resolved eagerly at `Agent.new`), Run is a stateless module of functions operating on an agent, and `agent.context` is the single mutable Hash that carries skills state and cumulative token usage across runs.

## Solution

**Resolution boundary.** Agent eagerly resolves everything it can hand to the loop on construction:

- model string → `agent.model_name` (Proc form called against `context` once)
- provider instance → `agent.provider` (built from `provider_class.new(**provider_options)`)
- skills context → `agent.context[:skills]` (or `nil`)
- structured output schema → `agent.structured_output`
- resolved tool catalog (incl. MCP + skill_activate) → `agent.tools`
- tool runtime → `agent.tool_runtime`

Errors that used to surface on the first `#generate` (unknown provider, invalid model string, missing tool description, tool name conflict, AWS region missing) now raise at `Agent.new`. `agent.context` is always a mutable `Hash` and is `dup`d from the caller's input so Riffer never mutates a hash it didn't create.

**`Riffer::Agent::Run`.** A module (`extend self`, no instances) with two public functions: `Riffer::Agent::Run.generate(agent:, prompt:, files:)` and `Riffer::Agent::Run.stream(agent:, prompt:, files:)`. Every helper takes `agent` as a positional first arg; nothing is held in ivars. The loop reads what it needs through `agent.session`, `agent.provider`, `agent.tools`, `agent.tool_runtime`, etc. Cumulative token usage is mutated into `agent.context[:token_usage]` as the loop progresses.

**Agent facade.** `Agent#generate` and `#stream` are two-liners that delegate to the module. The streaming guard ("structured_output not supported with streaming") fires on `Agent#stream` before the module is called.

**Test reshape.** `agent_test.rb` covers DSL, constructor, and the eagerly-resolved attrs (`#tools`, `#tool_runtime`, validation, MCP merge, dynamic-instructions Procs, structured_output, find/all). `run_test.rb` covers loop behaviour driven through `agent.generate` / `agent.stream` (tool calling, guardrails, streaming events, interrupt + healing, resume, multi-turn, token tracking, structured-output parsing).

## Proof

CI is sufficient — full suite runs `bundle exec rake` (Minitest + Standard + Steep). 1,479 tests / 2,276 assertions pass clean. RBS signatures regenerated under `sig/generated/`.

For reviewers: please verify the **BREAKING CHANGES in the commit body** match what your downstream consumers need to migrate. The biggest hotspots are:

- `Riffer::Run` → `Riffer::Agent::Run` (and it's a module now, not a class)
- `agent.provider_instance` → `agent.provider`
- `agent.resolved_tools` → `agent.tools`
- `agent.token_usage` → `agent.context[:token_usage]`
- `agent.skills_state` → `agent.context[:skills]`
- `Agent.resolved_tool_classes` removed; resolution happens at `Agent.new` via `agent.tools`
- Validation errors (unknown provider, missing tool description, tool name conflict) now raise at `Agent.new` rather than `#generate`

Maestro should be the main downstream check. If anything there does `instance_variable_set(:@structured_output, ...)`, that workaround is now obsolete — set it via `MyAgent.structured_output(params)` instead.

## Summary by CodeRabbit

* **Documentation**
  * Clarified agent.context as the mutable runtime context, showing token_usage, skills, and access examples.

* **Refactor**
  * Agent orchestration reorganized for clearer runtime initialization and a dedicated execution/run flow.

* **Enhancements**
  * Mock provider responses normalized for consistent tool-call and token_usage handling.
  * Session additions: silent message append, assistant-step count, and final-assistant lookup.
  * Message API: detection helper for pending tool calls.

* **Tests**
  * Expanded end-to-end and unit coverage across runs, streaming, tools, structured output, tokens, interrupts, files, and resume semantics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/janeapp/riffer/pull/268?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


bottrall:
@Thibaut don't be alarmed. It's not as big as it looks at first glance 😅 the vast majority of the churn is moving tests from `agent_test.rb` to `run_test.rb`.

docs/04_AGENT_LIFECYCLE.md
```diff
@@ -283,23 +283,22 @@ Mutating history while a `stream` enumerator is being consumed is undefined; mut
 
 Mutators do **not** fire `on_message` — that callback is reserved for messages produced by inference (LLM responses, tool execution results). Healing placeholders bypass `on_message` for the same reason; consumers learn that healing happened via `Response#healed_tool_call_ids` (and `StreamEvents::Interrupt#healed_tool_call_ids`).
 
-### token_usage
+### context
 
-Access cumulative token usage across all LLM calls:
+The mutable runtime context. A `Hash` threaded into every Proc-based DSL setting, guardrail, tool runtime, and skills resolution, and shared with every `Riffer::Agent::Run` this agent executes. Carries:
```
Thibaut, line 288: thought (non-blocking): at some point it may be beneficial to introduce `Riffer::Context` over a `Hash`. It would enable us to build a contract around some of the built-ins like `skills` and `token_usage` and also enforce whatever we want to enforce on context.

lib/riffer/agent/run.rb
```diff
@@ -0,0 +1,320 @@
...
+      model: agent.model_name,
+      tools: agent.tools,
+      **merged_model_options(agent)
+    )
+  end
+
+  #--
+  #: (Riffer::Messages::Assistant) -> bool
+  def has_tool_calls?(response)
+    response.is_a?(Riffer::Messages::Assistant) && !response.tool_calls.empty?
```
Thibaut: suggestion (non-blocking): move into `Riffer::Messages::Base` so we can do `response.has_tool_calls?` without type-checking (`Riffer::Messages::Assistant` can override default as needed)

lib/riffer/agent/run.rb
```diff
@@ -0,0 +1,320 @@
...
+    agent.session.set(new_messages)
+    response = final_assistant_message(agent)
+
+    build_response(agent, response&.content || "", modifications: all_modifications, interrupted: true, interrupt_reason: reason, structured_output: validate_structured_output(agent, response), healed_tool_call_ids: filled)
+  end
+
+  #--
+  #: (Riffer::Agent, Enumerator::Yielder) -> void
+  def run_stream_loop(agent, yielder)
+    step = agent.session.count { |m| m.is_a?(Riffer::Messages::Assistant) }
```
Thibaut: suggestion (non-blocking): moving into `Riffer::Session` so we can do:


```suggestion
    step = agent.session.steps
```

lib/riffer/agent/run.rb
```diff
@@ -0,0 +1,320 @@
...
+    opts[:structured_output] = agent.structured_output if agent.structured_output
+    opts
+  end
+
+  #--
+  #: (Riffer::Agent) -> Riffer::Messages::Assistant?
+  def final_assistant_message(agent)
+    # TODO: Replace with rfind when minimum Ruby is 4.0+
+    # rubocop:disable Style/ReverseFind
+    agent.session.reverse_each.find { |msg| msg.is_a?(Riffer::Messages::Assistant) } #: Riffer::Messages::Assistant?
```
Thibaut: suggestion (non-blocking): moving into `Riffer::Session` so we can do:


```suggestion
    agent.session.final_assistant_message
```

lib/riffer/agent/run.rb
```diff
@@ -0,0 +1,320 @@
...
+  def build_response(agent, content, tripwire: nil, modifications: [], interrupted: false, interrupt_reason: nil, structured_output: nil, healed_tool_call_ids: [])
+    messages = agent.session.messages
+    Riffer::Agent::Response.new(content, tripwire: tripwire, modifications: modifications, interrupted: interrupted, interrupt_reason: interrupt_reason, structured_output: structured_output, messages: messages.frozen? ? messages : messages.dup.freeze, healed_tool_call_ids: healed_tool_call_ids)
+  end
+
+  #--
+  #: (Riffer::Agent, String, files: Array[Hash[Symbol, untyped] | Riffer::FilePart]?) -> void
+  def append_user_message(agent, prompt, files:)
+    file_parts = (files || []).map { |f| convert_to_file_part(f) }
+    agent.session.set(agent.session.messages + [Riffer::Messages::User.new(prompt, files: file_parts)])
```
Thibaut: suggestion (non-blocking): replacing the whole session outside of the session kinda breaks the abstraction of `Riffer::Session`. Suggest refactoring so we can do something like:


```suggestion
    agent.session.append_message(Riffer::Messages::User.new(prompt, files: file_parts))
```

Thibaut (approved):
Took the opportunity to re-review some of the code that was moved. Feel free to address in a follow-up!

docs/04_AGENT_LIFECYCLE.md
bottrall: Agreed I had the same thought. Will do in a follow up.

PR #275, opened 2026-05-25
refactor: unify generate and stream loops (CL2-780)

bottrall:
## Problem

`Riffer::Agent::Run` carried two near-duplicate orchestration methods, `run_generate_loop` and `run_stream_loop`, that shared roughly 70% of their structure (per-step LLM call → token tracking → guardrail post-processing → tripwire/interrupt handling → tool execution) but were already drifting. The clearest symptom was `skills.on_activate`, which was wired only in the stream loop — sync agents silently dropped activation notifications even when a caller would have wanted them. The two-loop shape was the largest source of "fix it in one place, forget the other" risk in the gem.

This is Phase 4 of the riffer refactor; the per-call state extraction (Phase 3) had to land first.

## Solution

Collapsed both loops into a single `run_loop(agent, stream_yielder)`. Sync calls pass `nil`; stream calls pass a real `Enumerator::Yielder`. Every step of the loop is shared — the only divergences are the LLM call shape (atomic via `call_llm` vs. accumulated from `call_llm_stream`) and whether per-step events are emitted, both gated by `if stream_yielder`. `skills.on_activate` now lives next to the other emit sites, structurally guarded by the same condition, so the drift-prone footgun can't recur.

Along the way the two guardrail phases got pulled into shape: `run_before_guardrails` and `run_after_guardrails` each own their own modifications recording (via `record_modifications!`) and use a block-yield pattern to surface tripwire short-circuits — the caller's block does `return tripped`, which exits `run_loop` cleanly without nullable tuple destructure that steep would refuse to narrow. Three small helpers (`record_modifications!`, `tripwire_response`, `final_response`) absorbed the two-call-site patterns that remained.

A note on the bang in `record_modifications!`: the method mutates its `all_modifications` argument by `concat`, which technically rubs against the "favor immutability" rule in the global CLAUDE.md. We discussed the alternatives (return-and-reassign, move into `agent.context`, leave it) and landed on keeping the local accumulator with a Ruby-idiomatic bang to make the mutation explicit at the call site. The mutation is local to `run_loop`'s private helper chain — no caller boundary is crossed.

Public API is unchanged. `Riffer::Agent#generate`, `Riffer::Agent#stream`, `Riffer::Agent::Response`, every `Riffer::StreamEvents::*` type, every provider adapter, and `Riffer::Skills::Context` are all untouched.

## Proof

CI is sufficient. The existing 1486-test suite covers both modes (sync `generate`, stream `stream`), tripwires, modifications, interrupts (including the `interrupt!(nil)` case that the first refactor draft tripped on), max_steps healing, structured output, and stream-side skill activation events — all green without any test changes. Lint (`bundle exec rake standard`) and type check (`bundle exec steep check`) are both clean. The diff to `lib/riffer/agent/run.rb` is 180 insertions / 130 deletions, taking the file from 305 to ~290 lines.

## Summary by CodeRabbit

* **Chores**
  * Internal refactor to consolidate generation and streaming logic for more consistent behavior.
* **Bug Fixes**
  * Improved handling of interrupted runs to prevent orphaned tool calls and ensure a final structured response.
  * More reliable streaming with accurate accumulated assistant output and token accounting.
* **Behavior**
  * User messages with no prompt are now ignored; files still require a prompt.

[Review Change Stack](https://app.coderabbit.ai/change-stack/janeapp/riffer/pull/275?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/lib/riffer/agent/run.rb
+++ b/lib/riffer/agent/run.rb
@@ -23,17 +23,8 @@ module Riffer::Agent::Run
   #--
   #: (agent: Riffer::Agent, ?prompt: String?, ?files: Array[Hash[Symbol, untyped] | Riffer::FilePart]?) -> Riffer::Agent::Response
   def generate(agent:, prompt: nil, files: nil)
-    raise Riffer::ArgumentError, "files: requires a prompt" if files && !files.empty? && prompt.nil?
-
-    append_user_message(agent, prompt, files: files) if prompt
-
-    all_modifications = [] #: Array[Riffer::Guardrails::Modification]
-
-    tripwire, modifications = run_before_guardrails(agent)
-    all_modifications.concat(modifications)
-    return build_response(agent, "", tripwire: tripwire, modifications: all_modifications) if tripwire
-
-    run_generate_loop(agent, all_modifications)
+    append_user_message(agent, prompt, files: files)
+    run_loop(agent, nil)
   end
 
   # Runs the streaming loop for the given agent. See Riffer::Agent#stream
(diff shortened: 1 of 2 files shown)


lib/riffer/agent/run.rb
```diff
@@ -23,17 +23,8 @@ module Riffer::Agent::Run
...
-
-    all_modifications = [] #: Array[Riffer::Guardrails::Modification]
-
-    tripwire, modifications = run_before_guardrails(agent)
-    all_modifications.concat(modifications)
-    return build_response(agent, "", tripwire: tripwire, modifications: all_modifications) if tripwire
-
-    run_generate_loop(agent, all_modifications)
+    append_user_message(agent, prompt, files: files)
+    run_loop(agent, nil)
```
Thibaut: suggestion (non-blocking): use named arguments, so we don't have a mystery `nil` trailing here.

lib/riffer/agent/run.rb
```diff
@@ -89,96 +80,83 @@ def run_generate_loop(agent, all_modifications = [])
...
 
-        throw :riffer_interrupt, Riffer::Agent::INTERRUPT_MAX_STEPS if step >= agent.config.max_steps
+  # Appends +new_modifications+ to +all_modifications+ and emits a
+  # +GuardrailModification+ event for each one when streaming.
+  #
+  #--
+  #: (Enumerator::Yielder?, Array[Riffer::Guardrails::Modification], Array[Riffer::Guardrails::Modification]) -> void
+  def record_modifications!(stream_yielder, all_modifications, new_modifications)
+    all_modifications.concat(new_modifications)
+    new_modifications.each { |m| stream_yielder << Riffer::StreamEvents::GuardrailModification.new(m) } if stream_yielder
```
Thibaut, line 137: thought (non-blocking): seeing `agent`, `stream_yielder` and others whizzed around makes me wonder if we are missing an abstraction to represent this complicated "run" process, that could store state (an `@agent` and `@stream_yielder` would clean up much of this code). Should `Agent::Run` be a class instead of a module? 🤔

Thibaut (approved):
Awesome! I was on the verge of commenting this in the previous PR 😅

lib/riffer/agent/run.rb
bottrall: Good call. https://github.com/janeapp/riffer/pull/275/commits/0a14222e7e3ac2e5abc942ffe6595ce70cf11c7c

lib/riffer/agent/run.rb
bottrall: Tempting, but I'd keep it a module. I've been trying to be strict about only using classes when there's real identity, lifecycle, or mutable state that earns its keep. Modules for orchestration. A `Run` instance would live for one `generate`/`stream` call and mostly hold `@agent` (already on Agent) plus two per-call accumulators. Not enough state to justify it IMO.

PR #280, opened 2026-05-28
feat: add OpenRouter provider

tsov:
## Summary

- Add OpenRouter as a new LLM provider (`Riffer::Providers::OpenRouter`).
- Subclasses `Riffer::Providers::Base` directly. OpenRouter is Chat Completions only, not Responses API, so it doesn't inherit from `OpenAI`.
- Reuses the `openai` gem with a `base_url` override. No new dependency.
- Full hook coverage: `generate_text`, `stream_text`, tools, structured output, file inputs, reasoning, token usage.
- Pass-through for OpenRouter-specific options: `provider` routing, `models` fallback, `transforms`.
- Model string format: `openrouter/<openrouter-model-id>` (e.g. `openrouter/anthropic/claude-sonnet-4.6`).

## Files

- `lib/riffer/providers/open_router.rb` — Provider implementation
- `lib/riffer/config.rb` — Config struct
- `lib/riffer/providers/repository.rb` — Registry entry
- `test/riffer/providers/open_router_test.rb` — 46 tests, 11 VCR cassettes
- `docs/providers/09_OPENROUTER.md` — Provider documentation

## Summary by CodeRabbit

* **New Features**
  * Added OpenRouter support: chat completions, streaming, tool calling, structured JSON output, image/file handling, and token-usage reporting.

* **Documentation**
  * Comprehensive OpenRouter provider docs covering setup, model naming, options, streaming, tool usage, routing/fallbacks, and examples.

* **Tests**
  * Extensive test coverage and recorded fixtures for sync/streaming flows, tool calls, structured output, file handling, and usage reporting.

* **Chores**
  * Added provider configuration and registry entries for OpenRouter.

[Review Change Stack](https://app.coderabbit.ai/change-stack/janeapp/riffer/pull/280?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


lib/riffer/providers/open_router.rb
```diff
@@ -0,0 +1,314 @@
...
+      }
+    end
+
+    params.compact
+  end
+
+  #--
+  #: (Hash[Symbol, untyped]) -> OpenAI::Models::Chat::ChatCompletion
+  def execute_generate(params)
+    @client.chat.completions.create(**params)
```
cloetzi, line 69: question: I noticed that OpenAI uses ` @client.responses.create(params)`, what is the difference and why are we diverging for open router? It applies to the other client usages as well.

test/riffer/providers/open_router_test.rb
```diff
@@ -0,0 +1,628 @@
...
+    end
+  end
+
+  describe "#build_request_params" do
+    let(:provider) { Riffer::Providers::OpenRouter.new(api_key: api_key) }
+    let(:user_message) { Riffer::Messages::User.new("Hello") }
+
+    it "includes model and messages" do
+      params = provider.send(:build_request_params, [user_message], "anthropic/claude-3.5-sonnet", {})
+      expect(params[:model]).must_equal "anthropic/claude-3.5-sonnet"
```
bottrall: **Nitpick(non-blocking):** Use a more recent model.



Claude 3.5 sonnet is considered deprecated by Anthropic. So it's unclear how long it will continue to be accessible. I would recommend adjusting everything to haiku 4.5.

bottrall (approved):
Thanks @tsov.

Only other thing that might be worth a discussion is whether we should use the [responses api](https://openrouter.ai/docs/api/reference/responses/overview). It's considered beta, but would allow us to largely reuse the `Riffer::Providers::OpenAI` class.

lib/riffer/providers/open_router.rb
tsov: This is actually intentional. `responses.create` uses OpenAI's proprietary Response API, while `chat.completions.create` uses the older Chat Completions API, which OpenRouter implements as a drop-in OpenAI-compatible endpoint, so it's the canonical path for OpenRouter. It also derives the structured-output shape. But to @bottrall point, we could also use the Responses API. Let me confirm if we can use it

test/riffer/providers/open_router_test.rb
tsov: I thought I caught them all, will change it now

lib/riffer/providers/open_router.rb
tsov: Yeah their Responses API is in beta: https://openrouter.ai/docs/api/reference/responses/overview

I would actually recommend we wait until it stabilizes. Additionally, their schema also diverges from OpenAI, so we would not work as a "drop-in". `inputs` vs `input`, different `reasoning`/`format` shapes, meaning the openai gem's typed `responses.create` payload wouldn't match what OpenRouter expects without us bypassing the SDK.

Chat Completions is their canonical, stable path and works as a true drop-in with the openai gem via `base_url` override, which is why I went with it for v1. Happy to revisit once the Responses API is GA.

tsov:
@bottrall switched to haiku 4.5 across the board and re-recorded the VCR cassettes

PR #293, opened 2026-06-02
feat!: Riffer::Agent::Serializer for transferable agent definitions

bottrall:
## Problem

Riffer agents are defined in code via the class-level DSL, mixing pure **data** (`identifier`, `model`, `instructions`, options, `max_steps`, structured output) with **code** (`Proc`s, tool classes, tool runtime) that can't cross a process/service boundary or be persisted as plain data. We want to enable the ability to serialize an agent *definition* and reconstruct a **runnable** agent elsewhere — two targets from one format: persisted-and-rehydrated within a monolith (where tool classes still exist), and shipped to a distributed receiver that holds only the Riffer gem.

## Solution

Adds `Riffer::Agent::Serializer` — a pure module with `to_h`/`from_h` (and `to_json`/`from_json` convenience that handle `symbolize_names` for you), reached via `agent.to_h` / `Riffer::Agent.from_h` / `.from_json` delegators. `from_h`/`from_json` take an optional `context:` — the rebuilt agent's runtime context (threaded into tool dispatch, not used to re-resolve the already-resolved definition).

- **Resolved snapshot.** `to_h` reads the agent's *resolved* instance state, so `Proc`-based `model`/`instructions` are already evaluated against the origin's context — the dict carries plain data, never Procs.
- **Pluggable tool rehydration.** Tools cross as `{name, description, parameters_schema, timeout}` descriptors. A `tool_resolver:` lambda turns each into a `Riffer::Tool` class: the default synthesizes body-less **shells** (pair with an injected remote `Tools::Runtime` that forwards calls by overriding `#dispatch_tool_call`); pass a registry lookup to rebuild **real in-process** classes. MCP tools and `skill_activate` ride this path as ordinary tools.
- **Structured output as JSON Schema.** Round-trips as the provider-neutral `to_json_schema(strict: false)` via the new public `Riffer::Params.from_json_schema`; the rebuilt agent re-renders per its own provider at call time. To make it lossless, non-strict `Param#to_json_schema` now emits `default:`.
- **`max_steps` encoding.** The DSL keeps `nil` = unlimited; the serializer encodes that as `-1` on the wire (and decodes `-1` back to `nil`) so the dict stays portable to transports where JSON `null` is awkward (e.g. proto3, which can't distinguish `null` from an absent field). The magic value never leaves the wire.
- **Versioning seam.** `schema_version` integer with a version→decoder `case` in `from_h`; unknown versions raise `Serializer::VersionError`. A future incompatible change bumps the integer and adds a backwards-compatible decoder.

**Breaking changes** (noted in the commit footer):
- Unlimited `max_steps` is now `nil` at the agent level (`max_steps nil`), not `Float::INFINITY`. `Config#max_steps` may be `nil`.
- `Riffer::Params` parameter types are typed `Module` (widened from `Class`) to honestly include `Riffer::Params::Boolean`, which is a Module.

**Deferred (documented):** guardrails and the skills subsystem are not serialized yet; a skills-enabled agent's `skill_activate` tool still crosses as an ordinary tool as a stopgap. Secrets in `provider_options`/`model_options` **do** ride the wire — flagged with a warning to handle a serialized definition as a secret if its options carry sensitive values.

## Proof

CI is sufficient. `bin/rake` (test + standardrb + steep) is green: new coverage in `test/riffer/agent/serializer_test.rb` (round-trip, shell vs registry-lookup resolution, injected runtime, optional context, `max_steps` `nil`↔`-1` encoding, version handling, JSON helpers) and `test/riffer/params_test.rb` (`from_json_schema` + lossless `to_json_schema` round-trip, including nested objects, typed arrays, defaults, enums, and Boolean). Existing `run_test.rb` covers the `max_steps nil` loop behavior.

## Summary by CodeRabbit

* **New Features**
  * Agent snapshot/restore as JSON/hashes; reconstruct agents from serialized payloads
  * Parameters can be rebuilt from JSON Schema for structured validation

* **Bug Fixes**
  * Unlimited loop configuration now uses nil as the explicit "unlimited" sentinel

* **Documentation**
  * Added comprehensive Serialization guide and updated agent step-limit docs and overview

* **Tests**
  * Added end-to-end serializer and params test coverage


docs/03_AGENTS.md
```diff
@@ -152,7 +152,7 @@ end
 
 ### max_steps
 
-Sets the maximum number of LLM call steps in the tool-use loop. When the limit is reached, the loop interrupts with reason `:max_steps`. Defaults to `16`. Set to `Float::INFINITY` for unlimited steps:
+Sets the maximum number of LLM call steps in the tool-use loop. When the limit is reached, the loop interrupts with reason `:max_steps`. Defaults to `16`. Set to `nil` (`max_steps nil`) for unlimited steps:
```
bottrall, line 155: I was on the fence about going with `nil = infinity` or having a more explicit `"infinity" = infinity`. Decided to go with `nil`, as I think that makes sense that `no max steps = infinity`. Happy to revisit if anyone feels strongly about it.

lib/riffer/agent.rb
```diff
@@ -308,10 +343,10 @@ def initialize(session: nil, context: nil, config: nil)
     @config = config || self.class.config
     @context = Riffer::Agent::Context.new(context || {})
 
-    provider_class, @model_name = resolve_provider_and_model
-    @provider = provider_class.new(**@config.provider_options)
+    @provider_name, @model_name = resolve_provider_and_model
+    @provider = build_provider
 
-    @context.skills = resolve_skills(provider_class)
+    @context.skills = resolve_skills
```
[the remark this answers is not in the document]
bottrall: This was an explicit change and decision in a previous PR. No behavior has changed in this PR.

docs/15_SERIALIZATION.md
```diff
@@ -0,0 +1,94 @@
...
+```ruby
+{
+  schema_version:    1,                       # wire format version
+  riffer_version:    "0.29.1",                # diagnostic only
+  identifier:        "support_agent",
+  model:             "openai/gpt-4o",         # resolved "provider/model" string
+  instructions:      "You are…",              # resolved system prompt
+  model_options:     { temperature: 0.2 },
+  provider_options:  { … },                   # see the secrets warning below
+  max_steps:         8,                        # integer, or null = unlimited
```
cloetzi: issue(non-blocking): This is still not ideal with gRPC. In gRPC you can't determine if a value is `null` or missing. Both are identical. It can be solved by a translation layer like with the infinite approach earlier, so not a blocker, but not ideal.

docs/03_AGENTS.md
Thibaut: Agreed!

lib/riffer/agent/serializer.rb
```diff
@@ -0,0 +1,201 @@
...
+  #
+  #--
+  #: (agent: Riffer::Agent) -> Hash[Symbol, untyped]
+  def to_h(agent:)
+    config = agent.config
+    {
+      schema_version: SCHEMA_VERSION,
+      riffer_version: Riffer::VERSION,
+      identifier: config.identifier,
+      model: "#{agent.provider_name}/#{agent.model_name}",
```
Thibaut, line 64: suggestion (non-blocking): add `config.model` to abstract here and since it's supported in the initializer

docs/15_SERIALIZATION.md
```diff
@@ -0,0 +1,94 @@
...
+```ruby
+dict    = agent.to_h                              # snapshot
+rebuilt = Riffer::Agent.from_h(dict, context: {}) # reconstruct
+```
+
+The dict is plain data — symbol-keyed, JSON-safe. For the wire, use the JSON helpers, which handle generating and parsing for you:
+
+```ruby
+json    = agent.to_json                       # or Riffer::Agent::Serializer.to_json(agent:)
+rebuilt = Riffer::Agent.from_json(json, context: {})
```
cloetzi: question(non-blocking): Could context be an optional argument with a default of `{}`? All examples use an empty context. 

Further, what are the scenarios where you'd actually want to set a custom context on restore? I think the doc should be more explicit about that, I struggle understanding why we need it 😅

docs/15_SERIALIZATION.md
```diff
@@ -0,0 +1,94 @@
...
+
+You own what a resolved tool does: a resolver may return real in-process classes, shells, or classes that themselves make network calls. Riffer does not require a runtime — it only ships shells by default.
+
+## `max_steps`
+
+Unlimited steps are represented as `nil` at the agent level — set it with `max_steps nil`. `nil` serializes as JSON `null` and round-trips unchanged; a finite integer round-trips as-is. A dict missing the key falls back to the default (16) rather than running unbounded.
+
+## Versioning
+
+`schema_version` is an integer (`Riffer::Agent::Serializer::SCHEMA_VERSION`). `from_h` refuses any version it doesn't recognize with `Riffer::Agent::Serializer::VersionError` (a `Riffer::ArgumentError`). A future incompatible change bumps the integer and adds a back-compat decoder, giving distributed consumers a window to upgrade before the old format is dropped.
```
cloetzi: nitpick(non-blocking): Spell out `backwards compatible`.

docs/15_SERIALIZATION.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Secrets
+
+`provider_options` and `model_options` **ride on the wire as plain data** — they are part of the dict and _will_ transfer. Care needs to be taken when configuring API keys via `provider_options` and using serialization. Prefer environment/global provider configuration. **Never serialize an agent whose options carry sensitive values** — treat a serialized definition as non-sensitive.
```
cloetzi: nitpick(non-blocking): `treat a serialized definition as non-sensitive.` it might be my tired brain, but it doesn't quite make sense to me. I'd just delete that final sentence, or something like `**Never serialize an agent whose options carry sensitive values**, or treat them as secrets.`. Not sure.

lib/riffer/agent/serializer.rb
```diff
@@ -0,0 +1,201 @@
...
+    tool_name = descriptor[:name]
+    tool_description = descriptor[:description]
+    schema = descriptor[:parameters_schema]
+    tool_timeout = descriptor[:timeout]
+
+    # An anonymous Riffer::Tool subclass is the idiom for synthesizing a tool
+    # from data — the tool DSL is class-level, so there is no value-level
+    # builder to type against. Same approach as Riffer::Mcp::ToolFactory;
+    # steep can't type the dynamic class body, hence the ignore block.
+    Class.new(Riffer::Tool) do
```
Thibaut, line 200: thought (non-blocking): I don't know where Ruby is at nowadays but back in the day dynamic class definition at runtime (code paths that run not just on load but during requests) was an anti-pattern as it triggered some kind of cache bust (method cache?) that was expensive at scale. Caching or not, using static code eval (e.g. `module_eval` with a string) generally has better performance characteristics. Looks like it would be doable here since we can just serialize variables like `schema` directly into Ruby code.

Thibaut (approved):
Awesome!

lib/riffer/agent/serializer.rb
cloetzi, line 200: 🙌

lib/riffer/agent/serializer.rb
bottrall: `agent.config.model` [already exists](https://github.com/janeapp/riffer/blob/main/lib/riffer/agent/config.rb#L19-L20), but it's the unresolved `string` or `Proc`. So to correctly resolve it in this class we would need to provide the `context` and do:
```rb
model: Riffer::Helpers::CallOrValue.resolve(config.model, context: context),
```
I did consider this option, but felt that storing the resolved provider name on the agent and then reconstructing it here was cleaner.

lib/riffer/agent/serializer.rb
bottrall: Interesting call-out — I had CC benchmark it on 3.3 and 4.0 out of curiosity. The historical method-cache-bust penalty doesn't seem to reproduce, and string eval actually comes out slower here on both axes:

| Ruby | shell creation                                   | `parameters_schema` read |
| ---- | ------------------------------------------------ | ------------------------ |
| 3.3  | `Class.new`+block ~533k/s vs `class_eval` ~43k/s | ~14× slower              |
| 4.0  | block ~174k/s vs eval ~51k/s                     | ~12× slower              |


Claude's explanation:

> Two reasons: `class_eval "<string>"` pays a parse/compile cost per call, and the eval'd `parameters_schema` rebuilds the hash literal from source on every call (allocations), whereas the block closes over the one shared object. It also matches what `Mcp::ToolFactory` already does.
> 
> `from_h` runs once per agent reconstruction (per session for Maestro), not per inference step, so it's not hot. If it ever shows up in a profile, the lever would be memoizing shells by descriptor rather than string eval. Leaving as-is for now — thanks for the nudge to verify.

docs/15_SERIALIZATION.md
bottrall: https://github.com/janeapp/riffer/pull/293/commits/698d974028e1aa65949a23f56e345539b2626b19

docs/15_SERIALIZATION.md
bottrall: https://github.com/janeapp/riffer/pull/293/commits/698d974028e1aa65949a23f56e345539b2626b19

A scenario where we do want to provide the context is when an agent is hydrated within the same project. For example we persist an agent definition to the database & query it, then re-hydrate it in the same project. Context may be needed for that agent to use it's associated tools/skills.

docs/15_SERIALIZATION.md
bottrall: https://github.com/janeapp/riffer/pull/293/commits/698d974028e1aa65949a23f56e345539b2626b19

I think this is clearer. But let me know if it still needs work.

docs/15_SERIALIZATION.md
bottrall: I had a good think about this. How do we feel about this.

**In the Agent DSL**
- integer = bounded
- nil = unlimited
- omitted = default max steps defined by config

**In the Agent Serializer**
- integer = bounded
- -1 = unlimited
- omitted = default max steps defined by config

I think this is an ok compromise as it works for any persistence framework / transfer protocol. But is abstracted away from the public interface in favor of more semantic values. We don't have to worry about consumers dealing with a magic number as it's handled during encode and decode.

Thoughts?

cc @Thibaut

Thibaut: I hate it but I agree it's the last bad option 😅 (unless we are open to making 42 the infinity)

IIRC Maestro also already uses `-1`.

bottrall: https://github.com/janeapp/riffer/pull/293/commits/cec2f24fa2fee390fa1e0d379c6a49fe68d668dd

cloetzi (approved):

PR #295, opened 2026-06-03
feat: forward session: through Agent.from_h/from_json for history seeding

cloetzi:
## Why

`Riffer::Agent::Serializer.from_h` built `Agent.new(config:, context:)` with **no `session:`**, so a rehydrated agent seeded only the instruction message — there was no seam to restore conversation history. `Agent.new` already uses a provided session verbatim; the serializer just never offered the forward. Consumers that persist history separately (e.g. Maestro) need to seed it on rebuild.

## What

- **`lib/riffer/agent/serializer.rb`** — add a backward-compatible `session:` keyword to `from_h` / `from_json`, threaded through `decode_v1` into `Agent.new(..., session: session)`.
- **`lib/riffer/agent.rb`** — add `session:` to the `Agent.from_h` / `Agent.from_json` delegators.
- **`sig/generated/…`** — regenerated via `bin/rbs`.
- **`test/riffer/agent/serializer_test.rb`** — new `session` describe: omitted → instruction-seeded (unchanged); provided → used verbatim (same object, history preserved); forwarded through `from_json`; and the deliberate "instructions are *not* re-injected into a provided session" contract.
- **`docs/15_SERIALIZATION.md`** — new "Seeding conversation history" section.

## Semantics

When `session:` is omitted, behavior is **identical** to before (fresh session seeded from the dict's `instructions`). When supplied, the session is used **as-is** — the caller owns its full contents, **including the system instruction message** — so the dict's `instructions` are not re-injected. The dict still carries only the agent *definition*, not history; this is purely the rebuild-side seam for callers that persist history separately.

## Release

Additive + backward-compatible. As a `feat:`, release-please should cut **0.31.0**, which the downstream pin (Maestro Slice 2) targets.

## Checks

`bin/rake` (test + standard + steep:check): **1604 runs, 2469 assertions, 0 failures, 0 errors**; StandardRB clean; Steep reports no type errors.

## Summary by CodeRabbit

* **New Features**
  * Deserializing an agent can now accept an existing session to resume persisted conversations; a provided session is used as-is (history and system instruction preserved).

* **Behavior**
  * Omitting a session seeds a fresh session from the agent's instructions; serialized agent data does not contain prior conversation history.

* **Documentation**
  * Expanded docs explaining session seeding/resumption semantics and portability/versioning notes.

* **Tests**
  * Added tests for session seeding, JSON session forwarding, and prevention of system-instruction reinjection.

### DIFF
--- a/docs/15_SERIALIZATION.md
+++ b/docs/15_SERIALIZATION.md
@@ -22,6 +22,17 @@ The hash forms (`to_h` / `from_h`) are public too, if you want to embed the dict
 
 `from_h` / `from_json` accept an optional `context:` — the rebuilt agent's **runtime** context, exactly the value you'd pass to `Agent.new(context:)`. It is **not** used to re-resolve the serialized definition (that's already resolved); it's threaded into tool dispatch and read by tools/runtimes at call time. Pass it when a tool or a remote runtime needs per-call data — e.g. `context: { tenant: "acme" }` for multi-tenant dispatch, or Maestro passing `context: { agent: self }` so its runtime can call back. Omit it (defaults to empty) when nothing downstream reads context.
 
+### Seeding conversation history
+
+The dict carries the agent **definition**, not its conversation history (see [What does not transfer](#what-does-not-transfer)). To resume a persisted conversation, pass a `session:` — exactly the value you'd pass to `Agent.new(session:)`:
+
+```ruby
+rebuilt = Riffer::Agent.from_h(dict, session: persisted_session)
+rebuilt.generate # continues from the seeded history
+```
+
+The session is used **as-is**: the rebuilt agent does not prepend anything to it, so the caller owns its full contents — **including the system instruction message**. When you omit `session:`, behavior is unchanged: a fresh session is built and seeded with the dict's `instructions` (and an empty history). Because a supplied session is authoritative, the dict's `instructions` are not re-injected into it — make sure your persisted session already contains the system message.
+
 ## What the dict carries
 
 ```ruby
--- a/docs/15_SERIALIZATION.md
+++ b/docs/15_SERIALIZATION.md
@@ -1,22 +1,22 @@
 # Serialization
 
-`Riffer::Agent::Serializer` turns a **resolved agent** into a self-contained, provider-neutral data dict (`to_h`) and reconstructs a **runnable agent** from that dict (`from_h`). Use it to persist agent definitions outside of code, or to transfer them across a process/service boundary.
+`Riffer::Agent::Serializer` turns a **resolved agent** into a self-contained, provider-neutral data hash (`to_h`) and reconstructs a **runnable agent** from that hash (`from_h`). Use it to persist agent definitions outside of code, or to transfer them across a process/service boundary.
 
 You normally reach it through the delegators on `Riffer::Agent`:
 
 ```ruby
-dict    = agent.to_h          # snapshot
-rebuilt = Riffer::Agent.from_h(dict) # reconstruct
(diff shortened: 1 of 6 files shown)


docs/15_SERIALIZATION.md
```diff
@@ -22,6 +22,17 @@ The hash forms (`to_h` / `from_h`) are public too, if you want to embed the dict
...
+### Seeding conversation history
+
+The dict carries the agent **definition**, not its conversation history (see [What does not transfer](#what-does-not-transfer)). To resume a persisted conversation, pass a `session:` — exactly the value you'd pass to `Agent.new(session:)`:
+
+```ruby
+rebuilt = Riffer::Agent.from_h(dict, session: persisted_session)
+rebuilt.generate # continues from the seeded history
+```
+
+The session is used **as-is**: the rebuilt agent does not prepend anything to it, so the caller owns its full contents — **including the system instruction message**. When you omit `session:`, behavior is unchanged: a fresh session is built and seeded with the dict's `instructions` (and an empty history). Because a supplied session is authoritative, the dict's `instructions` are not re-injected into it — make sure your persisted session already contains the system message.
```
Thibaut: > When you omit `session:`, behavior is unchanged

nitpick: this turn of phrase sounds like it is documenting against a prior state (before this PR) that isn't relevant to the reader of those docs, who only knows the current state.

lib/riffer/agent/serializer.rb
```diff
@@ -79,6 +79,12 @@ def to_h(agent:)
   #   config (the dict is already resolved); it is threaded into tool dispatch
   #   and read by tools/runtimes at call time (e.g. a remote runtime keying off
   #   <tt>context[:tenant]</tt>). Defaults to an empty context.
+  # [session] an optional Riffer::Agent::Session to seed conversation history,
+  #   forwarded verbatim to +Agent.new(session:)+. The dict carries the agent
+  #   *definition*, not its history (see "What does not transfer"); pass a
+  #   session here to resume a persisted conversation. Used as-is — the caller
+  #   owns its contents, including the system instruction message. When omitted,
+  #   a fresh session seeded with the dict's instructions is built, as before.
```
Thibaut: nitpick: same comment as above regarding the ", as before"

Thibaut (approved):

docs/15_SERIALIZATION.md
```diff
@@ -22,6 +22,17 @@ The hash forms (`to_h` / `from_h`) are public too, if you want to embed the dict
 
 `from_h` / `from_json` accept an optional `context:` — the rebuilt agent's **runtime** context, exactly the value you'd pass to `Agent.new(context:)`. It is **not** used to re-resolve the serialized definition (that's already resolved); it's threaded into tool dispatch and read by tools/runtimes at call time. Pass it when a tool or a remote runtime needs per-call data — e.g. `context: { tenant: "acme" }` for multi-tenant dispatch, or Maestro passing `context: { agent: self }` so its runtime can call back. Omit it (defaults to empty) when nothing downstream reads context.
 
+### Seeding conversation history
+
+The dict carries the agent **definition**, not its conversation history (see [What does not transfer](#what-does-not-transfer)). To resume a persisted conversation, pass a `session:` — exactly the value you'd pass to `Agent.new(session:)`:
```
bottrall: **Nitpick(non-blocking):** It's only just occurred to me that the previous PR introduced the term `dict` a tonne. Which isn't a ruby term - it's a python term.



If it's not too much of a lift let's to do a quick find and replace for all `dict ` > `hash `

bottrall (approved):

PR #297, opened 2026-06-04
chore: reduce comment noise and consolidate comment/RDoc policy

bottrall:
## Problem

Comments across the codebase had drifted into noise. RDoc duplicated what rbs-inline already encodes (param/return labeled-lists, type restatement), descriptions restated method names, raise docs re-listed the declared type union, several comments narrated past changes, genuine "why"s ran long enough to belong in `docs/`, and value objects carried ceremonial examples that taught nothing. The written guidance was also split across `rdoc.md` and `code-style.md`, so it drifted.

## Solution

Rewrote the comment/RDoc guidance and consolidated it into `.agents/code-style.md` (removed `.agents/rdoc.md`, kept `rbs-inline.md` as the standalone type reference, updated `AGENTS.md`), then swept every in-scope file to that policy. The policy, in short:

- Types are rbs-inline's job — no parameter/return prose.
- Public surface gets a brief, verb-first description; a second sentence only for a non-obvious "why". Brevity is its own constraint — extended rationale goes to `docs/`.
- Raises: contract-only, phrased by intent rather than re-listing the `#:` type.
- Private-method comments survive only for a why a reader can't recover from the code.
- Examples only where a consumer uses the thing themselves, and only when they teach something the signature can't.
- No history — comments describe the present.
- Exemptions: self-evident constants, empty namespace modules, pure-restatement constructors.

Genuine "why"s were deliberately preserved (e.g. the provider SSE-socket closes, SDK-bug workarounds, the `-1` max_steps wire sentinel, the `tool_use`↔`tool_result` invariant). Net ~1,200 comment lines removed. The final commit regenerates `sig/generated` because rbs-inline copies prose comments into the generated RBS — comment-only, no signature changes.

## Proof

CI is sufficient — this is a comments-only change with no behavior or signature changes. Verified locally:

- `bin/rake` green: 1600 tests, 0 failures; StandardRB clean; Steep reports no type errors.
- `bin/docs` builds RDoc with no errors.
- The `sig/generated` diff was confirmed comment-only (no `#:`/declaration changes).

## Summary by CodeRabbit

* **Documentation**
  * Consolidated code style, RDoc, and rbs-inline guidelines into unified documentation standards.
  * Refined inline documentation and RBS type annotations across the library for improved clarity.
  * Streamlined API documentation while maintaining existing method signatures and behavior.


tsov (approved):
💯

Thibaut (approved):
✨

PR #298, opened 2026-06-04
refactor!: apply module/class conventions consistently

bottrall:
## Problem

The codebase mixed several shapes for the same job: classes whose methods were all `self.` (modules wearing pants), modules using `def self.`/`class << self` where `extend self` is the idiom, pure-function helper modules `include`d/`extend`ed into hosts purely for call-site convenience, and a `Converter` module whose factories really belonged on the types they build. None of it was wrong functionally, but it drifted from "my" code-organization guidelines (class only for identity/lifecycle/state; `module ... extend self` for pure functions; behavior on the type it belongs to).

## Solution

A focused sweep applying those rules:

- **Modules-wearing-pants → `module ... extend self`:** `Evals::EvaluatorRunner` and `Providers::Repository` (no instance state), plus the `def self.`/`class << self` modules `Riffer` (root), `Mcp`, `Agent::Session::Repair`, `Mcp::AuthenticatedTool`, `Mcp::ToolFactory`, and `Mcp::Registry`. `Registry` stays a deliberate stateful singleton — its `@store`/`@mutex` and mutex-guarded access are unchanged; only the declaration form moved from `class << self` to `extend self`.
- **De-mixined the pure-function helpers:** `Helpers::ClassNameConverter` and `Helpers::Dependencies` are now `extend self` modules called explicitly, dropping the convenience `include`/`extend` from `Agent`, `Providers::Base`, `Runner::Fibers`, and `Mcp::Client`. (`Tools::Toolable` stays a mixin — it's a genuine stateful class-level DSL.)
- **Relocated `Messages::Converter` onto its models:** the two factories now live as `Messages::Base.from_hash` (polymorphic) and `Messages::FilePart.from_hash`, and the converter module + mixin are deleted.
- **RBS:** added hand-written `sig/manual` re-extend stubs for each new `extend self` module (rbs-inline can't emit singleton methods for `extend self`) and regenerated `sig/generated`.

Notable trade-off: value objects were **left as immutable `attr_reader` classes** rather than converted to `Data.define`. rbs-inline (this repo's typing backbone) drops `Data.define` block methods and emits an all-required constructor, which would have forced `rbs_inline: disabled` plus a hand-written signature for all 13 of them — a convention break and maintenance burden out of proportion to the benefit.

## Proof

CI is sufficient — this is a behavior-preserving refactor fully covered by the existing suite, and `bin/rake` (test + standard + steep:check) is green locally: 1603 tests, 0 failures; Steep clean; StandardRB clean. The relocated `Converter` logic keeps its full test coverage, moved into `messages/base_test.rb` and `messages/file_part_test.rb` as `.from_hash` cases. Public surface change: `docs/08_MESSAGES.md` was updated to point hash-normalization at `Messages::Base.from_hash`.

## Release Notes

* **New Features**
  * Added `from_hash` class methods to message classes for improved hash-to-message conversion and consistency.

* **Documentation**
  * Updated guidance for restoring persisted conversation history to recommend new `from_hash` methods instead of previous converter utilities.


lib/riffer/mcp/client.rb
```diff
@@ -56,4 +54,11 @@ def tools_call(name, arguments = {})
     content = response.dig("result", "content") || []
     content.filter_map { |item| item["text"] }.join
   end
+
+  private
+
+  #: (String) -> true
+  def depends_on(gem_name)
+    Riffer::Helpers::Dependencies.depends_on(gem_name)
```
Thibaut, line 62: question (non-blocking): since this is just a simple proxy method, why not call `Riffer::Helpers::Dependencies.depends_on` directly inline, like we do with the other methods in this PR like `Riffer::Messages::Base.from_hash`?

Thibaut (approved):
✨ 

I wonder if we should update some agent files so the 🤖 keeps the established pattern going forward?

lib/riffer/mcp/client.rb
bottrall: I decided to add a helper because it's called twice. mcp + faraday. Makes it a bit nicer to read through.

bottrall:
> ✨ 
> 
> I wonder if we should update some agent files so the 🤖 keeps the established pattern going forward?

I've noticed for these types of patterns agents usually copy whatever is established in the repo. Monkey see, monkey do.

But if I notice divergence we can add to the agents files.

PR #303, opened 2026-06-10
fix: respect disable-model-invocation in skills

bottrall:
## Problem

Riffer documents itself as implementing the [Agent Skills spec](https://agentskills.io/), but the `disable-model-invocation` frontmatter flag was silently swallowed into the `metadata` passthrough and ignored. A skill whose author explicitly opted out of model-driven activation was still rendered in the model-facing catalog and could be activated via the `skill_activate` tool.

The spec is explicit: such skills must be **hidden from the catalog** and kept **unreachable by the model**, remaining reachable only through explicit/non-model activation. In riffer's framework context, that explicit path is the programmatic `activate` / `activate_tool` config — letting a developer inject a skill under their own logic (tenant, route, feature flag) instead of the model's judgment.

## Solution

Promote the flag to a first-class, typed field and enforce it through the per-agent skills layer, with a clean split of responsibilities:

- **`Frontmatter`** gains a `disable_model_invocation` boolean — consumed out of `metadata` (single source of truth), strict `== true` coercion so any non-boolean or absent value leaves the skill model-invocable.
- **`Context` owns the _definition_ of invocability**: `available_skills` now rejects disabled skills (catalog hidden), plus `model_invocable?(name)` and `activatable?` predicates. `Context#activate` is left untouched — a permissive primitive, so the programmatic `activate` config still reaches disabled skills (the escape hatch).
- **`ActivateTool` enforces the gate at the model's entry point**: it returns the same `"Unknown skill"` error (no existence leak) when a skill isn't `model_invocable?`. The tool is the model's only activation surface, so the gate belongs there rather than in the shared primitive.
- **`Agent`** registers `skill_activate` only when at least one skill is `activatable?`, per the spec's "don't register a tool with no valid options"; `build_skills_message` no longer emits an empty system message when nothing is renderable.

**Deliberately out of scope:** an enum-constrained `name` param on `skill_activate`. Hiding + the activation guard already make disabled skills unreachable and spec-compliant; the enum is an orthogonal anti-hallucination improvement (it would apply to all skills) and carries real complexity, so it's left for a separate change.

While scoping this, I found that subclassing `ActivateTool` does not inherit its class-level config (`identifier`/`description`/`params`/`timeout`), breaking the documented "subclass, override only `call`" pattern. It's independent of this change and filed as **CL2-898**.

## Proof

CI is sufficient — `bin/rake` passes (1683 runs, 0 failures; Steep clean; StandardRB clean). New tests cover:

- Frontmatter parsing/coercion of `disable-model-invocation` and that it does not leak into `metadata` (`frontmatter_test.rb`).
- `Context#model_invocable?`, `#activatable?`, catalog hiding, and permissive programmatic `#activate` (new `context_test.rb`).
- `ActivateTool` rejecting a disabled skill with `"Unknown skill"` (`activate_tool_test.rb`).
- Agent-level catalog hiding, `skill_activate` suppression when all skills are disabled, and programmatic activation of a disabled skill (`agent_integration_test.rb`).

## Release Notes

* **New Features**
  * Skills can now be marked with `disable-model-invocation: true` in their YAML frontmatter to prevent the model from invoking them via the skill activation tool, while still allowing programmatic activation.

* **Documentation**
  * Updated documentation to specify optional frontmatter fields and clarify how the `disable-model-invocation` flag controls model access to skills.

### DIFF
--- a/docs/13_SKILLS.md
+++ b/docs/13_SKILLS.md
@@ -36,7 +36,11 @@ Review the code for:
 - `name` — lowercase alphanumeric with hyphens, 1-64 chars (must match directory name)
 - `description` — 1-1024 chars, helps the LLM decide when to activate
 
-Additional frontmatter keys are passed through as metadata.
+**Optional frontmatter fields:**
+
+- `disable-model-invocation` — when `true`, the skill is hidden from the catalog and the LLM cannot activate it via the `skill_activate` tool. It stays reachable through the programmatic `activate` config (see [Activated Skills](#activated-skills)), so you can inject it under your own logic instead of the model's. Only the literal value `true` disables invocation; any other value (or its absence) leaves the skill model-invocable.
+
+Any other frontmatter keys are passed through as metadata.
 
 ## Configuring an Agent
 
@@ -81,7 +85,7 @@ end
 
 ### Activated Skills
 
-Load skill instructions into the system prompt at startup (no tool call needed):
+Load skill instructions into the system prompt at startup (no tool call needed). This is also the only way to surface a skill marked `disable-model-invocation: true`, which the model can never activate on its own:
 
 ```ruby
 skills do
--- a/lib/riffer/agent.rb
+++ b/lib/riffer/agent.rb
@@ -343,9 +343,9 @@ class Riffer::Agent
   #--
   #: () -> Riffer::Messages::System?
   def build_skills_message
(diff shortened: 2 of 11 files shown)


janetxinli (approved):

tsov (approved):
✨

PR #305, opened 2026-06-11
feat: support user-explicit skill activation and dedupe re-activations

bottrall:
## Problem

Riffer implements the [Agent Skills spec](https://agentskills.io/client-implementation/adding-skills-support)'s model-driven activation channel (catalog in system prompt, `skill_activate` tool) but has no supported path for the spec's second channel: **user-explicit activation**, where the harness intercepts a slash command and injects the skill body itself. The only documented escape hatch is the config-time `activate` option, which puts runtime-requested bodies in the system prompt — wrong tier per the spec's progressive-disclosure model, and it busts the provider prompt cache on every activation.

Two adjacent defects compound this. Re-activating a skill returns the full body again every time, duplicating instructions in context (the failure mode tracked as anthropics/claude-code#21891 in Claude Code, whose spec section "Deduplicate activations" recommends skipping re-injection). And the run loop permanently overwrites `Context#on_activate` with a lambda holding the stream's `Enumerator::Yielder`, so any harness-side activation after a completed stream would push into a dead yielder.

## Solution

`Riffer::Skills::Context` gains the primitives for the user channel while staying unopinionated about injection policy — the consumer decides what to send; riffer makes the correct thing a one-liner:

- `activation_prompt(name, args:)` — returns the body wrapped in `<skill_content name="...">` (the spec's structured-wrapping recommendation) ready to pass to `generate`, and records the activation
- `read(name)` — body fetch with no activation recorded, for UI previews and subagent delegation
- `deactivate(name)` — clears activation state, for consumers who prune skill content from history
- `preactivate(name)` — the config-time tier; only these bodies render in `system_prompt`, so runtime activations can never mutate the system prompt or shrink the catalog

Both channels share one activation state, so the shipped `ActivateTool` now answers re-activations with a short overridable pointer ("already active") instead of the body — including when the user channel injected it first. Consumers who prefer the full-body-always behavior of other harnesses subclass the tool (documented in `docs/13_SKILLS.md`). The adapters' catalog instructions tell the model not to re-activate skills whose `<skill_content>` blocks are already in context, and the run loop now composes with and restores the consumer's `on_activate` callback around streaming.

Behavior changes to note when reviewing: the activate tool's success result is now wrapped in `<skill_content>` tags, re-activation returns a pointer rather than the cached body, and pre-activated system-prompt bodies are wrapped too. Activation state is in-memory by design (the transcript is the durable record, consistent with every surveyed implementation), so a rebuilt agent on a persisted session returns one full body before dedupe resumes — documented in the skills guide.

## Proof

CI is sufficient — `bin/rake` (tests + StandardRB + Steep) passes with 38 new/updated cases covering every new Context method, pointer-on-re-activation at unit and end-to-end level, cross-channel dedupe (user injection → model tool gets the pointer), deactivate restoring full-body behavior, the system-prompt/catalog split, and the on_activate compose/restore across a stream (`test/riffer/skills/context_test.rb`, `activate_tool_test.rb`, `agent_integration_test.rb`, adapter tests).

## Summary by CodeRabbit

* **New Features**
  * Re-activation now returns a short "already active" pointer instead of duplicating full skill content.
  * Activated skill content is wrapped and surfaced to the model in a consistent tagged format; catalog prompts include activation guidance.

* **Bug Fixes**
  * Stream handler wiring is restored/cleaned up to avoid lingering callbacks after runs.

* **Documentation**
  * Clarified activation flows, model vs app-triggered activation, preactivation/system-prompt behavior, and examples.

* **Refactor**
  * Separated skill body caching from activation state; improved preactivate/deactivate lifecycle.

### DIFF
--- a/docs/13_SKILLS.md
+++ b/docs/13_SKILLS.md
@@ -1,6 +1,6 @@
 # Skills
 
-Skills are packaged AI agent capabilities per the [Agent Skills spec](https://agentskills.io/). Each skill is a directory containing a `SKILL.md` file with YAML frontmatter and Markdown instructions. The framework discovers skills through a pluggable backend, injects a compact catalog into the system prompt (~50 tokens/skill), and provides a tool for the LLM to activate skills on demand.
+Skills are packaged AI agent capabilities per the [Agent Skills spec](https://agentskills.io/). Each skill is a directory containing a `SKILL.md` file with YAML frontmatter and Markdown instructions. The framework discovers skills through a pluggable backend, injects a compact catalog into the system prompt (~50 tokens/skill), and supports both activation channels the spec describes: the LLM activates skills on demand through a tool, and your application injects skills the user asked for as conversation content (see [User-Triggered Activation](#user-triggered-activation)).
 
 ## Creating a Skill
 
@@ -38,7 +38,7 @@ Review the code for:
 
 **Optional frontmatter fields:**
 
-- `disable-model-invocation` — when `true`, the skill is hidden from the catalog and the LLM cannot activate it via the `skill_activate` tool. It stays reachable through the programmatic `activate` config (see [Activated Skills](#activated-skills)), so you can inject it under your own logic instead of the model's. Only the literal value `true` disables invocation; any other value (or its absence) leaves the skill model-invocable.
+- `disable-model-invocation` — when `true`, the skill is hidden from the catalog and the LLM cannot activate it via the `skill_activate` tool. It stays reachable through the programmatic `activate` config (see [Activated Skills](#activated-skills)) and through `activation_prompt` (see [User-Triggered Activation](#user-triggered-activation)), so you can inject it under your own logic instead of the model's. Only the literal value `true` disables invocation; any other value (or its absence) leaves the skill model-invocable.
 
 Any other frontmatter keys are passed through as metadata.
 
@@ -85,7 +85,7 @@ end
 
 ### Activated Skills
 
-Load skill instructions into the system prompt at startup (no tool call needed). This is also the only way to surface a skill marked `disable-model-invocation: true`, which the model can never activate on its own:
+Load skill instructions into the system prompt at startup (no tool call needed). Use this for skills that should govern the whole session — for skills the user requests mid-conversation, prefer [User-Triggered Activation](#user-triggered-activation), which keeps the system prompt (and its provider-side cache) stable:
 
 ```ruby
 skills do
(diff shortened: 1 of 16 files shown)


docs/13_SKILLS.md
```diff
@@ -107,8 +107,46 @@ end
...
+`activation_prompt("code-review", args: "focus on security")` returns:
+
+```
+<skill_content name="code-review">
+You are a code review assistant.
+...
+</skill_content>
+
+focus on security
+```
```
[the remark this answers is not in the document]
bottrall: This is a non-issue.

lib/riffer/skills/adapter.rb
```diff
@@ -21,4 +21,19 @@ def initialize(skill_activate_tool:)
   def render_catalog(skills)
     raise NotImplementedError, "#{self.class} must implement #render_catalog"
   end
+
+  # Renders an activated skill body wrapped in identifying tags.
+  #--
+  #: (Riffer::Skills::Frontmatter, String) -> String
+  def render_activation(skill, body)
+    %(<skill_content name="#{skill.name}">\n#{body}\n</skill_content>)
```
Thibaut, line 29: suggestion (non-blocking): make this `raise NotImplementedError` and implement in the sub-class. Not every adapter uses XML syntax. Last I checked only Anthropic used XML for their models, and everyone else (including OpenAI) used plain markdown, which is why the markdown adapter is the default. Even if we decide to use XML for the markdown adapter (I would ask how Codex and others do this), I think this is adapter-specific. Lastly, moving to the subclass makes it easier to read the subclass and understand how everything fits together in a single file.

lib/riffer/skills/adapter.rb
```diff
@@ -21,4 +21,19 @@ def initialize(skill_activate_tool:)
...
+  end
+
+  # The behavioral instructions rendered alongside the catalog.
+  #--
+  #: () -> String
+  def catalog_instructions
+    "When a user's request matches a skill description below, call the `#{skill_activate_tool.name}` tool with the skill name. After activation, follow the skill's instructions. " \
+      "If a skill's instructions already appear in your context (inside <skill_content> tags), follow them instead of activating the skill again."
```
Thibaut, line 37: Same suggestion as above given the reference to XML tagging.

lib/riffer/skills/context.rb
```diff
@@ -33,19 +48,58 @@ def initialize(backend:, skills:, adapter:)
...
+  end
+
+  # Activates a skill and returns its body wrapped for injection as a user
+  # message.
+  #
+  # Raises Riffer::ArgumentError if the skill is not in the catalog.
+  #
+  #--
+  #: (String, ?args: String?) -> String
+  def activation_prompt(name, args: nil)
```
Thibaut: suggestion (non-blocking): let the caller do the concatenation as it sees fit. I was very confused at first by the "focus on security" in the doc and what `args` actually is. I don't think it's the responsibility of the skills context to know about the broader prompt.

Thibaut (approved):
nice!

lib/riffer/skills/adapter.rb
bottrall: This one's deliberate: the `<skill_content name="...">` envelope is the wrapping format the Agent Skills integration guide itself recommends ([Structured wrapping](https://agentskills.io/client-implementation/adding-skills-support#structured-wrapping)), independent of how the catalog is rendered. We also surveyed how other harnesses and libraries handle this before landing on it, and tag-wrapped skill content is the common convention across providers, not an Anthropic/XML thing. Codex CLI wraps bodies in `<skill>` tags, opencode uses `<skill_content>` verbatim, pi uses `<skill name= location=>`, Cline `<user_command>`, Crush `<loaded_skill>`. All of them serving OpenAI/markdown-first models. So the wrapper is an identification envelope (it's what the dedupe guidance and the "already active" pointer reference), while `render_catalog` stays the provider-specific axis. It's still overridable in a subclass if an adapter ever needs a different envelope.

Happy to revisit if you feel strongly, but I think this is a sensible default and consumers can override if they desire.

PR #307, opened 2026-06-12
feat: add tracing foundation with optional OTEL backend

bottrall:
## Problem

Riffer has no structured tracing — when an eval score moves or an agent misbehaves in production, there's no way to see which step did it. Native OTEL instrumentation of the agent loop is the fix, but rubygems has no optional-dependency concept, so the first problem to solve is plumbing: riffer must emit OTEL spans when a host wires an SDK and behave identically (and cheaply) when it doesn't, without adding any runtime dependency to the gemspec.

## Solution

This is the foundation only — no spans are wired into the agent loop yet; follow-up PRs add the agent-run, LLM-call, and tool-execution spans on top of it.

`Riffer::Tracing` is an internal port with two duck-typed backends, resolved lazily at first span: `Otel` when a guarded `require "opentelemetry"` succeeds at a supported version (`>= 1.1, < 2` — outside that it warns once and falls back), and `Null`, a frozen no-op, otherwise. The surface is block-only `in_span` (yielding a riffer-owned span wrapper so `::OpenTelemetry` constants never leak past the backend), plus `current_context`/`with_context` for re-attaching trace context across fiber and thread boundaries — the streaming enumerator and the concurrent tool runners both need this, and the context tests pin the behavior the span work will rely on.

Config gains `config.tracing.enabled` (a kill switch consulted on every span, so a runtime flip takes effect immediately) and `config.tracing.tracer_provider` (injectable, primarily for tests; raises at assignment if OTEL isn't loadable). `capture_messages` is deliberately deferred to the LLM-span follow-up that will honor it, so no inert config knob ships.

Notable choices a reviewer might otherwise trip on:

- `opentelemetry-sdk` lives in a **Gemfile group**, not the gemspec dev deps, because gemspec dev deps can't be selectively excluded — and the no-OTEL CI lane needs to exclude it.
- `::OpenTelemetry` constants appear only inside method bodies of the `Otel` backend, so zeitwerk eager-loading can't `NameError` when the gem is absent. A new eager-load test enforces this in the lane where it can actually fail.
- The schema URL pin is a documented constant rather than tracer metadata because the Ruby OTEL API's `tracer(name, version)` accepts no schema URL.

## Proof

CI is sufficient — see `test/riffer/tracing_test.rb` (span export, attributes, kind, scope, nesting, error status, mid-process kill-switch toggle, cross-fiber and cross-thread context parenting against the real in-memory SDK), `test/riffer/tracing/null_test.rb`, `test/riffer/tracing/otel_test.rb` (version-guard predicate), and `test/eager_load_test.rb`. The new `test-no-otel` job runs the full suite with `BUNDLE_WITHOUT=opentelemetry`; five inverse-skip tests execute only there and assert the genuine absent-gem fallback. Both postures verified locally: `bin/rake` (1729 runs) and `BUNDLE_WITHOUT=opentelemetry bin/test` are green.

## Summary by CodeRabbit

* **New Features**
  * Global tracing controls with enable/disable and pluggable OpenTelemetry provider; new boolean coercion helper.

* **Documentation**
  * Configuration guide updated with a new tracing section and usage examples.

* **Tests**
  * Added coverage for tracing behavior, null fallback, OTEL integration, eager loading, and boolean coercion.

* **Chores**
  * CI lane added to run tests with the optional tracing dependency excluded; optional opentelemetry group declared.

### DIFF
--- a/lib/riffer/config.rb
+++ b/lib/riffer/config.rb
@@ -72,13 +72,7 @@ class Riffer::Config
     #--
     #: (untyped) -> void
     def enabled=(value)
-      @enabled = case value
-      when true, "true", 1, "1" then true
-      when false, "false", 0, "0", nil then false
-      else
-        raise Riffer::ArgumentError,
-          "enabled must be a boolean (or 'true'/'false'/'1'/'0'/1/0), got #{value.inspect}"
-      end
+      @enabled = Riffer::Helpers::Boolean.coerce(value, attribute: "enabled")
     end
 
     # Sets an explicit tracer provider, forcing the OTEL backend. Raises
@@ -172,13 +166,7 @@ class Riffer::Config
   #--
   #: (untyped) -> void
   def experimental_history_healing=(value)
-    @experimental_history_healing = case value
-    when true, "true", 1, "1" then true
-    when false, "false", 0, "0", nil then false
-    else
-      raise Riffer::ArgumentError,
-        "experimental_history_healing must be a boolean (or 'true'/'false'/'1'/'0'/1/0), got #{value.inspect}"
-    end
+    @experimental_history_healing = Riffer::Helpers::Boolean.coerce(value, attribute: "experimental_history_healing")
   end
 
   #--
--- a/lib/riffer/config.rb
+++ b/lib/riffer/config.rb
@@ -49,6 +49,53 @@ class Riffer::Config
     end
   end
 
+  # Tracing-related global configuration.
+  class Tracing
+    # Whether riffer emits OTEL spans; defaults to +true+, a no-op until a
+    # host wires an OTEL SDK.
+    attr_reader :enabled #: bool
+
+    # Explicit OTEL tracer provider; defaults to +nil+, which resolves the
+    # global <tt>OpenTelemetry.tracer_provider</tt> at first span.
+    attr_reader :tracer_provider #: untyped
+
+    #--
+    #: () -> void
+    def initialize
+      @enabled = true
+      @tracer_provider = nil
+    end
+
+    # Sets the enabled flag, coercing boolean-ish values so an env-var
+    # +"false"+ (truthy in Ruby) doesn't silently keep tracing on. Raises
+    # Riffer::ArgumentError on an unrecognized value.
+    #--
+    #: (untyped) -> void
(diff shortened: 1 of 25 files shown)


lib/riffer/config.rb
```diff
@@ -49,6 +49,53 @@ def default_backend=(value)
...
+      @tracer_provider = nil
+    end
+
+    # Sets the enabled flag, coercing boolean-ish values so an env-var
+    # +"false"+ (truthy in Ruby) doesn't silently keep tracing on. Raises
+    # Riffer::ArgumentError on an unrecognized value.
+    #--
+    #: (untyped) -> void
+    def enabled=(value)
+      @enabled = case value
```
Thibaut: suggestion (non-blocking): this could be extracted as a utility for all config classes and their boolean setters, to keep things DRY and consistent.

Thibaut (approved):
awesome!

lib/riffer/config.rb
bottrall: Extracted as `Riffer::Helpers::Boolean.coerce(value, attribute:)` in 1afd480 — both boolean setters (`tracing.enabled` and `experimental_history_healing`) now share it. Left the other boolean-ish sites alone deliberately: `add_mcp`'s strict validation and the frontmatter's lenient `== true` default have different intended semantics, so coercing there would change behavior.

test/riffer/config_test.rb
```diff
@@ -177,4 +177,53 @@
...
+      expect(config.tracing.tracer_provider).must_be_nil
+    end
+
+    it "allows setting a tracer_provider when opentelemetry is bundled" do
+      skip "opentelemetry is not bundled" unless OTEL_SDK_AVAILABLE
+      config = Riffer::Config.new
+      provider = OpenTelemetry::SDK::Trace::TracerProvider.new
+      config.tracing.tracer_provider = provider
+      expect(config.tracing.tracer_provider).must_equal provider
+    end
```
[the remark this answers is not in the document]
bottrall: Fixed in 1afd480 — both `tracer_provider` test guards (the positive and the inverse) now use `Riffer::Tracing::Otel.available?`, matching exactly what the setter validates.

PR #310, opened 2026-06-12
feat: emit invoke_agent span per agent run

bottrall:
## Problem

The tracing foundation (#307) gave riffer a backend-agnostic span port, but nothing emits spans yet. An agent run is invisible in a trace: how long it took, how many LLM steps it made, what it consumed in tokens, and how it ended — completed, interrupted, or blocked by a guardrail. Per-run token usage is also unobservable at the API level: `context.token_usage` accumulates for the lifetime of the agent instance, so a host can't attribute usage to a single `generate` call on a multi-turn agent.

## Solution

Wrap `Run.run_loop` in an `invoke_agent {agent identifier}` span (kind `internal`) following the OpenTelemetry GenAI semantic conventions (post-1.37 scheme):

- **At span open** (visible to samplers): `gen_ai.operation.name=invoke_agent`, `gen_ai.agent.name`, `gen_ai.provider.name`, `gen_ai.request.model`.
- **At close**: `riffer.steps` (LLM calls made this run), aggregate `gen_ai.usage.input_tokens` / `output_tokens` / `cache_read.input_tokens` / `cache_creation.input_tokens` (omitted when nothing was reported), `riffer.interrupt.reason` on interrupts, and `riffer.tripwire.guardrail` / `riffer.tripwire.reason` / `riffer.tripwire.phase` when a guardrail blocks.
- **Status**: error only when the run raises, with `error.type` set to the exception class. Tripwires and interrupts are the system working as designed — they stay OK-status and are distinguished by the `riffer.*` attributes, so guardrail enforcement doesn't inflate error-rate monitors.

The loop body moves to a private `execute_run` that returns the response plus the per-run step count, so span annotation is a pure function over the run's outcome. The per-run usage surfaces as new public API, `Response#token_usage`, accumulated locally in the loop — `context.token_usage` keeps its cumulative semantics, and a run blocked by an after-guardrail still reports the tokens it spent. Usage values pass through as `TokenUsage` reports them (semconv-aligned by #309).

Streaming needs one subtlety: the enumerator body runs in its own fiber, where the fiber-local OTel context is empty. `stream` captures the caller's context and re-attaches it inside the enumerator, so the span parents to the trace active at the stream call and opens only when consumption starts. An abandoned enumerator closes its span at fiber GC, with correspondingly inflated duration.

Per-step `chat` and `execute_tool` spans are follow-up work, as is the docs page contracting the span shape; `docs/04_AGENT_LIFECYCLE.md` documents `Response#token_usage` here.

## Proof

CI is sufficient — `bin/rake` passes (1802 runs, 0 failures; StandardRB; Steep). New specs in `test/riffer/run_tracing_test.rb` use the SDK's in-memory exporter to assert span naming, kind, request attributes, per-run step counting, usage aggregation (including cache buckets), interrupt and tripwire attributes, status semantics per exit path, `error.type`, stream laziness, and cross-fiber parenting. Per-run `Response#token_usage` semantics (run-scoped vs cumulative, tripwire paths) are covered in `test/riffer/run_test.rb`. The no-OTEL lane (`BUNDLE_WITHOUT=opentelemetry bin/test`) passes, exercising the Null backend through the wrapped loop.

## Summary by CodeRabbit

* **New Features**
  * Agent responses now include aggregated per-run token usage; runs return step counts with final responses.

* **Documentation**
  * Agent lifecycle docs clarify cumulative context-level token tracking and per-run response token metrics.

* **Observability / Tests**
  * Enhanced tracing/tests: run spans record step counts, interrupt reasons, tripwire metadata, and aggregated token-usage (including streamed usage).

* **Providers**
  * Added canonical provider names for multiple GenAI providers for semantic-convention tracing.


lib/riffer/agent/run.rb
```diff
@@ -274,7 +302,53 @@ def append_user_message(agent, prompt, files: nil)
...
+
+    span.set_attribute("riffer.interrupt.reason", response.interrupt_reason.to_s) if response.interrupt_reason
+
+    tripwire = response.tripwire
+    return unless tripwire
+
+    guardrail_name = tripwire.guardrail.name
+    span.set_attribute("riffer.tripwire.guardrail", guardrail_name) if guardrail_name
+    span.set_attribute("riffer.tripwire.reason", tripwire.reason)
+    span.set_attribute("riffer.tripwire.phase", tripwire.phase.to_s)
```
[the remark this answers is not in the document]
bottrall: @coderabbitai Declining this one — the redaction line is drawn deliberately, just one layer down from where this suggests.

These reason strings are developer-authored control-plane metadata, not user/LLM content: guardrail authors write `block("Input blocked")` and hosts write `interrupt!("needs human approval")`. They're the same trust class as exception messages, which OTel already exports verbatim by default via `record_exception` (including on this span's error path). The channels that genuinely carry payload content are excluded: tripwire `metadata` is not emitted at all, and message-content capture is a separate opt-in that defaults off.

Redacting every String reason would defeat the attribute's purpose — "which reason blocked this run" is exactly the query it exists to answer — and a stable-code mapping can't work for host-defined reasons, since riffer doesn't know the host's vocabulary.

The fair kernel here: the tracing docs (landing with the span-contract docs page in follow-up work) will state that reason strings are exported verbatim, so authors shouldn't interpolate payload content into them — the same guidance that applies to exception messages.

(The two ast-grep "hardcoded RSA passphrase" warnings on these lines are false positives.)

lib/riffer/agent/run.rb
```diff
@@ -54,9 +78,13 @@ def run_loop(agent, stream_yielder: nil)
         loop do
           response = stream_yielder ? accumulate_streamed_response(agent, stream_yielder) : call_llm(agent)
           step += 1
+          run_steps += 1
           track_token_usage(agent, response.token_usage)
+          run_usage = sum_usage(run_usage, response.token_usage)
 
-          processed_response = run_after_guardrails(agent, response, stream_yielder, all_modifications) { |tripped| return tripped }
+          processed_response = run_after_guardrails(agent, response, stream_yielder, all_modifications) do |tripwire|
+            return [tripwire_response(agent, stream_yielder, tripwire, all_modifications, token_usage: run_usage), run_steps]
```
Thibaut: question (non-blocking): would it make sense to attach `run_steps` to `Riffer::Agent::Response` to keep the response to a simple/single object? Could also fold `run_steps` into `run_usage` as some kind of object. The number of variables being passed around is getting a bit unwieldy. Maybe we are missing an object (or instance variables).

lib/riffer/providers/base.rb
```diff
@@ -20,6 +20,18 @@ def self.skills_adapter(model = nil)
...
+  # Returns the provider name stamped as <tt>gen_ai.provider.name</tt> on
+  # trace spans; defaults to the snake_cased class name — override with the
+  # GenAI semconv well-known value when one exists.
+  #--
+  #: () -> String
+  def self.semconv_provider_name
+    class_name = name
+    return "unknown" unless class_name
+
+    Riffer::Helpers::ClassNameConverter.convert(class_name.split("::").last.to_s)
```
Thibaut, line 33: question (non-blocking): why not `raise NotImplementedError`? We seem to override this in all the providers. I'd say best to be explicit than rely on a default. Without the `raise NotImplementedError`, someone implementing a new provider could miss that they need to override the default, because nothing is raised.

Thibaut (approved):
🚀

lib/riffer/agent/run.rb
bottrall: Good catch — agreed, and done in 7e05468. Moved `steps` onto `Response` next to `token_usage`, so `execute_run` returns a single `Response` instead of the `[response, steps]` tuple and `record_run_outcome` reads every attribute straight off the response.

Skipped the other two framings deliberately: `steps` aren't usage, so folding them into the usage object would make a junk-drawer — `Response` is already the run-outcome object, so they belong directly on it. And instance variables are out because `Run` is a stateless `extend self` module ("Run just orchestrates"); ivars there would be shared mutable state across concurrent runs.

`response.steps` is now public API: LLM calls made during this run (0 when a before-guardrail blocks first), distinct from the session's cumulative count.

lib/riffer/providers/base.rb
bottrall: I'd keep the default here rather than raise — the difference from the other provider methods is load-bearing-ness.

The methods that `raise NotImplementedError` are abstract: a provider literally can't generate without `execute_generate`, so there's no sensible default and crashing is correct. `semconv_provider_name` only labels a span — a provider missing it works fine, it just gets a snake_cased name in traces. And `run_span_attributes` builds eagerly on every run even with tracing disabled (the Null backend still evaluates the attributes hash), so making it raise would crash a working custom provider on every call, tracing off, over a cosmetic attribute. That cuts against the foundation's "tracing is passive, never breaks the host" principle.

The closer precedent is `skills_adapter` right above it, which also ships a default (`MarkdownAdapter`) instead of raising, for the same reason. I've tightened the doc comment (72e03b0) to spell out the intentional optionality so it doesn't read as an oversight.

PR #321, opened 2026-06-15
refactor: fold tracing tests into their unit test files (CL2-962)

bottrall:
## Problem

The tracing/span tests lived in dedicated `*_tracing_test.rb` files, split off from each unit's behavior tests:

| Unit | Behavior tests | Tracing tests |
| --- | --- | --- |
| `Riffer::Agent::Run` | `test/riffer/run_test.rb` | `test/riffer/run_tracing_test.rb` |
| `Riffer::Providers::Base` | `test/riffer/providers/base_test.rb` | `test/riffer/providers/chat_tracing_test.rb` |
| `Riffer::Tools::Runtime` | `test/riffer/tools/runtime_test.rb` | `test/riffer/tools/runtime_tracing_test.rb` |

That split was accepted for consistency while the tracing epic ([CL2-940](https://jira-jane.atlassian.net/browse/CL2-940)/941/942) was in flight, but the standing preference is one test file per unit. This is the follow-up cleanup ([CL2-962](https://jira-jane.atlassian.net/browse/CL2-962)) to fold the tracing assertions back home now that those tickets have landed.

## Solution

Each `*_tracing_test.rb` file's contents move into its unit test file under a nested `describe "tracing"` block, and the standalone file is deleted. The nested block scopes its own `before`/`after` so the in-memory tracer-provider install and the `tracing.enabled` / `capture_messages` toggles never leak onto the file's non-tracing tests, and it keeps `skip "opentelemetry is not bundled" unless OTEL_SDK_AVAILABLE` so the no-OTEL CI lane stays green and each unit's non-tracing tests still run there.

Notable decisions a reviewer might otherwise miss:

- **Flatten vs. nest.** Cosmetic sub-`describe`s are flattened straight into the `tracing` block. Groups whose `before` is load-bearing stay nested — `chat`'s content-capture group (toggles `capture_messages = true`) and `runtime`'s `span shape` / `returned error responses` / `raised exceptions` (each runs the tool call as shared arrange). Hoisting those would leak config or spurious spans into sibling tests.
- **Runtime dedup.** The byte-identical `weather_tool_class` is dropped in favour of the parent's `let`, and `make_tool_call` is unified by adding an `arguments: "{}"` default to the parent (purely additive for existing callers); the nested copy is deleted.
- **One description rename.** `chat`'s `#generate_text` and `#stream_text` each had `it "marks the span as a client span"`; once flattened these would be duplicate test names, so they're renamed to `"marks the generate_text span as a client span"` / `"marks the stream_text span as a client span"`. This is the only change to any test's wording — all assertions are verbatim.
- The tracing-module unit tests (`tracing_test.rb`, `tracing/{null,capture,otel}_test.rb`) are already one-file-per-unit and are intentionally untouched.

## Proof

CI is sufficient — this is a pure test-layout refactor with no library code changes. Verified locally on both lanes:

- **Count preservation:** 69 tracing `it`s across the three source files (25 + 28 + 16) → 69 `it`s in the merged `describe "tracing"` blocks → exactly **69 skips** in the three targets under the no-OTEL lane.
- **Default lane** (`bin/rake` = test + StandardRB + steep): 1961 runs, 0 failures, 0 errors; StandardRB clean; Steep clean.
- **No-OTEL lane** (`BUNDLE_WITHOUT=opentelemetry bin/test`): 1961 runs, 0 failures, 0 errors — the tracing blocks skip cleanly via the guard while each host file's non-tracing tests stay green.

[CL2-940]: https://jira-jane.atlassian.net/browse/CL2-940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CL2-962]: https://jira-jane.atlassian.net/browse/CL2-962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Tests**
  * Consolidated tracing test coverage from dedicated test files into main test suites for chat providers, agent runs, and tool runtime, improving test organization and maintainability.


tsov (approved):
💅

PR #322, opened 2026-06-16
feat: compute per-model cost on token usage

bottrall:
## Problem

Riffer surfaces normalized token usage per call and per run, but no cost. Anyone wanting to track LLM spend has to maintain their own price table and join it against usage out of band. The join is also easy to get wrong: because cache-read/cache-write tokens are subsets of `input_tokens`, pricing the buckets independently double-counts cached tokens.

## Solution

Add consumer-configured per-model pricing. Via `Riffer.configure`, register per-million-token rates (`input`, `output`, optional `cache_read`/`cache_write`) keyed by the same `provider/model` id you give an agent:

```ruby
Riffer.configure do |config|
  config.pricing.set("anthropic/claude-sonnet-4-6", input: 3.0, output: 15.0, cache_read: 0.30, cache_write: 3.75)
  config.pricing.set("openai/gpt-4", input: 30.0, output: 60.0)
end
```

When a model is priced, riffer computes the call's cost and carries it on `TokenUsage#cost`. No price table ships with the gem, so an unconfigured model carries no cost (`nil`) rather than a guess.

Notable design points a reviewer might otherwise miss:

- **Computed at `TokenUsage` creation** inside each provider (resolving the `provider/model` key via a new `Repository.key_for` reverse lookup), so cost is present everywhere usage surfaces — including the streamed `TokenUsageDone` event — not just on the final aggregate.
- **Cache subsets are subtracted before the input rate applies**, so cached tokens aren't double-counted; an unset cache rate falls back to the input rate.
- **Run-level cost sums through `TokenUsage#+`**, but is poison-`nil`: if any call in the run used an unpriced model, the aggregate is `nil` rather than a partial total masquerading as complete.
- Cost is a `Float` for observability, not billing.

## Proof

CI is sufficient — `bin/rake` (test + StandardRB + Steep) is green at 1990 runs, 0 failures. New tests cover the cost formula (subset subtraction, input-rate fallback, nil buckets, negative-input clamp), pricing config validation and lookup, the provider-class reverse lookup, `TokenUsage` cost/aggregation/serialization, and an end-to-end pass through the mock provider for both the generate and stream paths.

## Summary by CodeRabbit

* **New Features**
  * Token usage now includes an optional `cost` field computed from newly supported per-model pricing (input/output plus optional cache read/write tiers).
  * When pricing is configured, cost is surfaced in both returned token usage and streamed token-usage events; unpriced portions result in `cost` remaining `nil`.
* **Documentation**
  * Added a Pricing section explaining configuration, cache pricing fallback behavior, and cost aggregation semantics.
* **Tests**
  * Added/extended tests covering pricing configuration, cost computation, and `cost` serialization.

### DIFF
--- a/docs/10_CONFIGURATION.md
+++ b/docs/10_CONFIGURATION.md
@@ -128,12 +128,15 @@ Riffer.configure do |config|
   # Rates are per million tokens, keyed by the same "provider/model" id you give the agent.
   config.pricing.set("anthropic/claude-sonnet-4-6", input: 3.0, output: 15.0, cache_read: 0.30, cache_write: 3.75)
   config.pricing.set("openai/gpt-4", input: 30.0, output: 60.0)
+
+  # Pass an array to share one set of rates across a model family:
+  config.pricing.set(["openai/gpt-4", "openai/gpt-4-0613"], input: 30.0, output: 60.0)
 end
 ```
 
-| Argument       | Description                                                                                                                               |
-| -------------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
-| `model`        | The exact `provider/model` id (e.g. `"openai/gpt-4"`) — the same string you pass to `model`. No alias matching; raises on a malformed id. |
+| Argument       | Description                                                                                                                                                              |
+| -------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `models`       | A `provider/model` id (e.g. `"openai/gpt-4"`) — the same string you pass to `model` — or an array of ids that share one set of rates. No alias matching; raises on a malformed id. |
 | `input:`       | Price per **million** input tokens. Required. Applies to the uncached portion of `input_tokens`.                                          |
 | `output:`      | Price per **million** output tokens. Required.                                                                                            |
 | `cache_read:`  | Price per million cache-read tokens. Optional — when omitted, cache reads bill at the `input:` rate.                                      |
--- a/docs/10_CONFIGURATION.md
+++ b/docs/10_CONFIGURATION.md
@@ -141,7 +141,7 @@ end
 
 Because the cache buckets are subsets of `input_tokens`, the cost formula subtracts them before applying the input rate:
 
-```
+```text
 cost = (input − cache_read − cache_write) × input_rate
      + cache_read  × cache_read_rate
      + cache_write × cache_write_rate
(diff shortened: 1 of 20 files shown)


test/riffer/config/pricing_test.rb
```diff
@@ -0,0 +1,104 @@
...
+  describe "#cost_for" do
+    it "prices input and output at the per-million rates" do
+      rates = Riffer::Config::Pricing::Rates.new(input: 3.0, output: 15.0)
+      cost = rates.cost_for(input_tokens: 2_000_000, output_tokens: 1_000_000)
+      expect(cost).must_equal 21.0
+    end
+
+    it "subtracts the cache subsets and prices them at their own rates" do
+      rates = Riffer::Config::Pricing::Rates.new(input: 3.0, output: 15.0, cache_read: 1.0, cache_write: 5.0)
+      cost = rates.cost_for(input_tokens: 4_000_000, output_tokens: 0, cache_read_tokens: 1_000_000, cache_write_tokens: 1_000_000)
```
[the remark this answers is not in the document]
bottrall: #cycode_secret_false_positive it's an integer used in a test

docs/10_CONFIGURATION.md
```diff
@@ -119,6 +119,37 @@ end
...
 Hosts own SDK and exporter wiring — riffer only emits spans through whatever provider the host configures. See [Tracing](16_TRACING.md) for the emitted span contract — names, attributes, hierarchy, and host wiring.
 
+### Pricing
+
+Configure per-model token prices and riffer computes the cost of each LLM call onto its [`TokenUsage`](08_MESSAGES.md#token-usage-semantics). Riffer ships **no** price table — so an unconfigured model simply carries no cost (`token_usage.cost` is `nil`).
+
+```ruby
+Riffer.configure do |config|
+  # Rates are per million tokens, keyed by the same "provider/model" id you give the agent.
+  config.pricing.set("anthropic/claude-sonnet-4-6", input: 3.0, output: 15.0, cache_read: 0.30, cache_write: 3.75)
```
Thibaut, line 129: thought (non-blocking): since this configuration would typically grow to quite a long list of models and properties, I wonder if we should consider YAML.

Thibaut, line 129: question (non-blocking): there's going to be a lot of duplication across models of the same family. Should we support passing an array of models as first input?

Thibaut (approved):

docs/10_CONFIGURATION.md
bottrall: Agreed, however I'm not sure if it's riffer's place to solve that. It's trivial for a consumer to store this in yaml or json and load it in themselves then loop over it.

For now lets defer it to the consumer, if it becomes a pain we can re-evaluate riffer's role here.

bottrall: Good call https://github.com/janeapp/riffer/pull/322/commits/3259255cc2edf34a7b8eca7b1fba335fd90475f2

janetxinli (approved):

PR #323, opened 2026-06-16
feat: surface cost on LLM-call and run spans

bottrall:
## Problem

Riffer already traces LLM calls and computes per-model cost on `TokenUsage`, but that cost was never surfaced on the spans themselves. The trace contract bills itself as something you "build dashboards, alerts, and cost reporting against" — yet cost reporting was the one thing you couldn't do, because the number stopped at the message and never reached a span attribute. This PR is the trace-surfacing step of that cost-attribution work; its dependencies (LLM-call tracing and per-model pricing on `TokenUsage`) already landed.

## Solution

Stamp the cost carried on `TokenUsage` as `riffer.cost` on both the `chat` (per-call) and `invoke_agent` (run total) spans. The stamping is a single conditional folded into `Tracing.record_usage`, so all three call sites — generate, stream, and run-level aggregation — inherit it for free, mirroring how cache-token attributes are already handled.

A few deliberate decisions, all documented in the trace contract:

- **`riffer.cost`, not `gen_ai.*`.** The GenAI semantic conventions define no cost attribute by design, so squatting `gen_ai.*` would be wrong; cost lives in Riffer's own namespace.
- **Unitless on the wire.** The value is the sum of the consumer's configured per-token rates in whatever currency they chose — Riffer attaches no currency and emits the raw, unrounded float.
- **Nil-only guard.** Unpriced calls (cost `nil`) carry no attribute; a genuine `0.0` is still stamped.
- **Run cost is all-or-nothing.** Because cost sums with `nil` as absorbing, a single unpriced call leaves the run-level `riffer.cost` absent rather than reporting a partial that silently under-reports spend. The per-call `chat` costs are still present individually. This trap is called out in the docs.

No manual CHANGELOG edit — the `feat:` commit drives release-please.

## Proof

CI is sufficient — covered by new tests, and `bin/rake` (test + Standard + Steep) is green locally:

- `test/riffer/tracing_test.rb` — `record_usage` stamps `riffer.cost` when present, stamps a `0.0` (locks the nil-only guard), and omits it when absent.
- `test/riffer/providers/base_test.rb` — the `chat` span carries `riffer.cost` in both the generate and stream paths.
- `test/riffer/run_test.rb` — the `invoke_agent` span sums per-call cost, and omits `riffer.cost` on a genuinely mixed run (one priced call, one unpriced) — the all-or-nothing trap.

## Summary by CodeRabbit

* **New Features**
  * Tracing now captures a `riffer.cost` span attribute derived from provider token usage.
  * Run spans aggregate costs across tool-loop calls when every call is priced; otherwise the run cost is omitted.
  * Chat/model spans include per-call costs when available (including `0.0` when the provider reports zero).
* **Documentation**
  * Updated tracing docs to define `riffer.cost`, its wire behavior, and stability expectations.
* **Tests**
  * Added coverage for cost stamping in both non-streaming and streaming provider flows, plus run-span aggregation rules.

### DIFF
--- a/docs/16_TRACING.md
+++ b/docs/16_TRACING.md
@@ -82,6 +82,7 @@ The contract promise is: **when present**, a key carries the documented meaning 
 | `gen_ai.usage.output_tokens`               | int    | When the run made an LLM call that reported usage    |
 | `gen_ai.usage.cache_read.input_tokens`     | int    | When the provider reported cache reads               |
 | `gen_ai.usage.cache_creation.input_tokens` | int    | When the provider reported cache writes              |
+| `riffer.cost`                              | float  | When every call in the run was priced                |
 | `riffer.interrupt.reason`                  | string | On interrupt (e.g. approval needed, max steps)       |
 | `riffer.tripwire.guardrail`                | string | On a guardrail tripwire, when the guardrail is named |
 | `riffer.tripwire.reason`                   | string | On a guardrail tripwire                              |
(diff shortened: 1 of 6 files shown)


tsov (approved):
🚀

PR #325, opened 2026-06-17
feat: add the Riffer::Metrics OpenTelemetry port foundation

bottrall:
## Problem

Riffer emits OpenTelemetry traces but has no meter anywhere — there is no way to record metrics. The native GenAI-semconv client metrics (operation duration, token usage) and a Riffer-owned cost metric all need a metrics port before any of them can be emitted. This PR is that infrastructure: the port and configuration, with **no instruments emitted yet**.

## Solution

Add a `Riffer::Metrics` port mirroring `Riffer::Tracing`, plus `Riffer::Config::Metrics`.

- `Riffer::Metrics.create_histogram(name, unit:, description:)` returns a `Histogram` handle that is safe to hold as a constant: it defers the `enabled` gate, backend resolution, and instrument creation to record time, so it survives a meter-provider swap (e.g. between tests) and a runtime `enabled` flip. Instruments are created once and cached on the resolved backend; `reset!` discards the backend and clears that cache transitively.
- `Otel` and `Null` backends, resolved lazily. No hard OpenTelemetry dependency — every measurement is a silent no-op when the host SDK is absent.
- The OpenTelemetry metrics API is a **separate, still-pre-1.0 gem** from the 1.x traces API, so the port carries its own version guard (`opentelemetry-metrics-api >= 0.2, < 1.0`) and warns + falls back to no-op outside that range.
- The metrics API has no advisory for explicit bucket boundaries, so histogram buckets are a **host-side View concern** — documented as such rather than set by riffer.
- `Riffer::Config::Metrics` (`enabled` + `meter_provider`) is **independent** of `config.tracing`, so a host can run one signal without the other — the lever for avoiding double-counting when a provider-level instrumentation already emits the GenAI metrics.
- Docs: new `docs/17_METRICS.md` (framework, stability tiers, and double-instrumentation guidance; per-instrument tables land with the instruments) plus a Configuration section.

No instruments are emitted yet; this unblocks the metric-emitting work that follows.

## Proof

CI is sufficient — new tests cover the port end to end:

- `test/riffer/metrics_test.rb`, `test/riffer/metrics/otel_test.rb`, `test/riffer/metrics/null_test.rb`, and a metrics section in `test/riffer/config_test.rb`.
- They record through the port against an in-memory meter provider and assert the instrument name, attributes, unit/description, instrumentation scope, and that recording is a no-op when disabled.
- The no-OTEL CI lane (`BUNDLE_WITHOUT=opentelemetry`) exercises the `Null` fallback — verified locally that the suite stays green with the metrics SDK absent.
- `bin/rake` (test + standard + steep:check) passes locally.

## Summary by CodeRabbit

* **New Features**
  * Added OpenTelemetry metrics support with an independent `metrics` configuration switch and optional `meter_provider`.
  * Introduced histogram metric recording via the configured backend, with safe no-op behavior when metrics aren’t available.
* **Documentation**
  * Added a dedicated metrics contract/setup guide and expanded the configuration chapter with `metrics` options, examples, and notes on host-side setup.
  * Updated the overview and README to reference metrics documentation.
* **Tests**
  * Added coverage for metrics configuration, OpenTelemetry metrics availability/version gating, and histogram record/reset behavior.


lib/riffer/metrics/otel.rb
```diff
@@ -0,0 +1,79 @@
...
+    # Builds a backend when the OpenTelemetry metrics API is loadable at a
+    # supported version; returns +nil+ so resolution falls back to Null.
+    #--
+    #: (provider: untyped) -> Riffer::Metrics::Otel?
+    def build(provider:)
+      version = api_version
+      return nil unless version
+
+      unless supported?(version)
+        Kernel.warn "riffer: opentelemetry-metrics-api #{version} is outside the supported range (#{SUPPORTED_API_VERSIONS}); metrics are disabled"
```
Thibaut, line 20: thought (non-blocking): it's pretty standard practice for framework-style gems like Riffer to have a configurable logger (example: [sidekiq](https://github.com/sidekiq/sidekiq/wiki/Logging)) to route logs like above in different ways. Maybe something to consider for later.

Thibaut (approved):

lib/riffer/metrics/otel.rb
bottrall: Agreed 👍🏻 will think about this as a follow up

PR #326, opened 2026-06-17
feat: emit gen_ai.client.operation.duration metric

bottrall:
## Problem

Riffer records GenAI spans but no metrics. Operation latency could only be derived from span duration in a collector, which is sampling-dependent and post-processed. The OpenTelemetry GenAI semantic conventions define `gen_ai.client.operation.duration` as the canonical latency metric — sampling-independent, pre-aggregated client-side, and the key off-the-shelf GenAI dashboards build on. Riffer landed the metrics port foundation but recorded no instruments yet; this ships the first one.

## Solution

Record a `gen_ai.client.operation.duration` histogram (unit `s`) around all three GenAI operations — `chat`, `invoke_agent`, and `execute_tool` — reusing each operation's existing span wrap so the metric and span stay consistent, and distinguishing them by `gen_ai.operation.name`. A single shared instrument handle lives in a new `Riffer::Metrics::Instruments` catalog.

Timing uses a monotonic clock and records once on both the success and error paths via an `ensure`, independent of whether tracing is enabled, and no-ops when the host bundles no metrics SDK (no OTel dependency is added).

Attributes are scoped per operation rather than a uniform set, mirroring what each span already carries:

- `chat` → `gen_ai.operation.name`, `gen_ai.provider.name`, `gen_ai.request.model` (when set), `error.type` (on error)
- `invoke_agent` → the above plus `gen_ai.agent.name`
- `execute_tool` → `gen_ai.operation.name`, `gen_ai.tool.name`, `error.type` (on error) — the tool runtime has no provider/model reference, and `gen_ai.tool.call.id` is deliberately kept span-only to avoid unbounded metric cardinality

`error.type` mirrors the span: the raised exception class for `chat`/`invoke_agent`, and a tool's returned handled-error category (e.g. `validation_error`, `timeout_error`) or raised class for `execute_tool`.

Notable trade-offs, all documented in `docs/17_METRICS.md`:

- **`gen_ai.response.model` is deferred** — it isn't captured on any span today, so adding it is left to a follow-up that puts it on the chat span and this metric together, keeping them consistent.
- **Streamed `chat`/`invoke_agent` durations are consumption-paced** — reusing the span's lazy-enumerator wrap means the duration spans first-pull-to-drain, so a slow consumer inflates it (same as the span behaves).
- **`gen_ai.tool.name` cardinality** can grow with dynamic/MCP tool sets; the docs note hosts can drop it via a View.
- **Bucket boundaries stay host-side** — the metrics API can't attach them at instrument creation; the docs point at a View for the semconv-recommended boundaries.

## Proof

CI is sufficient. Contract tests are folded into the operation unit tests (`base_test.rb`, `run_test.rb`, `runtime_test.rb`), asserting the metric name, unit, and per-operation attributes on success and `error.type` on error — including both tool error modes (handled Response and raised exception), the stream-drain path, and recording while tracing is disabled. `bin/rake` (test + standard + steep:check) is green; RBS signatures are regenerated.

## Summary by CodeRabbit

* **New Features**
  * Added operation-duration metrics for AI client actions—chat, agent invocation, and tool execution—with consistent timing and `error.type` attribution when failures occur.
  * Metrics are emitted reliably even when tracing is disabled, including after streaming completes.

* **Documentation**
  * Updated the metrics specification to define `gen_ai.client.operation.duration`, units, and expected attributes by operation type.

* **Tests**
  * Added/extended OpenTelemetry metrics tests to validate datapoints, attribute sets, error recording, and stream-drain behavior across operations.

### DIFF
--- a/docs/17_METRICS.md
+++ b/docs/17_METRICS.md
@@ -35,7 +35,23 @@ Histogram bucket boundaries are a **host-side** concern. The OpenTelemetry metri
 
 ## Instruments
 
-This release lands the metrics **port and configuration only** — Riffer records **no instruments yet**. Each instrument is documented here as it ships, as a row carrying its name, instrument type, unit, and attribute set. Until then there is nothing on the wire to query.
+Each instrument is documented here as a row carrying its name, instrument type, unit, and attribute set.
+
+### `gen_ai.client.operation.duration`
+
+Histogram, unit `s`. The latency of a single GenAI operation, recorded around the same wrap as the matching [span](16_TRACING.md) on both the success and error paths and timed with a monotonic clock. Recording is independent of tracing — the metric fires even with `config.tracing.enabled = false`. Tell the three operations apart by `gen_ai.operation.name`.
+
+| `gen_ai.operation.name` | Recorded around                                  | Attributes                                                                                                                |
+| ----------------------- | ------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------- |
+| `chat`                  | each provider call (`generate_text`/`stream_text`) | `gen_ai.operation.name`, `gen_ai.provider.name`, `gen_ai.request.model` (when set), `error.type` (on error)               |
+| `invoke_agent`          | each agent run (`generate`/`stream`)             | `gen_ai.operation.name`, `gen_ai.provider.name`, `gen_ai.request.model`, `gen_ai.agent.name`, `error.type` (on error)     |
+| `execute_tool`          | each tool call                                   | `gen_ai.operation.name`, `gen_ai.tool.name`, `error.type` (on error)                                                      |
+
+`error.type` carries the exception class for a raised error; for `execute_tool` it carries the handled error category (e.g. `validation_error`, `timeout_error`) when a tool returns an error result instead of raising — matching the span. `gen_ai.response.model` is not recorded yet; it will land once it is also captured on the chat span.
+
+> **Streamed operations are consumption-paced.** A streamed `chat` or `invoke_agent` records its duration when the stream drains, so the value includes the time your consumer takes to iterate the events, not just provider latency. The matching span behaves the same way.
+
+> **`gen_ai.tool.name` cardinality.** One time series exists per distinct tool name. With a large or dynamic tool set (for example MCP-discovered tools) that can grow unbounded — drop the attribute with a [View](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#view) if your backend strains.
 
 ## Stability
 
--- a/lib/riffer/tools/runtime.rb
+++ b/lib/riffer/tools/runtime.rb
@@ -25,13 +25,7 @@ class Riffer::Tools::Runtime
     trace_context = Riffer::Tracing.current_context
     @runner.map(tool_calls, context: context) do |tool_call|
       Riffer::Tracing.with_context(trace_context) do
-        in_tool_span(tool_call) do |span|
-          result = around_tool_call(tool_call, context: context, assistant_message: assistant_message) do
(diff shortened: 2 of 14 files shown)


docs/17_METRICS.md
```diff
@@ -35,7 +35,23 @@ Histogram bucket boundaries are a **host-side** concern. The OpenTelemetry metri
...
+
+| `gen_ai.operation.name` | Recorded around                                  | Attributes                                                                                                                |
+| ----------------------- | ------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------- |
+| `chat`                  | each provider call (`generate_text`/`stream_text`) | `gen_ai.operation.name`, `gen_ai.provider.name`, `gen_ai.request.model` (when set), `error.type` (on error)               |
+| `invoke_agent`          | each agent run (`generate`/`stream`)             | `gen_ai.operation.name`, `gen_ai.provider.name`, `gen_ai.request.model`, `gen_ai.agent.name`, `error.type` (on error)     |
+| `execute_tool`          | each tool call                                   | `gen_ai.operation.name`, `gen_ai.tool.name`, `error.type` (on error)                                                      |
```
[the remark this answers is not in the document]
bottrall: Holding on this one — the asymmetry is intentional and matches the code.

For `chat`, `request.model` is conditional: `chat_metric_attributes` sets it only `if model`, because `generate_text` can be called without a model — hence "(when set)".

For `invoke_agent` it's unconditional: `run_metric_attributes` sets `"gen_ai.request.model" => agent.model_name`, and `agent.model_name` is a non-nil `String` resolved from the agent's `provider/model` at construction (it raises if either half is missing). An agent can't run without a model, so the attribute is always present.

Adding "(when set)" here would imply the agent metric sometimes omits the model, which it never does. The "same as chat plus agent name" phrasing in the PR description was a simplification; the rows differ because the behaviour differs.

lib/riffer/tools/runtime.rb
```diff
@@ -58,6 +52,29 @@ def around_tool_call(tool_call, context:, assistant_message: nil)
...
   private
 
+  #--
+  #: (Riffer::Messages::Assistant::ToolCall, tools: Array[singleton(Riffer::Tool)], context: Riffer::Agent::Context?, assistant_message: Riffer::Messages::Assistant?) -> [Riffer::Messages::Assistant::ToolCall, Riffer::Tools::Response]
+  def run_tool_call(tool_call, tools:, context:, assistant_message:)
+    start = Riffer::Metrics.monotonic_now
+    error_type = nil #: String?
+    begin
+      pair = in_tool_span(tool_call) do |span|
+        result = around_tool_call(tool_call, context: context, assistant_message: assistant_message) do
```
Thibaut: suggestion (non-blocking): this kind of buries the core logic of `run_tool_call` and `execute`, which is this block here. If possible, I'd suggest extracting the whole instrumentation/tracing stack in a method that takes a block, and then inline the core "business" logic in `execute`. Something like:

```ruby
instrument(...) do
  result = around_tool_call(tool_call, context: context, assistant_message: assistant_message) do
    dispatch_tool_call(tool_call, tools: tools, context: context, assistant_message: assistant_message)
  end
  record_tool_outcome(span, result)
end
```

then reading `execute` and understanding how tools are ultimately called would be much more straightforward.

Thibaut (approved):

lib/riffer/tools/runtime.rb
bottrall: good call https://github.com/janeapp/riffer/pull/326/commits/c7352a8508b51dc9128a8f1525d33487bc0d856b

PR #328, opened 2026-06-18
feat: emit riffer.gen_ai.cost metric from TokenUsage cost

bottrall:
## Problem

Spend is observable today only as a span attribute (`riffer.cost`). That makes it inspectable per call but not queryable as an aggregate — there's no way to get total spend or a per-request cost distribution (p95, etc.) the way latency (`gen_ai.client.operation.duration`) and token volume (`gen_ai.client.token.usage`) already expose as metric histograms. This adds the metric-side counterpart to the cost already surfaced on spans.

## Solution

Add a `riffer.gen_ai.cost` histogram (unit `USD`), recorded once per `chat` call from the cost already attached to the normalized `TokenUsage`. No pricing is computed here — the value is read straight off `TokenUsage#cost`, the same source as the cost span attribute, just a different sink.

Notable behaviors a reviewer might otherwise miss:

- **Owned namespace.** It lives under `riffer.*`, not `gen_ai.*`, so it won't collide if the GenAI semantic conventions later define a cost instrument.
- **Skip-when-unpriced.** Pricing is consumer-configured and no price table ships with the gem, so an unpriced model yields `nil` cost and records no data point — `operation.duration` and `token.usage` still record. A priced call that computes to `0.0` does record a zero point; only an absent price means there's nothing to measure (`return unless cost`, relying on `0.0` being truthy in Ruby).
- **Per-call only**, never run-level, to avoid double-counting under metric pre-aggregation — matching how token usage is handled.
- Recorded on both the `generate_text` and stream-drain paths, and independent of tracing (fires with `config.tracing.enabled = false`).
- Attributes match the sibling chat metrics by reusing the shared base: `gen_ai.operation.name` (always `chat`), `gen_ai.provider.name`, and `gen_ai.request.model` (when set). Bucket boundaries stay host-side, like the other histograms.

## Proof

CI is sufficient — see the new contract tests in `test/riffer/providers/base_test.rb` under `describe "metrics"`. They assert the metric name + unit (`["riffer.gen_ai.cost", "USD"]`), the recorded value, the full attribute set, that nothing is emitted when the call carries no cost, and that it records on both the stream-drain and tracing-disabled paths. `bin/rake` (test + StandardRB + Steep) is green, and `sig/generated/` is regenerated from the rbs-inline annotations.

## Summary by CodeRabbit

* **New Features**
  * Added cost tracking metric for GenAI operations, recording USD-denominated cost per call with operation-level attributes.

* **Documentation**
  * Updated metrics reference with expanded histogram configuration guidance and new cost metric specifications.

### DIFF
--- a/docs/17_METRICS.md
+++ b/docs/17_METRICS.md
@@ -33,7 +33,7 @@ Instruments are recorded under the instrumentation scope named `riffer`, version
 
 Histogram bucket boundaries are a **host-side** concern. The OpenTelemetry metrics API does not let an instrumenting library attach bucket boundaries at instrument creation, so Riffer does not set them — the SDK's default buckets apply unless you override them. To match the GenAI semantic conventions' recommended boundaries (or your own), register a [View](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#view) on the meter provider that targets the instrument by name and sets explicit bucket boundaries.
 
-The convention recommends boundaries scaled to each instrument, so register one View per histogram — the token-count buckets below are for `gen_ai.client.token.usage`; `gen_ai.client.operation.duration` wants its own latency-scaled set.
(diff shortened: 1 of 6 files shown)


tsov (approved):
Nice 🚀

PR #329, opened 2026-06-19
feat(telemetry): pluggable tracing/metrics backend (OTEL opt-in)

bottrall:
## Problem

riffer's native telemetry only reached a host that bundles the OpenTelemetry gems — and for tracing, only one running the full OTEL **SDK** (the API-only `datadog/opentelemetry` bridge leaves `OpenTelemetry.tracer_provider` a no-op proxy, so no spans are produced). A host already invested in another stack — Datadog APM + DogStatsD — had no way to receive riffer's spans/metrics without adopting OpenTelemetry. Telemetry hooks shouldn't force a consumer to adopt a specific vendor.

On top of that, OTEL was **auto-detected**: whenever the gem happened to be loaded, riffer resolved the global provider and started emitting. That's surprising for a host that wants to opt in deliberately (or route telemetry elsewhere).

## Solution

Each telemetry signal now resolves through a single, consumer-assigned backend, and OpenTelemetry becomes one backend you opt into rather than a default.

- `config.tracing.backend` / `config.metrics.backend` — duck-typed, validated on assignment (`respond_to?(:in_span)` / `:record_histogram`); `nil` or a non-conforming object raises `Riffer::ArgumentError`.
- Resolution is simply `config.X.backend || Null` — **no auto-detection**, no `Otel.build` fallback. An unset backend is a silent no-op and riffer carries no OTEL gem dependency.
- OpenTelemetry ships as a built-in backend the consumer assigns explicitly: `config.tracing.backend = Riffer::Tracing::Otel.build` (and the metrics equivalent). `build` defaults to the global provider, accepts `provider:` to override, and returns `nil` (→ no-op) when the gem is absent/unsupported — so the same config is safe without OTEL installed.
- `Riffer::Metrics.recording?` is generalized from `backend.is_a?(Otel)` to "any live (non-Null) backend", so providers still compute the token/cost data a custom metrics sink needs.

**Breaking change (1.0 work):** `config.tracing.tracer_provider` and `config.metrics.meter_provider` are removed. Assign the OTEL backend instead — `config.tracing.backend = Riffer::Tracing::Otel.build(provider: ...)`. Docs (`10_CONFIGURATION`, `16_TRACING`, `17_METRICS`) are rewritten to lead with the opt-in model.

The consumer side wires a `Datadog::Tracing` tracing backend and a `$stats` DogStatsD metrics backend straight into these setters — no `opentelemetry-*` gem required.

## Proof

**CI is sufficient.** Full `bin/rake` gate (test + Standard + Steep) is green, and both lanes pass — the default lane (OTEL bundled) and the no-OTEL lane (`BUNDLE_WITHOUT=opentelemetry`, the `test-no-otel` CI job). RBS regenerated with no drift. Coverage added/updated for the change:

- `test/riffer/config_test.rb` — `backend` setter validation for both signals: accepts a conforming object, clears with `nil`, raises otherwise.
- `test/riffer/tracing_test.rb` — delegation of `in_span` / `current_context` / `with_context` to a fake backend, a no-op span when no backend is configured, and the `enabled` kill switch short-circuiting ahead of the backend.
- `test/riffer/metrics_test.rb` — record delegation and the four `recording?` cases (custom backend live, disabled, no backend, OTEL backend live).
- `test/riffer/{tracing,metrics}/otel_test.rb` — `build` with no args wraps the global provider; returns `nil` without the gem.
- `test_helper.rb` + `run_test` / `runtime_test` / `base_test` — in-memory provider helpers now assign `Riffer::X::Otel.build(provider:)`.

## Summary by CodeRabbit

* **Breaking Changes**
  * Tracing and metrics now require explicit backend configuration (no OpenTelemetry auto-detection).
  * Renamed `config.tracing.tracer_provider` → `config.tracing.backend` and `config.metrics.meter_provider` → `config.metrics.backend`.
  * When no backend is set (or when disabled), tracing/metrics become no-ops with no implicit OpenTelemetry dependency. Custom backends must implement the expected span/metrics interfaces.

* **Documentation**
  * Updated tracing/metrics configuration and backend routing guides, including the `enabled`/`backend` options and how to opt into OpenTelemetry.

* **Tests**
  * Refreshed and extended tests to validate backend-based routing and the new option names.


lib/riffer/tracing/otel.rb
```diff
@@ -52,10 +52,11 @@ def recording?
...
   class << self
     # Builds a backend when the OpenTelemetry API is loadable at a supported
-    # version; returns +nil+ so resolution falls back to Null.
+    # version; returns +nil+ so resolution falls back to Null. +provider+
+    # defaults to the global <tt>OpenTelemetry.tracer_provider</tt>.
     #--
-    #: (provider: untyped) -> Riffer::Tracing::Otel?
-    def build(provider:)
+    #: (?provider: untyped) -> Riffer::Tracing::Otel?
+    def build(provider: nil)
```
cloetzi, line 59: suggestion(non-blocking): `def build(provider: ::OpenTelemetry.tracer_provider)`

That makes the default obvious when reading the code and would allow to remove the default selection below.

lib/riffer/config.rb
```diff
@@ -91,17 +91,15 @@ def capture_messages=(value)
...
-      if !value.nil? && !Riffer::Tracing::Otel.available?
-        raise Riffer::ArgumentError,
-          "tracer_provider requires the opentelemetry-api gem (#{Riffer::Tracing::Otel::SUPPORTED_API_VERSIONS})"
+    def backend=(value)
+      unless value.nil? || value.respond_to?(:in_span)
+        raise Riffer::ArgumentError, "tracing backend must respond to #in_span"
       end
-      @tracer_provider = value
+      @backend = value
       Riffer::Tracing.reset!
```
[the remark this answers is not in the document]
cloetzi: question(non-blocking): I think this is kinda valid, is there a reason we only check exactly one method existence when the API expects three or four?

lib/riffer/metrics.rb
```diff
@@ -80,14 +81,14 @@ def reset!
...
   def backend
     @backend || MUTEX.synchronize { @backend ||= resolve_backend }
   end
 
   #--
-  #: () -> (Riffer::Metrics::Otel | singleton(Riffer::Metrics::Null))
+  #: () -> untyped
   def resolve_backend
-    Otel.build(provider: Riffer.config.metrics.meter_provider) || Null
+    Riffer.config.metrics.backend || Null
```
cloetzi, line 92: nitpick(non-blocking): I found the `Null` naming a bit confusing. Maybe `None`, or `NoOp` would be more clear that it is a custom class and not a built-in. Doesn't matter in practice of course.

cloetzi (approved):

lib/riffer/tracing/otel.rb
bottrall: While I agree this would be nicer, setting it as the default would bring `OpenTelemetry` into the global scope and raise a `NameError` for consumers without the otel gem installed. So we've got to keep it as `nil`.

lib/riffer/metrics.rb
bottrall: Agreed - but I think that's out of scope for this PR. Will address in a follow up.

lib/riffer/config.rb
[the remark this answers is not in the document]
bottrall: https://github.com/janeapp/riffer/pull/329/commits/329b423ccfce15ad1a74bf95b878e5b7d6d4f4da

PR #338, opened 2026-06-22
feat: emit a riffer.guardrail.duration metric for guardrail execution

bottrall:
## Problem

The metrics port records `gen_ai.client.operation.duration` for three operations — `chat`, `invoke_agent`, and `execute_tool` — but not for guardrails. `Riffer::Guardrails::Runner#execute_guardrail` already opens an `execute_guardrail` span, yet records no metric, so guardrail latency is invisible to any metrics backend (OTEL or a custom non-OTEL sink). Guardrails run on the request hot path — before every model turn and after every response — so their cost is worth measuring. This adds the metrics-side counterpart to the existing `execute_guardrail` span.

## Solution

Add a new histogram `riffer.guardrail.duration` (unit `s`) to `Riffer::Metrics::Instruments` and record it around `execute_guardrail`, timed with `Riffer::Metrics.monotonic_now`, on both the success and raise paths — the same `begin/rescue/ensure` wrap used by the `chat` and `execute_tool` duration metrics.

The instrument is deliberately **riffer-owned** (`riffer.*`) rather than folded into `gen_ai.client.operation.duration`. A guardrail is not a GenAI semantic-convention operation, so the `execute_guardrail` span carries no `gen_ai.operation.name` and lives in riffer's own `riffer.guardrail.*` namespace; reusing the GenAI duration instrument with a squatted operation name would contradict that. Keeping it riffer-owned also means it won't collide if the semantic conventions later define an equivalent.

Attributes mirror the span: `riffer.guardrail.name`, `riffer.guardrail.phase`, and `error.type` on a raise. A pass, transform, or block is a handled outcome and records no `error.type`. Guardrail names are bounded by the configured guardrails, so cardinality is safe — unlike dynamic tool names. Recording stays independent of tracing, firing even with `config.tracing.enabled = false`.

One judgment call: `guardrail_metric_attributes` reuses `guardrail_span_attributes` as its base rather than duplicating the keys (unlike `providers/base.rb`, which keeps a separate base because its chat span carries extra request params the metric must not have). For guardrails the span and metric attribute sets are identical, so reuse is correct; if a span-only attribute is ever added, that would need revisiting.

## Proof

CI is sufficient. New Minitest coverage in `test/riffer/guardrails/runner_test.rb` asserts a duration data point is recorded on pass, transform, and block (with no `error.type`), and that `error.type` is recorded on a raise, plus the name/phase attributes — using the same in-memory meter-provider pull exporter as the existing `execute_tool` metric tests. RBS signatures were regenerated (`bin/rbs`) and the full gate (`bin/rake`: test + standard + steep:check) is green: 2090 runs, 0 failures, 0 errors. The instrument and its attributes are documented in `docs/17_METRICS.md` with the `riffer.*` stability tier noted.

## Summary by CodeRabbit

* **New Features**
  * Added `riffer.guardrail.duration` histogram to measure guardrail execution latency (seconds) across outcomes.
  * Duration metrics include guardrail name and phase (`before`/`after`).
* **Bug Fixes**
  * Guardrail duration metric attributes now include `error.type` when a guardrail raises (absent when errors are handled).
* **Documentation**
  * Updated the non-OpenTelemetry metrics contract to describe the new instrument, attributes, and expected units.
* **Tests**
  * Added tests validating duration recording for pass/transform/block and correct `error.type` behavior.

### DIFF
--- a/lib/riffer/guardrails/runner.rb
+++ b/lib/riffer/guardrails/runner.rb
@@ -81,9 +81,7 @@ class Riffer::Guardrails::Runner
   #--
   #: (Riffer::Guardrail, untyped, messages: Array[Riffer::Messages::Base]?) -> Riffer::Guardrails::Result
   def execute_guardrail(guardrail, data, messages:)
-    start = Riffer::Metrics.monotonic_now
-    error_type = nil #: String?
-    begin
+    instrument_guardrail(guardrail) do
       Riffer::Tracing.in_span("execute_guardrail #{guardrail.name}", attributes: guardrail_span_attributes(guardrail), kind: :internal) do |span|
         result = run_guardrail_phase(guardrail, data, messages: messages)
         record_guardrail_outcome(span, result)
@@ -94,9 +92,17 @@ class Riffer::Guardrails::Runner
         span.set_attribute("error.type", error.class.name)
         raise
       end
+    end
+  end
+
+  #--
+  #: (Riffer::Guardrail) { () -> Riffer::Guardrails::Result } -> Riffer::Guardrails::Result
+  def instrument_guardrail(guardrail)
+    start = Riffer::Metrics.monotonic_now
+    error_type = nil #: String?
+    begin
+      yield
     rescue => error
-      # The inner rescue tags the span; capture error.type here too, at method
-      # scope, where the ensure can read it onto the metric.
       error_type = error.class.name #: String?
       raise
     ensure
--- a/lib/riffer/guardrails/runner.rb
+++ b/lib/riffer/guardrails/runner.rb
@@ -81,15 +81,26 @@ class Riffer::Guardrails::Runner
   #--
   #: (Riffer::Guardrail, untyped, messages: Array[Riffer::Messages::Base]?) -> Riffer::Guardrails::Result
   def execute_guardrail(guardrail, data, messages:)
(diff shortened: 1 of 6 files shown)


lib/riffer/guardrails/runner.rb
```diff
@@ -81,15 +81,26 @@ def detect_changed_indices(old_data, new_data)
   #--
   #: (Riffer::Guardrail, untyped, messages: Array[Riffer::Messages::Base]?) -> Riffer::Guardrails::Result
   def execute_guardrail(guardrail, data, messages:)
-    Riffer::Tracing.in_span("execute_guardrail #{guardrail.name}", attributes: guardrail_span_attributes(guardrail), kind: :internal) do |span|
-      result = run_guardrail_phase(guardrail, data, messages: messages)
-      record_guardrail_outcome(span, result)
-      result
+    start = Riffer::Metrics.monotonic_now
```
Thibaut: suggestion (non-blocking): extract the instrumentation code in a `instrument_guardrail` block to keep this method leaner and easier to read. I generally prefer long methods that tell a full story, except when there is logic like instrumentation that is really secondary to the business logic and encapsulated, then I think extracting a method makes sense.

Thibaut (approved):

lib/riffer/guardrails/runner.rb
bottrall: Good call 👍🏻 https://github.com/janeapp/riffer/pull/338/commits/e78505902bc9d49955234e7c34891c7214da8fdc

PR #339, opened 2026-06-23
feat: per-call tags: for Agent#generate and #stream

cloetzi:
## What

Adds an optional `tags:` hash to `Agent#generate` / `#stream` (and the class-method forms). Tags attach attribution labels to a **single** LLM call and propagate to two places: the provider's native request-metadata field, and observability (`riffer.tag.*` on every span + metric the call emits). Per-call only — no global or agent-class tier.

```ruby
agent.generate("Summarize this ticket.",
  tags: { team: "growth", feature: "summarizer", user_id: "u_abc123" })
```

## Design

- **Normalize once at the boundary.** `Run#normalize_tags` stringifies keys/values and drops `nil`-valued entries; a non-`Hash` raises `Riffer::ArgumentError`; `nil`/empty is a no-op. Everything downstream receives a clean `String=>String` map.
- **Explicit threading, no shared state.** Tags ride in `options[:tags]` for the provider channel and are passed as an ordinary argument to each span/metric builder (`run`, `chat`, `execute_tool`, `execute_guardrail`). Because a captured local crosses the streaming-enumerator and tool-worker fiber/thread boundaries for free, there's no ambient/global holder and no re-propagation. `Tools::Runtime#execute` gains an optional `tags:` kwarg (backward compatible).

### Provider mapping

| Provider | Native request field | `user_id` |
| --- | --- | --- |
| Amazon Bedrock | `request_metadata` (all tags) | ordinary entry |
| OpenAI / Azure | `metadata` (all tags) | also → `safety_identifier` |
| OpenRouter | `metadata` (all tags) | also → `user` |
| Anthropic | `metadata.user_id` **only** | the only tag forwarded; others **dropped** |
| Gemini | _(none — observability only)_ | tag only |

- **Anthropic** has no free-form request-metadata field, so only `metadata.user_id` is forwarded and all other tags are dropped from the request (still present on spans/metrics). Documented + tested.
- **Gemini** targets the Developer API (`generativelanguage`), which has no `labels` field and rejects unknown body fields — so tags are stripped from the request and propagate to observability only. Native labels would arrive with a future Vertex adapter.
- Every adapter `.except(:tags)` so the internal key never leaks to an SDK call.

## Observability

`riffer.tag.<key>` is stamped on all four span types and on the duration / token-usage / cost metrics. No allow-list — high-cardinality values (e.g. unique `user_id`) multiply metric series; documented in `17_METRICS.md`.

## Docs

- `10_CONFIGURATION.md`: new **Per-Call Tags** section (mapping table, Anthropic-drop note, Gemini/Vertex note, provider limits, cardinality caveat).
- `16_TRACING.md` / `17_METRICS.md`: shared `riffer.tag.*` notes.
- `08_MESSAGES.md`: unchanged — tags don't touch messages.

## Test plan

- `bin/rake` green: **2102 runs, 0 failures, 0 errors**; `steep` clean; `standard` clean; RBS regenerated.
- New tests: run-loop threading (generate + stream), `riffer.tag.*` presence on all four spans + duration/token metrics, symbol/`nil` coercion, no-op, non-`Hash` raise, per-adapter mapping (incl. Anthropic drop + Azure inheritance), Gemini no-`labels`.

## Summary by CodeRabbit

* **New Features**
  * Added optional per-call tags to generate and stream requests.
  * Tags now flow into tracing spans, metrics, tool execution, and guardrail reporting.
  * Several providers now pass tags through to request metadata, with special handling for `user_id` where supported.

* **Documentation**
  * Updated docs to explain tag behavior, provider mapping, and observability impact.

* **Bug Fixes**
  * Prevented tags from being sent as unsupported request fields for providers that don’t use them directly.


lib/riffer/providers/open_ai.rb
```diff
@@ -38,9 +39,17 @@ def build_request_params(messages, model, options)
...
+      **options.except(:reasoning, :tools, :structured_output, :web_search, :tags)
     } #: Hash[Symbol, untyped]
 
+    unless tags.empty?
+      params[:metadata] = tags
+      # The reserved user_id also maps to the native safety identifier while
+      # staying in metadata as an ordinary tag.
+      user_id = tags["user_id"]
+      params[:safety_identifier] = user_id if user_id
+    end
```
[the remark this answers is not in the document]
cloetzi: The doc explicitly calls this out as not handled. It was a conscious decision not to implement it.

docs/10_CONFIGURATION.md
```diff
@@ -302,6 +302,74 @@ class ReasoningAgent < Riffer::Agent
...
+what you give it and lets the provider reject anything out of bounds. Known
+limits: Bedrock `requestMetadata` ≤16 entries, key/value ≤256 chars; OpenAI /
+OpenRouter `metadata` ≤16 pairs (OpenRouter: 64-char keys, 512-char values).
+
+### Metric cardinality
+
+Tags are promoted onto metric dimensions verbatim — there is no allow-list.
+High-cardinality values (e.g. a unique `user_id` per request) will create a
+correspondingly large number of metric series in your backend. Keep
+metric-bound tags low-cardinality.
```
bottrall: **Nitpick(non-blocking):** Claude did that thing where it wraps the lines at ~100 characters.



Ok for comments but I prefer to avoid it in the docs.

docs/10_CONFIGURATION.md
```diff
@@ -302,6 +302,74 @@ class ReasoningAgent < Riffer::Agent
 end
 ```
 
+## Per-Call Tags
+
+`#generate` and `#stream` accept an optional `tags:` hash — a flat map of
+attribution labels scoped to that single call (for cost/usage attribution,
+filtering audit logs, slicing telemetry). It is **per-call only**: there is no
```
bottrall: **Suggestion(non-blocking):** Remove the line about no agent-class or global tier. I'm guessing that context leaking from the claude code session on the decision to thread it through the method arguments.

docs/10_CONFIGURATION.md
```diff
@@ -302,6 +302,74 @@ class ReasoningAgent < Riffer::Agent
 end
 ```
 
+## Per-Call Tags
```
bottrall: **Suggestion(non-blocking):** We might want to move this to `docs/AGENTS.md`.



Tags aren't necessarily config now that they're per call. But also not sure if it fits with the content in `docs/AGENTS.md` 🤔 idk.

lib/riffer/agent/run.rb
```diff
@@ -329,27 +340,51 @@ def sum_usage(current, usage)
...
     } #: Hash[String, untyped]
     attributes["error.type"] = error_type if error_type
-    attributes
+    attributes.merge(tag_attributes(tags))
+  end
+
+  # Normalizes a raw tags hash into a flat <tt>String => String</tt> map: keys
+  # and values are stringified and +nil+-valued entries dropped. Run is the
+  # single place this happens, so providers and span/metric builders downstream
+  # all receive already-clean tags.
```
bottrall: **Suggestion(non-blocking):** Remove this comment, it restates the how and doesn't offer any new information that can't be understood by the name of the method `normalize_tags`

lib/riffer/agent/run.rb
```diff
@@ -329,27 +340,51 @@ def sum_usage(current, usage)
...
+    tags.each do |key, value|
+      next if value.nil?
+      result[key.to_s] = value.to_s
+    end
+    result
+  end
+
+  # Maps normalized tags to their namespaced span/metric attribute form, e.g.
+  # <tt>{"team" => "growth"}</tt> becomes <tt>{"riffer.tag.team" => "growth"}</tt>.
+  # An empty map yields an empty hash, so merging it is a no-op.
```
bottrall: **Suggestion(non-blocking):** Same as above - remove. Doesn't add any value.

lib/riffer/agent/run.rb
```diff
@@ -10,37 +10,43 @@ module Riffer::Agent::Run
   # for prompt/files semantics.
   #
   #--
-  #: (agent: Riffer::Agent, ?prompt: String?, ?files: Array[Hash[Symbol, untyped] | Riffer::Messages::FilePart]?) -> Riffer::Agent::Response
-  def generate(agent:, prompt: nil, files: nil)
+  #: (agent: Riffer::Agent, ?prompt: String?, ?files: Array[Hash[Symbol, untyped] | Riffer::Messages::FilePart]?, ?tags: Hash[(String | Symbol), untyped]) -> Riffer::Agent::Response
+  def generate(agent:, prompt: nil, files: nil, tags: {})
     append_user_message(agent, prompt, files: files)
-    run_loop(agent)
+    run_loop(agent, tags: normalize_tags(tags))
```
bottrall: **Suggestion(non-blocking):** Normalize the tags in `run_loop`.



`generate` and `stream` both pipe to `run_loop` so we can centralize the normalization there.

lib/riffer/providers/anthropic.rb
```diff
@@ -48,18 +48,25 @@ def build_request_params(messages, model, options)
...
+      **options.except(:tools, :max_tokens, :structured_output, :web_search, :tags)
     } #: Hash[Symbol, untyped]
 
     params[:system] = partitioned_messages[:system] if partitioned_messages[:system]
 
+    # Anthropic's only request-metadata field is metadata.user_id (opaque, no
+    # PII). It carries the reserved user_id tag; all other tags are dropped
+    # here and survive only on spans/metrics.
+    user_id = tags["user_id"]
+    params[:metadata] = {user_id: user_id} if user_id
```
bottrall, line 68: **Thought(non-blocking):** Considering anthropic strips everything - would it be more future proof to just pass everything through and let anthropic handle it as it sees fit?

lib/riffer/agent.rb
```diff
@@ -332,6 +339,17 @@ def to_json(*)
...
+  # Tags must be a flat hash; key/value coercion happens later in the run.
+  # Validated here so a bad type fails fast at the public boundary. A +nil+ is
+  # tolerated and collapses to an empty hash.
+  #--
+  #: (Hash[(String | Symbol), untyped]?) -> Hash[(String | Symbol), untyped]
+  def validate_tags(tags)
+    return tags if tags.is_a?(Hash)
+    return {} if tags.nil?
+    raise Riffer::ArgumentError, "tags: must be a Hash, got #{tags.class}"
+  end
```
bottrall: **Suggestion(non-blocking):** Move this into `normalize_tags` in `Riffer::Agent::Run`.



Tags are a run concern, so we should validate and normalize there.

bottrall (approved):

lib/riffer/providers/anthropic.rb
cloetzi: The gem would raise, they have it part of the schema.

PR #369, opened 2026-07-21
feat: record time to first chunk on streaming chat spans

bottrall:
## Problem

Consumers of riffer's tracing have no way to see time-to-first-token latency for streaming LLM calls — the `chat` span only exposes total duration, which conflates queueing/prefill latency with generation time. TTFT is the primary responsiveness metric for streaming UX, and observability backends can't derive it from the span data riffer emits today.

## Solution

Adds the OpenTelemetry GenAI semantic convention attribute `gen_ai.response.time_to_first_chunk` (a Float in seconds, per the convention's definition in `open-telemetry/semantic-conventions-genai`) to streaming `chat` spans. `Tracing::StreamRecorder` captures a monotonic timestamp at construction — immediately before the provider request is issued — and memoizes the elapsed time on the first stream event of any type, so tool-call-only and reasoning-only streams are measured too, matching the convention's "first chunk" (not "first text token") semantics. The value is stamped in `record_stream_outcome` alongside the other terminal attributes.

Two deliberate boundaries: the attribute is never set on the non-streaming `generate_text` path (the convention marks it streaming-only), and streams that produce no events omit the attribute rather than recording a nil or zero. The clock is injected as a constructor parameter (defaulting to `Process::CLOCK_MONOTONIC`) so tests are deterministic without sleeps.

Note the attribute is Development-stability in the semconv and postdates the schema version riffer pins (1.37.0); it lives in the newer dedicated GenAI conventions repo. `docs/16_TRACING.md` documents it in the `chat` span table.

## Proof

Covered by CI: new unit tests for `StreamRecorder` (first-event measurement, memoization across subsequent events, nil when no events arrive) and span-level tests in `base_test.rb` (streamed span gets a positive value, silent streams and `generate_text` omit the attribute). Full `bin/rake` (tests + StandardRB + Steep) is green locally.

## Summary by CodeRabbit

* **New Features**
  * Chat streaming traces now include the time from request start until the first streamed response chunk.
  * This metric is recorded only when streaming produces an event and is not added to non-streaming text generation spans.

* **Documentation**
  * Documented the new streaming-only OpenTelemetry attribute and its meaning.

### DIFF
--- a/lib/riffer/tracing/stream_recorder.rb
+++ b/lib/riffer/tracing/stream_recorder.rb
@@ -5,8 +5,12 @@
 # forwarding every event downstream untouched.
 class Riffer::Tracing::StreamRecorder # :nodoc: all
   # @rbs @yielder: Enumerator::Yielder
+  # @rbs @clock: ^() -> Float
+  # @rbs @started_at: Float
 
   attr_reader :token_usage #: Riffer::Providers::TokenUsage?
+
+  attr_reader :time_to_first_chunk #: Float?
 
   attr_reader :finish_reason #: Symbol?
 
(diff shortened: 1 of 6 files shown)


lib/riffer/tracing/stream_recorder.rb
```diff
@@ -17,15 +21,18 @@ class Riffer::Tracing::StreamRecorder # :nodoc: all
   attr_reader :tool_calls #: Array[Riffer::Messages::Assistant::ToolCall]
 
   #--
-  #: (Enumerator::Yielder) -> void
-  def initialize(yielder)
+  #: (Enumerator::Yielder, ?clock: ^() -> Float) -> void
+  def initialize(yielder, clock: -> { Process.clock_gettime(Process::CLOCK_MONOTONIC) })
```
cloetzi, line 25: question(non-blocking): Why do we need to inject a clock?

cloetzi (approved):

lib/riffer/tracing/stream_recorder.rb
bottrall: Short answer is It makes it easier to test. This way we don't need to stub anything.

Long answer is my personal `CLAUDE.md` has some parts which makes the agent always prefer dependency injection over alternatives, with test mocking as an example of why this is a preferable pattern. So it's always quick to refactor interfaces to DI instead of stubbing/mocking.

PR #371, opened 2026-07-24
chore: migrate from StandardRB to RuboCop

bottrall:
## Problem

StandardRB's fixed ruleset is not opinionated or strict enough for a codebase primarily written and maintained through agentic coding — agents benefit from tighter, machine-enforceable conventions (line limits, trailing commas, minitest cops) than standard exposes.

## Solution

Replaces standard with RuboCop plus the rubocop-minitest, rubocop-performance, and rubocop-rake plugins, configured in a single `.rubocop.yml` with cops sorted alphabetically. Riffer-specific exceptions carry a why-comment; the notable ones protect rbs-inline `#:` annotations from layout cops, shipped public API names (`has_tool_calls?`, `validate_as_tool!`) from renaming cops, and keyword arguments on block-based hooks (tool `execute` blocks, evaluator lambdas) from unused-argument underscoring that would rename the keyword and break callers.

Supporting changes: dev dependencies move from the gemspec to the Gemfile per `Gemspec/DevelopmentDependencies` (the gemspec keeps only zeitwerk, and gains `rubygems_mfa_required` — compatible with the trusted-publishing release flow); `bin/lint` is a plain rubocop passthrough (use `-a`/`-A` to autocorrect); the default rake task becomes `test rubocop steep:check`; docs updated accordingly.

The bulk of the diff is `rubocop --autocorrect` output. The judgment fixes were reviewed individually — the one behavioral near-miss worth a reviewer's eye is `AzureOpenAI#initialize`, which now resolves its credentials and delegates to the parent OpenAI initializer explicitly rather than relying on implicit construction; the resulting client is identical and covered by the provider test suite.

## Proof

CI covers it: the full suite (2,075 tests), `rubocop` (0 offenses across 218 files), and Steep all pass via the default rake task, and the lint lane runs `bin/lint`. Split-string byte-identity and AST-equivalence of the reformatted files were verified mechanically during review.


lib/riffer/providers/anthropic.rb
```diff
@@ -210,11 +213,21 @@ def execute_stream(params, yielder)
           handle_input_json_event(event, state: current_state, yielder: yielder)
         when ::Anthropic::Helpers::Streaming::ContentBlockStopEvent
           block = event.content_block
-          handle_content_block_stop_text(event, state: current_state, yielder: yielder) if block.is_a?(::Anthropic::Models::TextBlock) && current_state[:text]
-          handle_content_block_stop_tool_use(event, state: current_state, yielder: yielder) if block.is_a?(::Anthropic::Models::ToolUseBlock)
-          handle_content_block_stop_thinking(event, state: current_state, yielder: yielder) if block.is_a?(::Anthropic::Models::ThinkingBlock) && current_state[:reasoning]
-          handle_content_block_stop_server_tool_use(event, state: current_state, yielder: yielder) if block.is_a?(::Anthropic::Models::ServerToolUseBlock)
-          handle_content_block_stop_web_search_result(event, state: current_state, yielder: yielder) if block.is_a?(::Anthropic::Models::WebSearchToolResultBlock)
+          if block.is_a?(::Anthropic::Models::TextBlock) && current_state[:text]
```
cloetzi: suggestion(non-blocking): I think that should be a case statement, no?

test/riffer/providers/amazon_bedrock_test.rb
```diff
@@ -605,7 +605,7 @@ def build_response(content_blocks)
         VCR.use_cassette("Riffer_Providers_AmazonBedrock/_stream_text/when_prompt_is_provided/yields_TextDelta_events") do
           provider = Riffer::Providers::AmazonBedrock.new(api_token: api_token, region: "us-east-1")
           events = provider.stream_text(prompt: "Say hello", model: "us.anthropic.claude-haiku-4-5-20251001-v1:0").to_a
-          deltas = events.select { |e| e.is_a?(Riffer::StreamEvents::TextDelta) }
+          deltas = events.grep(Riffer::StreamEvents::TextDelta)
```
cloetzi, line 649: TIL!

lib/riffer/agent/response.rb
```diff
@@ -44,8 +44,18 @@ class Riffer::Agent::Response
 
   #--
   #: (String, ?tripwire: Riffer::Guardrails::Tripwire?, ?modifications: Array[Riffer::Guardrails::Modification], ?interrupted: bool, ?interrupt_reason: (String | Symbol)?, ?structured_output: Hash[Symbol, untyped]?, ?messages: Array[Riffer::Messages::Base], ?healed_tool_call_ids: Array[String], ?token_usage: Riffer::Providers::TokenUsage?, ?steps: Integer) -> void
-  def initialize(content, tripwire: nil, modifications: [], interrupted: false, interrupt_reason: nil,
-                 structured_output: nil, messages: [], healed_tool_call_ids: [], token_usage: nil, steps: 0)
+  def initialize(
```
cloetzi, line 58: nitpick(non-blocking): It's now misaligned with the type annotations, I think we should keep them consistent (best effort).

lib/riffer/params.rb
```diff
@@ -153,7 +153,8 @@ def build_nested(type, of, &block)
       raise Riffer::ArgumentError, "of: can only be used with Array type, got #{type}" unless type == Array
       unless Riffer::Params::Param::PRIMITIVE_TYPES.include?(of)
         raise Riffer::ArgumentError,
-              "of: must be a primitive type (#{Riffer::Params::Param::PRIMITIVE_TYPES.map(&:name).join(', ')}), got #{of}"
+              "of: must be a primitive type (#{Riffer::Params::Param::PRIMITIVE_TYPES.map(&:name).join(', ')}), " \
```
cloetzi: nitpick(non-blocking): As mentioned in Slack already, that's strictly worse. Not a blocker, but I am wondering if we should be fairly tolerant for rule exceptions for line length, or increase it slightly? The default is probably not based on modern wide-screen monitors.

test/riffer/agent/context_test.rb
```diff
@@ -17,7 +17,7 @@
     end
 
     it "exposes caller keys via #dig" do
-      expect(Riffer::Agent::Context.new(user_id: 42)[:user_id]).must_equal 42
+      expect(Riffer::Agent::Context.new(user: { id: 42 }).dig(:user, :id)).must_equal 42
```
cloetzi, line 20: question(non-blocking): Where is this coming from? That's sus.

test/riffer/providers/gemini_test.rb
```diff
@@ -770,7 +829,8 @@
 
   describe "file handling" do
     let(:image_base64) do
-      "iVBORw0KGgoAAAANSUhEUgAAADIAAAAyCAIAAACRXR/mAAAAQ0lEQVR4nO3OMQ0AMAwDsPAnvRHonxyWDMB5yaD+QEtLS0tLa0N/oKWlpaWltaE/0NLS0tLS2tAfaGlpaWlpbegPTh97K7rEaOcNTQAAAABJRU5ErkJggg=="
+      "iVBORw0KGgoAAAANSUhEUgAAADIAAAAyCAIAAACRXR/mAAAAQ0lEQVR4nO3OMQ0AMAwDsPAnvRHonxyWDMB5yaD+QEtLS0tLa0N/" \
```
cloetzi: nitpick(non-blocking): Another one of those strings that's probably best left alone.

cloetzi (approved):
Most of the stuff is strictly better, thanks for pushing this forward! My vote would be to increase the allowed line-length (ideally limited to strings only). Outside of strings, pretty much everything is folded by other rules already, so it shouldn't affect much else.

Also found some weird coding pattern and a sus test change, but shouldn't be blocking.

test/riffer/agent/context_test.rb
bottrall: `Style/SingleArgumentDig` changed it to a key access. But the test is "exposes caller keys via #dig". So we changed what is provided so we continue to test the dig method.

PR #378, opened 2026-07-29
perf: avoid O(n^2) text accumulation in streaming providers

cloetzi:
## Problem

Streamed text and tool-call JSON fragments were accumulated with `String#+=`. Each `+=` allocates a **new** string and copies the entire accumulated buffer, so assembling an `n`-token response costs O(n²) work and O(n²) transient allocations — incurred once per token, synchronously on the fiber reactor, even when client-side streaming is disabled.

## Fix

Switch the accumulators to in-place `String#<<`, seeding each buffer with an unfrozen `+""` so the append is legal under `# frozen_string_literal: true`.

Sites changed:
- **Bedrock** (`amazon_bedrock.rb`): streamed text + tool-call input JSON.
- **Anthropic** (`anthropic.rb`): streamed text, reasoning, tool-call input JSON, and web-search input JSON.
- **`Agent::Run#accumulate_streamed_response`**: the consumer-side text accumulator.

`OpenRouter` and `Gemini` already used `<<`, so this aligns the remaining providers with the existing repo idiom.

### Aliasing safety

Every mutated buffer is handed off to a done-event (`TextDone` / `ToolCallDone` / `ReasoningDone`) and then cleared before the next content block starts, so no reader ever observes a still-mutating string. In `accumulate_streamed_response`, `TextDone` now uses `String#replace` instead of reassignment; this keeps `accumulated_content` an owned buffer so a subsequent delta's `<<` can never mutate the string held by an already-emitted `TextDone` event. Behavior is byte-for-byte identical to the previous code.

## Tests

Adds a Bedrock unit test that streams 500 deltas through the provider and asserts the assembled `TextDone` content is byte-identical to the plain concatenation — guarding both the frozen-string seed regression (a frozen `""` would raise `FrozenError` on the first delta) and any mutation-aliasing. Full suite (`bin/rake`: test + StandardRB + Steep) passes.

## Impact

Micro-benchmark, 10k × 20-char deltas:

| approach | time |
| -------- | ---- |
| `+=`     | 27.2 ms |
| `<<`     | 0.3 ms |

The `+=` cost grows quadratically with response length; `<<` is linear.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming text and tool-call handling across supported providers.
  * Prevented previously emitted streaming content from being unintentionally modified.
  * Preserved streamed text and JSON content accurately during incremental updates.

* **Tests**
  * Added coverage verifying streamed text remains byte-identical and individual text fragments stay unchanged.

### DIFF
--- a/lib/riffer/agent/run.rb
+++ b/lib/riffer/agent/run.rb
@@ -121,7 +121,7 @@ module Riffer::Agent::Run
   #--
   #: (Riffer::Agent, Enumerator::Yielder, ?Hash[String, String]) -> Riffer::Messages::Assistant
   def accumulate_streamed_response(agent, stream_yielder, tags = {})
-    accumulated_content = ""
+    accumulated_content = +""
     accumulated_tool_calls = [] #: Array[Riffer::Messages::Assistant::ToolCall]
     accumulated_token_usage = nil #: Riffer::Providers::TokenUsage?
     accumulated_finish_reason = nil #: Symbol?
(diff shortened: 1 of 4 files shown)


lib/riffer/agent/run.rb
```diff
@@ -121,7 +121,7 @@ def execute_run(agent, stream_yielder, tags = {})
   #--
   #: (Riffer::Agent, Enumerator::Yielder, ?Hash[String, String]) -> Riffer::Messages::Assistant
   def accumulate_streamed_response(agent, stream_yielder, tags = {})
-    accumulated_content = ""
+    accumulated_content = +""
```
ChadMcCallum-Jane, line 124: TIL what `+""` does :D

ChadMcCallum-Jane (approved):

PR #391, opened 2026-08-11
feat!: config-injected provider clients, retire provider_options

cloetzi:
## Summary

Redesigns how providers obtain their SDK clients. Riffer works out of the box by building a default client from configured credentials; production deployments bring their own client.

- **Credential-only constructors.** Providers take explicit keywords — `api_key:` (plus `base_url:` on OpenAI for OpenAI-compatible servers, `endpoint:` on AzureOpenAI, `api_token:`/`region:` on Bedrock). The `**options` splat into SDK constructors is gone, so riffer's surface is no longer coupled to each SDK's constructor signature and typos fail fast.
- **Config-injected clients.** Each provider config struct gains a `client` member: `config.openai.client = OpenAI::Client.new(...)`. It accepts a client instance or a **no-argument** `Proc` resolved on **every** LLM call, covering fork safety and expiring credentials (Azure AD/STS). Explicit constructor credentials pin a provider-built default client and are never overridden by a shared configured client.
- **Providers hold no agent state.** A provider never holds a context or a reference to its owning agent — client selection is process-global by design. Agents construct providers bare and never touch them afterwards.
- **Credentials fall through to the SDK.** Default clients are built from compacted kwargs, so a credential left unset in riffer stays *absent* rather than being passed as `nil`, and each vendor SDK still applies its own resolution (`OPENAI_API_KEY`/`OPENAI_BASE_URL`, `ANTHROPIC_API_KEY`, the AWS region chain). `OpenRouter` and `AzureOpenAI` deliberately opt out: they borrow `OpenAI::Client` to reach a *different* vendor, so falling through would send an OpenAI credential to `openrouter.ai` or an Azure endpoint.
- **Riffer-owned Gemini client.** Gemini has no vendor SDK, so `Riffer::Providers::Gemini::Client` now owns the HTTP transport, exposing `base_url`, `open/read/write` timeouts, and proxy settings behind a duck-typed `post`/`post_stream` contract — any conforming object can be assigned to `config.gemini.client`.
- **`provider_options` retired.** The DSL macro, `Agent::Config#provider_options`, and `Evals::Judge`'s `provider_options:` are removed. The serializer no longer emits the field, so credentials can no longer ride the wire (the compromise deferred in #293).
- Custom providers inherit the same resolution via `Riffer::Providers::Base` hooks (`build_default_client`, `provider_config`); docs updated accordingly.

## Bugs fixed along the way

Passing an explicit `nil` credential is not equivalent to omitting it — SDKs use the distinction to decide whether to run their own resolution. Both of these were silent or misleading:

- `base_url: nil` suppressed the OpenAI SDK's `OPENAI_BASE_URL` lookup and pinned every request to `api.openai.com`. A deployment routing through a gateway (LiteLLM, vLLM, corporate egress) would silently start sending its gateway-scoped key to OpenAI, with no error.
- `region: nil` made Bedrock raise `Aws::Errors::MissingRegionError` **even with `AWS_REGION` exported**, because an explicit nil bypasses the AWS resolution chain.

## Breaking changes

- SDK option passthrough in provider constructors is removed; unknown constructor options now raise `Riffer::ArgumentError` (including Gemini, which previously ignored them silently).
- `Agent.provider_options`, `Agent::Config#provider_options`, and `Evals::Judge.new(provider_options:)` are removed.
- Gemini's `open_timeout`/`read_timeout` constructor kwargs and config members move to `Riffer::Providers::Gemini::Client`.
- `Riffer::Providers::Mock.new` takes `responses:` only.
- Provider credentials and clients are **process-global**. Agents do not take per-agent credentials, so per-agent or per-tenant client routing is not expressible; a process that must reach two accounts uses a provider directly instead of the agent layer.

## Serialization compatibility

`schema_version` stays 1. `to_h` no longer emits `provider_options`; `from_h` only reads named keys, so previously serialized definitions still decode and the receiving side's provider configuration applies.

## Test plan

- [x] `bin/rake` (test + rubocop + steep:check) green: 2128 runs, 3184 assertions, 0 failures/errors (2 pre-existing OTEL-conditional skips), no RuboCop offenses, no type errors
- [x] New `Gemini::Client` tests (knobs, error mapping, stream chunking) and per-provider client-resolution tests (configured instance, per-call Proc, constructor-credentials precedence, strict-kwargs raise)
- [x] SDK fall-through covered per provider: `OPENAI_API_KEY`/`OPENAI_BASE_URL`, `ANTHROPIC_API_KEY`, `AWS_REGION`, each verified to fail against the pre-fix code
- [x] Cross-vendor guards for `OpenRouter`/`AzureOpenAI` assert a stray `OPENAI_API_KEY` is never picked up — verified to fail if those providers are compacted
- [x] VCR cassettes untouched — request building is unchanged and all cassette-backed tests pass as-is

## Summary by CodeRabbit

- **New Features**
  - Added global provider configuration for credentials, endpoints, regions, and custom clients.
  - Added reusable clients and per-call client factories, including custom Gemini HTTP transport.
  - Providers now support configuration and supported environment-variable credential resolution.

- **Breaking Changes**
  - Removed agent-level `provider_options`; provider constructors no longer accept credential overrides.
  - Serialized agents retain model options only; provider credentials and clients are excluded.

- **Documentation**
  - Expanded configuration and provider setup guidance for the updated client model.
  - Clarified custom provider and client configuration requirements.


docs/providers/CUSTOM_PROVIDERS.md
```diff
@@ -121,16 +121,30 @@ class Riffer::Providers::MyProvider < Riffer::Providers::Base
 end
 ```
 
+## Client resolution
+
+Providers take **no constructor arguments** — agents construct them bare (`provider_class.new`), and every credential is read from configuration inside `build_client`. `Riffer::Providers::Base` provides a private `client` method your `execute_generate`/`execute_stream` should call instead of holding a client ivar. It resolves, in order:
```
bottrall: suggestion(non-blocking): Remove this. It's stating how things changed in this PR and doesn't make sense reading it without the context.

docs/providers/PROVIDERS.md
```diff
@@ -30,6 +30,21 @@ class MyAgent < Riffer::Agent
 end
 ```
 
+## Credentials and Clients
+
+Provider constructors take **no arguments** — every credential comes from configuration, so `Riffer::Providers::OpenAI.new` is the only valid form. Passing anything raises `ArgumentError`.
```
bottrall: suggestion(non-blocking): As above remove this as it is stating how things changed.

docs/CONFIGURATION.md
```diff
@@ -305,17 +355,51 @@ end
 
 ## Multiple Configurations
 
-For different environments or use cases, use agent-level overrides:
+Provider credentials and clients are **global**, resolved per process rather than per agent. For different environments, branch at boot:
```
bottrall: suggestion(non-blocking): Remove "rather than per agent". As a above this is referencing how things have changed as opposed to keeping the docs to how things currently are.

docs/CONFIGURATION.md
```diff
@@ -305,17 +355,51 @@ end
...
+Riffer.configure do |config|
+  config.openai.client = if Rails.env.production?
+    OpenAI::Client.new(api_key: ENV['PRODUCTION_OPENAI_KEY'], max_retries: 4)
+  else
+    OpenAI::Client.new(api_key: ENV['DEV_OPENAI_KEY'], timeout: 10)
+  end
+end
+```
+
+Agents do not take per-agent credentials, and neither do providers — `Riffer::Providers::OpenAI.new` takes no arguments and reads everything from `config.openai`. When a single process genuinely has to reach two different accounts or endpoints, give the second one its own provider class and config, then register it under its own identifier:
```
bottrall: suggestion(non-blocking): as above

docs/SERIALIZATION.md
```diff
@@ -101,7 +100,9 @@ A finite integer round-trips as-is; a hash missing the key falls back to the def
 
 ## Secrets
 
-`provider_options` and `model_options` **ride on the wire as plain data** — they are part of the hash and _will_ transfer. Prefer configuring API keys via environment/global provider configuration rather than `provider_options`. **Never serialize an agent whose options carry sensitive values** — and if a serialized definition ever does, handle it as a secret (encrypt it, keep it out of logs).
+`model_options` **rides on the wire as plain data** — it is part of the hash and _will_ transfer. Keep sensitive values out of it; credentials live in global provider configuration and never serialize.
+
+That is also where the hash stops: it describes the agent, not how to authenticate as one. The destination process has to configure the provider itself — `config.<provider>.client`, or the credentials riffer builds a default client from — before the rebuilt agent can make a call. Clients resolve lazily on the first LLM call, so `from_h` succeeds in a process with no provider configuration at all; the `generate` is what fails. See [Configuration](CONFIGURATION.md).
```
bottrall: suggestion(non-blocking): as above.

lib/riffer/providers/base.rb
```diff
@@ -105,6 +106,33 @@ def depends_on(gem_name)
...
+    @client ||= build_client
+  end
+
+  # Returns the provider's global config struct (the +client+ / credential
+  # holder); nil for providers without one.
+  #--
+  #: () -> untyped
+  def provider_config
+    nil
+  end
```
bottrall: suggestion(non-blocking): `provider_config` doesn't seem right. Maybe `global_client`?

lib/riffer/config.rb
```diff
@@ -3,12 +3,16 @@
...
-  AmazonBedrock = Struct.new(:api_token, :region)
-  Anthropic = Struct.new(:api_key)
-  AzureOpenAI = Struct.new(:api_key, :endpoint)
-  Gemini = Struct.new(:api_key, :open_timeout, :read_timeout)
-  OpenAI = Struct.new(:api_key)
-  OpenRouter = Struct.new(:api_key)
+  # Each provider struct's +client+ member optionally carries a consumer-owned
+  # client replacing the default one riffer builds from the credential members —
+  # either a client instance or a no-argument Proc returning one, resolved on
+  # every LLM call.
```
bottrall: suggestion(non-blocking): Remove this comment. I think it's stating the obvious and therefore doesn't meet the bar.

test/riffer/agent/serializer_test.rb
```diff
@@ -45,17 +45,21 @@ def build_agent_class(&block)
...
 
-      expect(dict[:provider_options]).must_equal({ foo: "bar" })
       expect(dict[:model_options]).must_equal({ temperature: 0.2 })
     end
 
+    it "carries no provider options" do
+      dict = build_agent_class.new.to_h
+
+      expect(dict.key?(:provider_options)).must_equal false
+    end
```
bottrall: suggestion(non-blocking): Remove this test. Without the context of this PR its a bit odd.

test/riffer/agent/serializer_test.rb
```diff
@@ -134,11 +138,11 @@ def build_agent_class(&block)
...
-    it "restores provider and model options" do
-      klass = build_agent_class { provider_options foo: "bar" }
-      agent = Riffer::Agent.from_h(klass.new.to_h, context: nil)
+    it "ignores a legacy provider_options key" do
+      dict = build_agent_class.new.to_h.merge(provider_options: { api_key: "legacy" })
+      agent = Riffer::Agent.from_h(dict, context: nil)
 
-      expect(agent.config.provider_options).must_equal({ foo: "bar" })
+      expect(agent).must_be_instance_of Riffer::Agent
     end
```
bottrall: suggestion(non-blocking): Same comment as above.

bottrall (commented):
Thanks @cloetzi. Everything is optional except for retaining the provider client types.

bottrall (approved):

bottrall (approved):

PR #406, opened 2026-08-21
feat!: one identifier registry pattern for tools and agents

bottrall:
## Problem

Riffer had three unrelated identity mechanisms and no shared lookup story. `Riffer::Tool` had no `.all`/`.find` at all, so host applications hand-rolled their own subclass registries (typically a linear scan over `subclasses`). `Riffer::Agent.find` was an O(n) scan that re-derived every identifier per comparison, and identifier derivation (three regex passes) was recomputed on every call across tools, agents, and guardrails — on hot paths like tool dispatch, span naming, and guardrail execution. In a host with a large tool set this compounds to O(n²) derivations per request: profiling one such host showed a significant share of request wall time in tool lookup, dominated by regex/string churn, and a spike there measured a 166-identifier resolution at 31.5 ms versus 0.013 ms with a memoized registry (~2,400×). Duplicate identifiers — an ambiguous LLM tool call — silently resolved to the first match.

## Solution

One shared registry mixin, `Riffer::Registrable`, extended by `Riffer::Tool` (new `.find`/`.all`) and `Riffer::Agent` (replaces the scan): a per-class, lazily built, frozen identifier→class hash over **named direct subclasses**, busted in `inherited` (self-scoped, which Ruby's hook semantics make exactly sufficient — and loading-model-agnostic across eager load, Zeitwerk lazy load, and test-defined classes). Anonymous classes never register, even with explicit identifiers — this keeps the MCP tool factory's and serializer's per-request shells out of the registry and immune to GC-timing duplicate flakes. Colliding identifiers raise the new `Riffer::DuplicateIdentifierError` inside the build, so the memo never caches a partial hash and the error re-raises deterministically.

`Riffer::Helpers::ClassNameConverter` is consolidated into `Riffer::Helpers::Identifier`: `derive` keeps the exact conversion; `for(klass)` memoizes per class (never while `Module#name` is nil, so late-named classes still derive correctly) and is shared by `Toolable#identifier`, `Agent.identifier`, and `Guardrail#name`. `sig/manual` narrows `find`/`all` back to `singleton(Riffer::Agent)`/`singleton(Riffer::Tool)` return types so typed consumers keep `Agent.find(id)&.generate(...)`.

Subtleties for review: the registry is direct-children-only by design (lineages like a host's `Context < Tool` stay separate; grandchildren are looked up on their intermediate base — documented loudly in TOOLS.md/AGENTS.md); the tripwire span attribute keeps its documented named-guardrails-only contract via an empty-identifier guard; `Providers::Repository` and guardrail declaration are deliberately untouched (explicit registration and class-reference semantics are different shapes, and neither has a find-by-identifier need). Breaking: duplicate identifiers now raise, anonymous subclasses drop out of `Agent.all`/`.find`, and `ClassNameConverter` is removed (its `separator:` kwarg had no callers).

## Proof

`bin/rake` green locally: 2150 runs, 3221 assertions, 0 failures, 0 errors (2 pre-existing skips); RuboCop 224 files, no offenses; Steep, no type errors. Registry semantics are covered by new gem specs (derived/explicit/symbol lookup, miss→nil, late-defined subclass invalidation, anonymous exclusion, grandchild isolation, duplicate raise incl. repeat-lookup re-raise) plus memoization specs for the identifier helper. Empirical checks during review confirmed one derivation per class across dispatch/span/guardrail paths, one registry build per lookup burst, warm `find` at ~0.1 µs/op flat from n=2000 to n=5000, and that host-lineage registries are unaffected by anonymous-shell churn. Host-side before/after benchmarks will accompany the downstream gem bump.

## Summary by CodeRabbit

- **New Features**
  - Added lookup and enumeration for named agent and tool subclasses.
  - Added automatic identifier generation, including support for explicit identifiers.
  - Added duplicate-identifier error handling.
  - Anonymous subclasses are excluded from registries.

- **Documentation**
  - Expanded guidance for agent and tool lookup behavior, inheritance, anonymous classes, and duplicate identifiers.

- **Refactor**
  - Consolidated identifier generation for consistent naming across agents, tools, guardrails, and tracing.


docs/AGENTS.md
```diff
@@ -306,6 +306,27 @@ agent.generate('Hello')
 
 When `config:` is supplied, the class-level configuration is ignored for that instance.
 
+## Looking Up Agents
+
+`Riffer::Agent.find` and `Riffer::Agent.all` are backed by a per-class registry keyed by identifier, so lookups are O(1) regardless of how many agents are defined:
```
bottrall: Think about the audience of these docs. The consumer of the library doesn't need to know the implementation details, just how to use it.

docs/TOOLS.md
```diff
@@ -143,6 +143,23 @@ class CreateOrderTool < Riffer::Tool
 end
 ```
 
+## Looking Up Tools
+
+`Riffer::Tool.find` and `Riffer::Tool.all` are backed by a per-class registry keyed by identifier, so lookups are O(1) regardless of how many tools are defined:
```
bottrall: As above, ensure the docs contain info relevant to the target audience.

lib/riffer/providers/base.rb
```diff
@@ -32,7 +32,7 @@ def self.semconv_provider_name
     class_name = name
     return "unknown" unless class_name
 
-    Riffer::Helpers::ClassNameConverter.convert(class_name.split("::").last.to_s)
+    Riffer::Helpers::Identifier.derive(class_name.split("::").last)
```
bottrall: We should memoise this as well.

lib/riffer/helpers/identifier.rb
```diff
@@ -0,0 +1,40 @@
...
+    # Tool classes shadow Module#name with the identifier DSL, so the real
+    # class-path name must come from Module's own implementation.
+    real_name = Module.instance_method(:name).bind_call(klass) #: String?
+    return "" if real_name.nil?
+
+    derived = derive(real_name)
+    klass.instance_variable_set(:@derived_identifier, derived)
+    derived
+  end
+end
```
bottrall, line 41: What's the benefit of caching here? Why not cache in the consuming class? e.g.
```rb
  # Gets or sets the agent identifier.
  #
  #--
  #: (?String?) -> String
  def self.identifier(value = nil)
    value.nil? ? (config.identifier ||= Riffer::Helpers::Identifier.for(self)) : (config.identifier = value)
  end
```

docs/AGENTS.md
bottrall: Rewritten for the consumer audience in e082fc10 — the section now leads with how to use `find`/`all` and keeps only the user-facing contracts (named direct subclasses, anonymous classes never found, duplicates raise); the registry/memoization mechanics are dropped.

— 🤖 Claude (beep boop)

docs/TOOLS.md
bottrall: Same treatment as the AGENTS.md section in e082fc10, kept structurally parallel with it.

— 🤖 Claude (beep boop)

lib/riffer/providers/base.rb
bottrall: Memoized in e082fc10 into a per-subclass class ivar. The anonymous branch deliberately stays uncached — a class assigned to a constant later must pick up its real name rather than a frozen "unknown" (same guard principle as `Identifier.for`). Shipped providers override this with literals, so only custom providers hit the derive path at all.

— 🤖 Claude (beep boop)

lib/riffer/helpers/identifier.rb
bottrall: Resolved offline: keeping the helper-side cache. Summary of the reasoning — `config.identifier ||=` would permanently latch `""` for anonymous classes (`""` is truthy), breaking the anonymous-then-named case unless every consumer reimplements the nil-name guard; it would also make a derived identifier indistinguishable from an explicitly-set one in config, and Guardrail#name has no config slot to cache into. The guard lives with the cache, implemented once. The comment on `Identifier.for` now states this rationale (e082fc10).

— 🤖 Claude (beep boop)

lib/riffer/registrable.rb
```diff
@@ -0,0 +1,81 @@
...
+  # Returns all registered subclasses. Only *named direct* subclasses are
+  # registered: grandchildren are not included (call +all+ on their direct
+  # parent instead), anonymous classes are never registered, and duplicate
+  # identifiers raise Riffer::DuplicateIdentifierError at first lookup.
+  #
+  #--
+  #: () -> Array[Class]
+  def all
+    identifier_registry.values
+  end
```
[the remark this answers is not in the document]
bottrall: Intentional: the frozen, memoized structure is the internal registry hash — `all` never claims to return a frozen array. Handing out a fresh array per call is the same contract `Class#subclasses` had (which `Agent.all` returned before this PR), and it's safer than a shared memoized frozen array: callers can sort/filter their copy in place, and nothing they do can corrupt the registry. The cost is one O(n) `Hash#values` copy on a rarely-called reflection method, so we're keeping it as is.

— 🤖 Claude (beep boop)

lib/riffer/tools/toolable.rb
```diff
@@ -57,9 +58,7 @@ def description(value = nil)
   #--
   #: (?String?) -> String
   def identifier(value = nil)
-    if value.nil?
-      return @identifier || Riffer::Helpers::ClassNameConverter.convert(Module.instance_method(:name).bind_call(self))
-    end
+    return @identifier || Riffer::Helpers::Identifier.for(self) if value.nil?
```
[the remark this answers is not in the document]
bottrall: The behavior is real but deliberately out of contract: `identifier` is a class-body declaration, not runtime state. For the supported pattern the invalidation is already correct — `inherited` busts the parent's registry memo the moment a subclass is created, and the class body's `identifier "x"` executes before any subsequent `find` rebuilds, so a class-body identifier is always registered under its final name. A stale entry requires re-identifying an already-registered class at runtime after lookups began, which nothing in the gem (or its known hosts) does, and which the previous code only half-supported anyway (`Agent.find`'s live scan reflected it; `Tool.find` did not exist). If runtime re-identification becomes a use case we want to support, the fix is setter-side invalidation of the direct parent's registry across the four assignment paths plus regression coverage — happy to do that as a follow-up, but we're not taking it in this PR.

— 🤖 Claude (beep boop)

ChadMcCallum-Jane (approved):

PR #407, opened 2026-08-21
feat: file attachment downloading for user messages

ChadMcCallum-Jane:
## Problem

`Riffer::Messages::User` supports file attachments (`FilePart`), but there's no policy governing how they reach a provider: no way to verify a URL-sourced file's integrity, no control over whether riffer is allowed to download an attachment at all, and a provider that can't accept a given attachment format only fails deep inside its own wire-format conversion instead of failing predictably up front.

## Solution

Adds `FilePart#sha256` for integrity verification, and `Riffer.config.files` — a download policy off by default (`allow_downloads`), with size/timeout caps, a per-message file count cap, a pluggable runner for concurrent downloads, and a pluggable downloader. `Riffer::Files::Resolver` runs before every `generate_text`/`stream_text` call: per file, it asks the provider (via a new `file_delivery` hook) whether it can take a URL as-is, needs the bytes inline, or can't accept the attachment at all — downloading, and verifying `sha256` if given, only when required and only when downloads are allowed. `max_per_message` is checked before consecutive messages are merged, against each message as the caller actually authored it. Bedrock's and Gemini's own unsupported-file-source guards are removed, since the resolver already raises earlier for anything it can't deliver.

## Proof

`bin/rake` passes, aside from pre-existing environment-flaky `amazon_bedrock_test.rb` VCR/encoding failures already present on `main`. New unit coverage for `Riffer::Files::Resolver` and `Riffer::Files::Downloader`, including download concurrency under `Threaded` vs `Sequential`, plus coverage in `Providers::Base` proving `generate_text`/`stream_text` actually invoke resolution (not just the resolver in isolation) and that `max_per_message` is checked pre-merge. Manually verified against the real network and a real Anthropic call: a real HTTPS download, `sha256` verification both succeeding and correctly raising on a deliberate mismatch, and a full round-trip PDF attachment with an accurate model response describing its actual content.

## Summary by CodeRabbit

- **New Features**
  - Added configurable file-attachment downloads with size, timeout, concurrency, and per-message limits.
  - Added HTTPS downloads, redirect handling, caching, custom downloaders, and optional SHA-256 verification.
  - Added provider-specific file delivery using URLs, inline Base64, or raw bytes.
  - Added URL-based file support for Gemini and expanded Amazon Bedrock file handling.
  - Added clear errors for unsupported, disabled, oversized, failed, invalid, or checksum-mismatched files.
- **Documentation**
  - Documented download configuration, checksums, URL attachments, custom downloaders, and provider behavior.
- **Tests**
  - Added comprehensive coverage for downloads, resolution, validation, limits, caching, and provider behavior.


docs/CONFIGURATION.md
```diff
@@ -182,6 +182,38 @@ end
...
+| `timeout`         | Open and read timeout, in seconds, for a single download attempt. Defaults to `60`.                                                                                                                               |
+| `max_per_message` | Maximum number of files allowed on a single user message; checked against each message as originally authored, before consecutive messages are merged. `nil` (default) means uncapped.                          |
+| `runner`          | A `Riffer::Runner` instance that resolves every file across a call's messages. Defaults to `Riffer::Runner::Sequential.new`; assign `Riffer::Runner::Threaded.new` (or `Riffer::Runner::Fibers.new` inside a fiber-based host) to resolve multiple files concurrently. |
+| `downloader`      | The object that fetches a URL's bytes; must respond to `#call(url, max_bytes:, timeout:)` returning base64-encoded content. Defaults to `Riffer::Files::Downloader.new`, which fetches over HTTPS only, following up to 3 redirects. Assign your own to add logging/metrics, or to fetch from a non-HTTPS store (e.g. `s3://`).                          |
```
cloetzi: praise(non-blocking): Nice!!

lib/riffer/files/downloader.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+# rbs_inline: enabled
+
+require "net/http"
+require "uri"
+require "base64"
```
cloetzi: suggestion(non-blocking): If it isn't already, we should add it as a dependency to the gemspec. It isn't loaded by default in all Ruby version I believe (I just went through that with maestro). No version constraint required.

lib/riffer/providers/amazon_bedrock.rb
```diff
@@ -413,13 +419,8 @@ def append_tool_result(conversation_messages, message)
...
 
-    source = if file.data
-               { bytes: Base64.decode64(file.data) }
-             elsif file.url&.start_with?("s3://")
-               { s3_location: { uri: file.url } }
-             else
-               raise Riffer::ArgumentError, "Amazon Bedrock only supports S3 URI or base64 data file sources"
-             end
+    data = file.data
+    source = data ? { bytes: Base64.decode64(data) } : { s3_location: { uri: file.url } }
```
cloetzi: question(non-blocking): Is bedrock the only provider that takes raw bytes? Are all others supporting URLs? 

I noticed that we transform the downloaded bytes to base64, then convert them back to bytes here. I am wondering if we could minimize the number of conversions by having a `data_bytes` method or something to minimize the number of conversions between base64 and bytes? With e.g. a 1 Mb file, that's not an insignificant overhead - especially in the context of fibers.

lib/riffer/files/downloader.rb
```diff
@@ -0,0 +1,79 @@
...
+
+class Riffer::Files::Downloader
+  MAX_REDIRECTS = 3 #: Integer
+
+  #: (String, max_bytes: Integer, timeout: Integer) -> String
+  def call(url, max_bytes:, timeout:)
+    Base64.strict_encode64(fetch(url, max_bytes: max_bytes, timeout: timeout, redirects_remaining: MAX_REDIRECTS))
```
cloetzi: question(non-blocking): I hinted at it in another comment, do we need the conversion to base64 here?

lib/riffer/files/resolver.rb
```diff
@@ -0,0 +1,85 @@
...
+  # +cache:+ is false for a :url-delivery provider verifying a sha256 — the
+  # request still sends the URL, never the downloaded bytes, so caching them
+  # would hold memory nothing reads and let later turns skip re-verifying.
+  #: (Riffer::Messages::FilePart, cache: bool) -> void
+  def download!(file, cache:)
+    raise Riffer::FileDownloadsDisabledError, "File attachments are disabled" unless @config.allow_downloads
+
+    encoded = @config.downloader.call(file.url, max_bytes: @config.max_bytes, timeout: @config.timeout)
+    verify_encoded!(encoded, file.sha256) if file.sha256
+    file.cache_downloaded_data(encoded) if cache
```
cloetzi: Finding from my robot reviewer:

security / broken integrity guarantee — Base64.decode64 silently truncates at the first = padding, so the sha256 is verified against a prefix of the payload, not the payload that gets transmitted.

  Confirmed by running it against this repo's Ruby:

```
smuggled = Base64.strict_encode64("hello world") + Base64.strict_encode64("EVIL PAYLOAD")
# => "aGVsbG8gd29ybGQ=RVZJTCBQQVlMT0FE"
Base64.decode64(smuggled)  # => "hello world"   (trailing block discarded)
```

A caller passing data: that string with sha256: Digest::SHA256.hexdigest("hello world") passes verification. Riffer then sends file.data verbatim — open_ai.rb:401,404 interpolate it into a data: URI, anthropic.rb:437 and
gemini.rb:283 put it straight in the base64 data field — so the provider's decoder sees bytes riffer never hashed. Only Bedrock re-decodes with the same lenient decode64.

This affects the caller-supplied inline path (verify_inline!, line 60). The download path is safe because Downloader#call re-encodes with strict_encode64, so it round-trips exactly.

Suggested fix — strict_decode64 after stripping whitespace, so Base64.encode64's 60-char line wrapping still works and malformed input surfaces as a FileError rather than a bare ArgumentError:

```
  def verify_encoded!(encoded, sha256)
    decoded = begin
      Base64.strict_decode64(encoded.gsub(/\s/, ""))
    rescue ArgumentError
      raise Riffer::FileChecksumMismatchError, "File checksum mismatch"
    end
    return if Digest::SHA256.hexdigest(decoded) == sha256

    raise Riffer::FileChecksumMismatchError, "File checksum mismatch"
  end
```

I verified both halves of that fix: strict_decode64 on Base64.encode64 output raises invalid base64, and strict_decode64(x.gsub(/\s/, "")) round-trips correctly.

cloetzi (commented):
Overall looks good, and I think it _would_ be acceptable for releasing. That said, I think we can get rid of quite a few base64 encodes/decodes which do matter in the context of fiber based applications. I think it is worth doing another pass to keep the downloaded data in raw bytes by default and only convert to base64 when needed.

My Claude reviewer also flagged one potential security issue.

lib/riffer/providers/amazon_bedrock.rb
ChadMcCallum-Jane: yeah, this is only ever a problem with bedrock client - I'll poke at it and see if we can come up with something better

ChadMcCallum-Jane: Reworked this quite a bit.  

- Now a provider can either accept :url, :base64, :bytes, or :none
- The downloder (when used) only ever downloads raw bytes
- The resolver uses the raw bytes to perform sha256 checks as needed
- The resolver only encodes to base64 only if the provider needs :base64

So for bedrock client, we'll never encode to base64

lib/riffer/files/resolver.rb
ChadMcCallum-Jane: Applied the fix, thanks for the catch!

lib/riffer/files/resolver.rb
```diff
@@ -0,0 +1,90 @@
...
+      raise Riffer::FileUnsupportedError,
+            "Provider does not support user message file attachments"
+    end
+    return verify_inline!(file) if file.inline_data?
+
+    case delivery
+    when :url
+      download!(file, cache: false) if file.sha256
+    when :data
+      file.data ? verify_inline!(file) : download!(file, cache: true)
```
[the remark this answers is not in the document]
ChadMcCallum-Jane: This is already addressed in documentation.  I don't think we want to start dealing with s3 access issues in Riffer that the bedrock client itself may not have

cloetzi (approved):

PR #412, opened 2026-08-26
feat!: tools are the error boundary — call_with_validation never raises

bottrall:
## Problem

A tool that raised anything outside the narrow set the runtime rescued (`Riffer::ValidationError`, `Riffer::TimeoutError`, `Riffer::ToolExecutionError`, `RuntimeError`) crashed the entire agent run — a `NoMethodError` in one tool killed the batch and discarded every sibling tool call's result, and malformed argument JSON from a provider (`JSON::ParserError`) did the same. The handling also lived inside `Runtime#dispatch_tool_call`, so any custom runtime overriding that method silently lost all error handling.

The underlying framing: a tool call is a sub-program, not a method call. Like a Rails controller action, its failures should become an error *response* at the framework wrapper, not an exception that takes down the host loop.

## Solution

The error boundary moves from the runtime into the tool. `Riffer::Tool#call_with_validation` now never raises: validation failures → `:validation_error`, timeouts → `:timeout_error`, `Riffer::ToolExecutionError` → `:execution_error`, and any other `StandardError` → a new `:unhandled_error` type with `Error executing tool: <Class>: <message>` content. Because the rescue travels with the tool, every dispatch path (inline, or a custom HTTP/gRPC runtime's server side) gets identical semantics for free.

Subtleties a reviewer might miss:

- The unhandled exception rides on the Response (`Response#exception`, deliberately excluded from `to_h` so it never reaches the LLM); the runtime uses it to call `span.record_exception` and set error status, preserving the loud telemetry the old raise gave us. Deliberate error Responses keep an unset span status — the 4xx/5xx distinction survives in traces.
- `NotImplementedError` still propagates: an unimplemented `#call` is a broken deploy, not a bad request (it's a `ScriptError`, deliberately outside `StandardError`).
- The runtime keeps only pre-tool concerns: unknown tools, and malformed argument JSON, which now returns `:validation_error` instead of crashing the run.
- `Riffer::TimeoutError` is now the sentinel `Timeout.timeout` raises, and only it maps to `:timeout_error` — an external library timeout inside a tool (`Net::ReadTimeout < Timeout::Error`) lands in `:unhandled_error` instead of masquerading as the tool's own budget. It's raised *inside* `call`, so tools can rescue it for cleanup. The validation rescue is likewise scoped to param validation only.
- **Breaking (telemetry):** on an unhandled tool failure, the `execute_tool` span's `error.type` is now always `unhandled_error` rather than the exception class name; the class still appears as `exception.type` on the recorded exception event. Dashboards or alerts grouping on exception class names in `error.type` need updating.

Docs updated: `TOOLS.md`, `TOOL_ADVANCED.md`, `AGENT_LOOP.md`, `TRACING.md`.

## Proof

CI covers it: `bin/rake` (2163 runs, 3246 assertions, 0 failures; RuboCop clean; Steep clean). New tests cover the full rescue table at the tool layer, the folded non-Response return, `NotImplementedError` still raising, malformed-JSON arguments at the runtime layer, and span assertions that `:unhandled_error` records the exception with error status while host-code raises still propagate.

## Summary by CodeRabbit

* **New Features**
  * Tool failures now return structured responses for validation, timeout, execution, and unhandled errors.
  * Unhandled exceptions are preserved for diagnostics but excluded from serialized responses.
  * Tool tracing records unhandled exceptions and marks affected operations as errors.

* **Bug Fixes**
  * Agent runs continue after tool errors, allowing the LLM to handle returned error information.
  * Malformed or non-object tool-call JSON is reported as a validation error.

* **Documentation**
  * Expanded guidance for tool errors, timeouts, malformed input, custom runtimes, and tracing behavior.

### DIFF
--- a/lib/riffer/tool.rb
+++ b/lib/riffer/tool.rb
@@ -56,11 +56,10 @@ class Riffer::Tool
     Riffer::Tools::Response.error(message, type: type)
   end
 
-  # Executes the tool with validation and timeout (used by Agent).
-  #
-  # Raises Riffer::ValidationError if validation fails.
-  # Raises Riffer::TimeoutError if execution exceeds the configured timeout.
-  # Raises Riffer::Error if the tool does not return a Response object.
+  # Executes the tool with validation and timeout, folding every +StandardError+
+  # into an error Response. Anything outside +StandardError+ — an unimplemented
+  # +#call+ above all — still propagates, because a broken tool is a broken
+  # deploy rather than a bad request.
   #
   #--
   #: (context: Riffer::Agent::Context?, **untyped) -> Riffer::Tools::Response
@@ -77,7 +76,20 @@ class Riffer::Tool
     end
 
     result
+  rescue Riffer::ValidationError => e
+    Riffer::Tools::Response.error(e.message, type: :validation_error)
   rescue Timeout::Error
-    raise Riffer::TimeoutError, "Tool execution timed out after #{self.class.timeout} seconds"
+    Riffer::Tools::Response.error(
+      "Tool execution timed out after #{self.class.timeout} seconds",
+      type: :timeout_error,
+    )
+  rescue Riffer::ToolExecutionError => e
+    Riffer::Tools::Response.error(e.message, type: :execution_error)
+  rescue StandardError => e
+    Riffer::Tools::Response.error(
+      "Error executing tool: #{e.class}: #{e.message}",
+      type: :unhandled_error,
+      exception: e,
+    )
   end
 end
--- a/lib/riffer/tool.rb
+++ b/lib/riffer/tool.rb
@@ -65,9 +65,14 @@ class Riffer::Tool
   #: (context: Riffer::Agent::Context?, **untyped) -> Riffer::Tools::Response
   def call_with_validation(context:, **kwargs)
     params_builder = self.class.params
-    validated_args = params_builder ? params_builder.validate(kwargs) : kwargs
 
-    result = Timeout.timeout(self.class.timeout) do
+    begin
+      validated_args = params_builder ? params_builder.validate(kwargs) : kwargs
+    rescue Riffer::ValidationError => e
+      return Riffer::Tools::Response.error(e.message, type: :validation_error)
+    end
+
+    result = Timeout.timeout(self.class.timeout, Riffer::TimeoutError) do
       call(context: context, **validated_args) #: untyped
     end
 
@@ -76,9 +81,7 @@ class Riffer::Tool
     end
 
     result
-  rescue Riffer::ValidationError => e
-    Riffer::Tools::Response.error(e.message, type: :validation_error)
-  rescue Timeout::Error
+  rescue Riffer::TimeoutError
     Riffer::Tools::Response.error(
       "Tool execution timed out after #{self.class.timeout} seconds",
       type: :timeout_error,
--- a/lib/riffer/tools/runtime.rb
+++ b/lib/riffer/tools/runtime.rb
@@ -89,14 +89,8 @@ class Riffer::Tools::Runtime
     arguments = parse_arguments(tool_call.arguments)
 
     tool_instance.call_with_validation(context: context, **arguments)
-  rescue Riffer::TimeoutError => e
-    Riffer::Tools::Response.error(e.message, type: :timeout_error)
-  rescue Riffer::ValidationError => e
-    Riffer::Tools::Response.error(e.message, type: :validation_error)
-  rescue Riffer::ToolExecutionError => e
-    Riffer::Tools::Response.error(e.message, type: :execution_error)
-  rescue RuntimeError => e
-    Riffer::Tools::Response.error("Error executing tool: #{e.message}", type: :execution_error)
+  rescue JSON::ParserError => e
+    Riffer::Tools::Response.error("Invalid JSON in tool arguments: #{e.message}", type: :validation_error)
   end
 
   #--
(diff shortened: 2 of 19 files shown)


lib/riffer/tool.rb
```diff
@@ -77,7 +76,20 @@ def call_with_validation(context:, **kwargs)
...
-    raise Riffer::TimeoutError, "Tool execution timed out after #{self.class.timeout} seconds"
+    Riffer::Tools::Response.error(
+      "Tool execution timed out after #{self.class.timeout} seconds",
+      type: :timeout_error,
+    )
+  rescue Riffer::ToolExecutionError => e
+    Riffer::Tools::Response.error(e.message, type: :execution_error)
+  rescue StandardError => e
+    Riffer::Tools::Response.error(
+      "Error executing tool: #{e.class}: #{e.message}",
```
tsov, line 93: question(non-blocking): Could this leak application data? `e.message` gets sent to the LLM, so if a tool raises with sensitive info in the message (ids, SQL, URLs), it goes out. What about sending only `e.class` and making the message opt-in?

lib/riffer/tool.rb
```diff
@@ -77,7 +76,20 @@ def call_with_validation(context:, **kwargs)
     end
 
     result
+  rescue Riffer::ValidationError => e
+    Riffer::Tools::Response.error(e.message, type: :validation_error)
   rescue Timeout::Error
```
tsov: suggestion(non-blocking): This rescue also catches `Net::ReadTimeout` (it subclasses `Timeout::Error`), so an HTTP timeout inside a tool would be labeled as the tool hitting its own timeout. Could we pass a sentinel instead, `Timeout.timeout(self.class.timeout, Riffer::TimeoutError)`, and rescue exactly that? Would also keep the constant around, so no breaking change.

lib/riffer/tools/runtime.rb
```diff
@@ -144,13 +145,21 @@ def tag_attributes(tags)
...
-  # A returned error Response is a handled outcome, so its status stays unset —
-  # an error span status is reserved for a raised exception.
+  # A deliberate error Response is a handled outcome, so its status stays unset.
+  # An error status is reserved for a Response carrying the exception it was
+  # folded from — the tool failed for a reason nobody anticipated.
   #--
   #: ((Riffer::Tracing::Otel::Span | Riffer::Tracing::NoOp::Span), Riffer::Tools::Response) -> void
   def record_tool_outcome(span, result)
     error_type = result.error_type
     span.set_attribute("error.type", error_type.to_s) if error_type
```
tsov, line 155: suggestion(non-blocking): `error.type` used to be the exception class name and is now always `"unhandled_error"`. Is that another breaking change we should list in the description? Anyone grouping or alerting on exception class in their traces would be affected.

lib/riffer/tool.rb
```diff
@@ -77,7 +76,20 @@ def call_with_validation(context:, **kwargs)
     end
 
     result
+  rescue Riffer::ValidationError => e
```
tsov: suggestion(non-blocking): Doesn't this also catch `ValidationError` raised inside the tool body? The LLM would think its arguments were wrong and retry. Should we scope the rescue to `params_builder.validate`?

tsov (approved):
Great idea! Just a bunch of questions/suggestions. Will approve to unblock

lib/riffer/tool.rb
bottrall: Great catch — adopted in 79a2e10f exactly as suggested: `Timeout.timeout(n, Riffer::TimeoutError)` with a rescue on precisely that class, so `Net::ReadTimeout`/`Net::OpenTimeout` from a tool's own HTTP call now land in `:unhandled_error` instead of masquerading as the tool's budget. It also brings `Riffer::TimeoutError` back (no constant removal), with one knowing trade-off: the sentinel is raised *inside* the block, so a tool's blanket `rescue => e` can intercept it — which we're treating as a feature (cleanup), and it's documented on the constant and in TOOL_ADVANCED.

— 🤖 Claude (beep boop)

lib/riffer/tool.rb
bottrall: Also adopted in 79a2e10f — the rescue is now scoped to just the `params_builder.validate` call, so a `Riffer::ValidationError` raised inside `#call` (e.g. from nested riffer machinery) lands in `:unhandled_error` instead of telling the LLM its arguments were wrong. Regression test added for exactly that case.

— 🤖 Claude (beep boop)

lib/riffer/tools/runtime.rb
bottrall: Agreed it's a real behavior change worth calling out — added a "Breaking (telemetry)" bullet to the PR description. The exception class name isn't lost, it moves to `exception.type` on the span's recorded exception event; keeping `error.type` sourced uniformly from the response's `error_type` is what lets traces distinguish "tool deliberately errored" from "tool crashed" by a single attribute.

— 🤖 Claude (beep boop)

lib/riffer/tool.rb
bottrall: It can, and that's a deliberate trade-off: class-only (`NoMethodError`) is rarely enough for the LLM to self-correct, and riffer's tool authors own what their exceptions say. Hosts that need redaction already have seams — rescue inside the tool, override `around_tool_call`, or use `response.exception` (which never serializes). If a real need shows up, an opt-in redaction config for `:unhandled_error` content is a natural follow-up rather than something to build speculatively — happy to open an issue if you'd like one tracked.

— 🤖 Claude (beep boop)

PR #415, opened 2026-08-28
feat: registry liveness, explicit registration, and Riffer::Testing stub factories

bottrall:
## Problem

`Riffer::Registrable` derives membership from `Class#subclasses`, which keeps returning superseded class generations — RSpec's `stub_const` re-creates a class per example, and Zeitwerk dev-mode reload does the same. Old generations linger with the same name and identifier until GC, so the first `find` after a reload or across stubbed examples could raise `DuplicateIdentifierError` from classes whose constant already points at a newer object.

Downstream test suites also had no sanctioned way to make an ephemeral tool or agent class resolvable by identifier — and per review feedback, consumer tests shouldn't need any knowledge of the registry to do it.

## Solution

Three layers:

- **Liveness filter.** `build_identifier_registry` keeps only subclasses whose real name still resolves back to the same object, walking each namespace segment with `autoload?(segment, false)` / `const_get(segment, false)` so probing a pending autoload never triggers a load mid-build. The check runs at rebuild time only — `find`/`all` stay O(1) memoized hash reads; a constant removed or restored with no subsequent rebuild trigger keeps serving the old memo until the next rebuild, a documented trade-off.
- **Minimal explicit-registration primitives.** Public `register(klass)` / `unregister(klass)` on `Registrable`: identifier from the class itself, direct subclass only, `Riffer::DuplicateIdentifierError` for any identifier collision between distinct classes in both directions (a class present in both the explicit overlay and the implicit sweep is not a duplicate of itself). Unsynchronized, matching `Providers::Repository`.
- **`Riffer::Testing`** — the consumer-facing story. `stub_agent("FaqAgent") { … }` / `stub_tool(...)` build a subclass, optionally assign it a top-level constant (so workflow runners or persisted class names resolve it like a real class, with the identifier derived from the name exactly as in production — a stub that needs no constant sets `identifier "..."` in its block, like any other config), evaluate the body, and register it in one call; `Riffer::Testing.reset!` removes everything stubbed since the last reset. Opt-in one-line adapters (`require "riffer/testing/rspec"` / `require "riffer/testing/minitest"`) wire the cleanup into each framework's lifecycle, so tests never see the registry. A leaked stub fails loudly — the next stub of the same identifier raises, pointing at missing adapter wiring. The adapters are Zeitwerk-ignored and neither rspec nor minitest becomes a dependency.

An earlier `with_registered` block API was removed in favor of the factories. New `docs/TESTING.md` is the consumer guide; riffer's own suite dogfoods the factories and the minitest adapter throughout (~300 call sites converted from ad-hoc `Class.new` fixtures).

## Proof

CI covers it: 33 new tests across `test/riffer/registrable_test.rb` and `test/riffer/testing_test.rb`, including regression tests verified red without each guard (pending-autoload segment walk, register guard ordering, self-collision identity exemption, adapter cleanup via an inner `Minitest::Test`). Eager-load test proves the Zeitwerk ignores; `shipped_signatures_test` forbids `RSpec`/`Minitest` types in shipped sigs. Full `bin/rake` (2284 runs, RuboCop, Steep) and `bin/rake docs:check` (28 pages) green locally.

## Summary by CodeRabbit

- **New Features**
  - Added reusable testing helpers for stubbing agents and tools, with registration tracking and reset support.
  - Added RSpec and Minitest integrations for automatic test cleanup.
  - Added explicit registration and unregistration for agents and tools.
  - Improved handling of anonymous and renamed classes during registry lookups.

- **Bug Fixes**
  - Registry rebuilding now removes stale implicit registrations while preserving explicit registrations.

- **Documentation**
  - Expanded registration guidance and added a dedicated testing guide.

### DIFF
--- a/docs/AGENTS.md
+++ b/docs/AGENTS.md
@@ -326,6 +326,37 @@ Only **named direct subclasses** are found:
 - Grandchildren are not visible to a grandparent's `find` or `all`. If your app defines an intermediate base class (`class ApplicationAgent < Riffer::Agent`), call `find`/`all` on the intermediate class to look up its subclasses.
 - Anonymous classes (`Class.new(Riffer::Agent)`) are never findable, even when they set an explicit `identifier`.
 - Two subclasses sharing an identifier raise `Riffer::DuplicateIdentifierError` at the first lookup.
+- A subclass whose constant no longer points at it — after a Zeitwerk reload or an RSpec `stub_const` — drops out of `find` and `all`. The check runs when the registry is rebuilt, which defining, registering, or unregistering a subclass triggers; removing or restoring a constant on its own does not, so lookups keep returning the old class until the next rebuild.
+
+### Registering an agent explicitly
+
+`register` adds an agent to its parent's registry by hand, `unregister` removes it, and `with_registered` scopes the pair to a block. This exists for test suites: an anonymous agent class is never found implicitly, so registering it is the way to make the code under test resolve it by identifier.
+
+```ruby
+stub_agent = Class.new(Riffer::Agent) do
+  identifier 'support_agent'
+  model 'mock/gpt-5-mini'
+end
+
+Riffer::Agent.with_registered(stub_agent) do
+  TriageWorkflow.new.run('What are your hours?')
+end
+# => stub_agent is unregistered again, even if the block raises
+```
+
+```ruby
+Riffer::Agent.register(stub_agent)   # findable until unregistered
+Riffer::Agent.unregister(stub_agent) # no-op if it was never registered
+```
+
+Explicit registration differs from implicit in a few ways:
+
+- Anonymous classes are allowed, as long as they declare an `identifier`. A blank identifier raises `Riffer::ArgumentError`.
+- The class must be a **direct** subclass of the receiver, mirroring implicit registration. `Riffer::Agent.register(SomeAppAgent)` raises `Riffer::ArgumentError` when `SomeAppAgent` descends from an intermediate base — call `register` on that base instead.
+- Taking an identifier already held by another agent, implicit or explicit, raises `Riffer::DuplicateIdentifierError`. Re-registering the same class raises too; there is no idempotent path.
+- The registration survives until you remove it — it is never dropped for a stale constant.
+
+Registration is not synchronized. Register during boot or from a single-threaded test, before concurrent lookups begin.
 
 ## Per-Call Tags
 
--- a/docs/AGENTS.md
+++ b/docs/AGENTS.md
@@ -321,10 +321,10 @@ Riffer::Agent.find('missing')         # => nil
 Riffer::Agent.all                     # => [SupportAgent, ...]
 ```
 
-Only **named direct subclasses** are found:
+Only **named direct subclasses** are found implicitly:
 
 - Grandchildren are not visible to a grandparent's `find` or `all`. If your app defines an intermediate base class (`class ApplicationAgent < Riffer::Agent`), call `find`/`all` on the intermediate class to look up its subclasses.
-- Anonymous classes (`Class.new(Riffer::Agent)`) are never findable, even when they set an explicit `identifier`.
+- Anonymous classes (`Class.new(Riffer::Agent)`) are never findable implicitly, even when they set an explicit `identifier` — see [Registering an agent explicitly](#registering-an-agent-explicitly).
 - Two subclasses sharing an identifier raise `Riffer::DuplicateIdentifierError` at the first lookup.
 - A subclass whose constant no longer points at it — after a Zeitwerk reload or an RSpec `stub_const` — drops out of `find` and `all`. The check runs when the registry is rebuilt, which defining, registering, or unregistering a subclass triggers; removing or restoring a constant on its own does not, so lookups keep returning the old class until the next rebuild.
 
--- a/docs/AGENTS.md
+++ b/docs/AGENTS.md
@@ -330,23 +330,13 @@ Only **named direct subclasses** are found:
 
 ### Registering an agent explicitly
 
-`register` adds an agent to its parent's registry by hand, `unregister` removes it, and `with_registered` scopes the pair to a block. This exists for test suites: an anonymous agent class is never found implicitly, so registering it is the way to make the code under test resolve it by identifier.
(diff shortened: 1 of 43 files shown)


docs/AGENTS.md
```diff
@@ -326,6 +326,37 @@ Only **named direct subclasses** are found:
 - Grandchildren are not visible to a grandparent's `find` or `all`. If your app defines an intermediate base class (`class ApplicationAgent < Riffer::Agent`), call `find`/`all` on the intermediate class to look up its subclasses.
 - Anonymous classes (`Class.new(Riffer::Agent)`) are never findable, even when they set an explicit `identifier`.
 - Two subclasses sharing an identifier raise `Riffer::DuplicateIdentifierError` at the first lookup.
+- A subclass whose constant no longer points at it — after a Zeitwerk reload or an RSpec `stub_const` — drops out of `find` and `all`. The check runs when the registry is rebuilt, which defining, registering, or unregistering a subclass triggers; removing or restoring a constant on its own does not, so lookups keep returning the old class until the next rebuild.
+
+### Registering an agent explicitly
```
Thibaut, line 331: question(blocking): have you considered going all the way and creating methods to stub agents and tools? (do the equivalent of `Class.new(Riffer::Agent)`, via `Riffer::Agent.stub` or something)

I feel the current approach leaks the internals of Riffer onto the client, when what the client wants is just to stub an agent or a tool as close to a real agent/tool as possible. Riffer's registry is completely irrelevant to the test in question.

It's common for gems to offer test helper methods. This way, if anything else needed to happen when one calls `Class.new(Riffer::Agent)`, the helper method would abstract it all out, instead of more internals needed to be "leaked".

bottrall: Yeah we did think through that as a possibility when planning this out.

But I ended up deciding that it may be overkill, considering the registration is only required if your code is using the registry (find/all). Anonymous classes work fine otherwise, now without any side effects which was one of the desired outcomes here.

`with_registration` is the only thing a consumer needs to know about in order to use anonymous classes in tests and doesn't bind us to any specific test runner.

Thibaut: Ok. I personally would take the `Riffer::Agent.stub` approach. IMHO tests should have no knowledge of the registry, and even less have to debug why an agent is missing from the registry (something that I'd expect every test that needs `with_registered` will run into). Also, that approach would likely result in considerably less (or less complex) documentation, which is my heuristic for "good interfaces". But not a blocking opinion.

lib/riffer/registrable.rb
```diff
@@ -13,32 +17,108 @@
...
+    @explicit_registrations = explicit_registrations.merge(key => klass)
+    @identifier_registry = nil
+  end
+
+  # Removes an explicit registration of +klass+, leaving implicit registrations
+  # untouched.
+  #--
+  #: (Class) -> void
+  def unregister(klass)
+    @explicit_registrations = explicit_registrations.reject { |_key, registered| registered.equal?(klass) }
```
Thibaut: suggestion(non-blocking): change this to update in place (it's an internal object) and break the loop after the class is found (or find its index and delete that)

Thibaut (approved):

docs/AGENTS.md
bottrall: Hmmmm ok so in this world `Riffer::Agent.stub` (I'd want to workshop the name a bit more 😅) stub defines the agent + registers it. Then we would need a public reset/cleanup api that the consuming test suite would have to add in via an after each (possibly global?).

bottrall: Went with this — you were right that `with_registered` leaked the registry onto every consumer test. It's now `Riffer::Testing` (ae58276c):

```ruby
tool = stub_tool(identifier: 'kb_search') do
  def call(context:, **) = text('We are open 9–5.')
end
```

One call builds the anonymous class, sets the identifier, and registers it — findable immediately, no scope to manage, no registry knowledge required. Cleanup is automatic: `require "riffer/testing/rspec"` or `require "riffer/testing/minitest"` in your test setup wires `Riffer::Testing.reset!` into the framework's lifecycle (riffer's own suite runs on the minitest one). If the wiring is missing, the failure is loud rather than silent — the next stub of the same identifier raises `DuplicateIdentifierError` naming both classes.

Two deliberate departures from the literal `Riffer::Agent.stub` shape: the factories live on `Riffer::Testing` (included into your example groups by the adapter) rather than on the production classes, and they're named `stub_agent`/`stub_tool` — minitest monkey-patches `Object#stub`, so `Riffer::Agent.stub` already means something in minitest suites.

`with_registered` is gone; `register`/`unregister` remain as the primitives underneath. Docs-wise this collapsed the registration sections in TOOLS/AGENTS to a short reference, with the consumer story now a single page: `docs/TESTING.md`.

— 🤖 Claude (beep boop)

Thibaut: Right. There's more to it 😅 I think a global cleanup makes sense? It's a well-established pattern for test functionality like this, and all testing frameworks support global hooks. We could also take inspiration on RSpec's `stub_const` and take a fake class name + agent/tool definition that way. I'm not attached to `Riffer::Agent.stub`. Was just the first thing that came to mind.

edit: hadn't yet seen the reply above when I wrote this.

docs/TESTING.md
```diff
@@ -0,0 +1,105 @@
...
+stub_agent(identifier: 'support_agent') { uses_tools [tool] }
+```
+
+The block is evaluated in the new class, so the whole [tool DSL](TOOLS.md) is available inside it — `description`, `params`, `timeout`, `call`.
+
+## Stubbing an agent
+
+```ruby
+it 'routes to the support agent' do
+  stub_agent(identifier: 'support_agent') do
```
Thibaut: suggestion(non-blocking): we have a DSL for identifier that would already fit perfectly into the `do` block. I would take a class name here, so the identifier is optional, and this gives you maximum flexibility to define an agent and use it like a normal agent (for example, the class name could be passed to a service like a workflow runner, that then runs said agent, or in same codebases the agent class could be stored in a model/database and you'd want to run an integration test on that, etc)


```suggestion
  stub_agent("SupportAgent") do
    identifier "support_agent" # optional (default is same as regular agents) 
```

(same for tools)

Thibaut (approved):
nice!

PR #422, opened 2026-09-04
fix(anthropic): merge caller output_config with structured output

bottrall:
## Problem

When an Anthropic agent had structured output configured, the provider assigned `params[:output_config]` from scratch. Any `output_config` a caller had supplied through `model_options`, such as `output_config: { effort: "high" }`, was silently dropped and only the structured-output `format` reached the API. Callers had no way to combine effort with structured output.

## Solution

`build_request_params` now spreads the caller's `output_config` first and overlays the structured-output `format` on top. Keys like `effort` survive. The structured-output `format` still wins over a caller-supplied `format`, because the run loop validates the response against the structured-output class and a foreign schema would break that validation. Without structured output, `output_config` passes through untouched as before.

The `reasoning:` option is deliberately not translated for Anthropic in this change. Riffer takes no position on mapping it to a native knob; callers use `output_config: { effort: ... }` directly.

`docs/CONFIGURATION.md` gains an `output_config` row and an example in the Anthropic section, noting that it composes with structured output.

## Proof

Three new unit tests on `build_request_params` in `test/riffer/providers/anthropic_test.rb` cover effort plus structured output, a caller-supplied format being overridden by structured output, and effort alone passing through. `bin/test`, `bin/lint`, and `bin/typecheck` all pass locally. CI covers the same.

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring Anthropic output effort through `output_config`.
  - Structured JSON responses now preserve additional output settings while applying the required response format.

- **Documentation**
  - Updated file-download documentation formatting.
  - Added an example showing Anthropic output-effort configuration.


docs/CONFIGURATION.md
```diff
@@ -365,8 +366,16 @@ class ReasoningAgent < Riffer::Agent
...
 end
+
+# With an output effort level
+class EffortAgent < Riffer::Agent
+  model 'anthropic/claude-opus-5'
+  model_options output_config: {effort: "high"}
+end
 ```
 
+`output_config` composes with `structured_output`: the JSON schema format is added alongside any other `output_config` keys you supply.
```
Thibaut: suggestion(non-blocking): remove this line. It's expected / doesn't require explanation that `model_options` and `structured_output` can work together.

Thibaut (approved):
good catch

PR #425, opened 2026-09-04
feat!: report how a run ended through Response#outcome

bottrall:
## Problem

When a structured-output response fails JSON parsing or schema validation, `response.structured_output` is nil and nothing on the response says why. `Run` discarded the message carried by `StructuredOutput::Result#error`, so a consumer whose schema requires `:a` and whose model returned `{"b": 1}` saw only a nil and hit `NoMethodError` far from the cause. `docs/AGENTS.md` said output is "automatically parsed and validated" and never described the failure case.

The fix exposed a broader shape problem. Knowing whether a run produced a usable result already meant checking `blocked?`, `interrupted?`, and `messages.last.finish_reason` separately, and adding a fourth accessor for structured output would have made that worse. Providers constrain generation but none of them report a schema mismatch on the wire, so Riffer's own validation is the only uniform signal and it needs a first-class home.

## Solution

`Riffer::Agent::Outcome` is a small value object with a normalized `reason` and an optional `detail` string, exposed as `response.outcome`. It mirrors the shape of `Providers::FinishReason`. The vocabulary is `:completed`, `:guardrail_blocked`, `:interrupted`, `:max_steps`, `:invalid_structured_output`, plus the abnormal provider finish reasons passed through verbatim. Precedence runs most-causal first: tripwire, then max_steps, then interrupt, then provider finish, then schema failure, then completed. A truncated response that also fails validation reports `:length`, with the raw stop value in `detail`.

`Run` now validates whenever a schema is configured rather than only when the provider's own JSON parse succeeded, which makes the parse-error branch reachable for the first time. The assistant message keeps the provider's parsed hash, and `MESSAGES.md` now says that it is parsed rather than schema-validated.

`Messages::Assistant` gains `finish_reason_raw` so the provider's raw stop value survives in generate mode instead of only on the stream event and trace attribute. `Session#rebuild_message` had been dropping `finish_reason` on history mutation and now carries both fields. Tracing gains `riffer.outcome.reason` and `riffer.outcome.detail`; the existing interrupt and tripwire span attributes are kept so dashboards keep working.

This is a breaking change. `Response#blocked?`, `Response#interrupted?`, and `Response#interrupt_reason` are removed in favour of `outcome.reason`, and `Response.new` requires `outcome:`. `response.tripwire` still carries the full tripwire object. No deprecation shim is provided, matching how prior API changes in this repo have shipped. Streaming is unchanged, no exception is raised, and a repair loop that feeds validation errors back to the model is out of scope.

## Proof

CI runs `bin/rake` (test, RuboCop, Steep). Locally: 2325 runs, 3485 assertions, 0 failures, 0 errors; no RuboCop offenses; no Steep type errors.

New run-level tests cover the reported reproduction (`{"b": 1}` against a schema requiring `:a` yields `:invalid_structured_output` with "a is required" in `detail`), non-JSON content, tripwire, interrupt with and without a reason, max_steps, and both precedence cases. Message tests cover `finish_reason_raw` round-tripping through `to_h`/`from_h` and surviving `rebuild_message`. A grep confirms no reference to the removed members remains in `lib/`, `docs/`, or `test/`.

## Summary by CodeRabbit

- **New Features**
  - Added unified response outcomes with reasons, optional details, and success status.
  - Preserved provider-specific raw finish reasons alongside normalized values.
  - Structured-output failures now report validation details through the response outcome.

- **Breaking Changes**
  - Replaced separate interruption and blocking response helpers with the unified outcome API. Update integrations accordingly.

- **Documentation**
  - Updated lifecycle, guardrail, streaming, tracing, message, and structured-output guidance to reflect the new response outcome and finish-reason behavior.

### DIFF
--- a/docs/AGENT_LIFECYCLE.md
+++ b/docs/AGENT_LIFECYCLE.md
@@ -30,9 +30,8 @@ agent.generate(prompt = nil, files: nil)
 ```ruby
 # New conversation (class method — recommended for simple calls)
 response = MyAgent.generate('Hello', context: {user_id: 123})
-puts response.content       # Access the response text
-puts response.blocked?      # Check if guardrail blocked (always false without guardrails)
-puts response.interrupted?  # Check if a callback interrupted the loop
+puts response.content         # Access the response text
+puts response.outcome.reason  # How the run ended (:completed, :guardrail_blocked, :interrupted, ...)
 
 # New conversation (instance method — when you need message history or callbacks)
 agent = MyAgent.new(context: {user_id: 123})
@@ -142,9 +141,9 @@ Works with both `generate` and `stream`. Only emits agent-generated messages (As
 
 Callbacks can interrupt the agent loop. This is useful for human-in-the-loop approval, cost limits, or content filtering.
 
-Use `agent.interrupt!` (or the lower-level `throw :riffer_interrupt`) to stop the loop. The response will have `interrupted?` set to `true` and contain the accumulated content up to the point of interruption.
+Use `agent.interrupt!` (or the lower-level `throw :riffer_interrupt`) to stop the loop. The response's `outcome.reason` will be `:interrupted` and `content` will hold the accumulated content up to the point of interruption.
 
-An optional reason can be passed to `interrupt!`. It is available via `interrupt_reason` on the response (generate) or `reason` on the `Interrupt` event (stream):
+An optional reason can be passed to `interrupt!`. It is available via `outcome.detail` on the response (generate) or `reason` on the `Interrupt` event (stream):
 
 ```ruby
 agent = MyAgent.new
@@ -155,9 +154,9 @@ agent.session.on_message do |msg|
 end
 
 response = agent.generate('Call the tool')
-response.interrupted?      # => true
-response.interrupt_reason  # => "needs human approval"
-response.content           # => last assistant content before interrupt
+response.outcome.reason  # => :interrupted
+response.outcome.detail  # => "needs human approval"
+response.content         # => last assistant content before interrupt
 ```
 
 **Streaming** — interrupts emit an `Interrupt` event:
@@ -188,7 +187,7 @@ agent.session.on_message { |msg| throw :riffer_interrupt if needs_approval?(msg)
 
 response = agent.generate('Do something risky')
 
-if response.interrupted?
+if response.outcome.reason == :interrupted
   approve_action(agent.session.messages)
   response = agent.generate('Approved, go ahead')  # executes pending tools, then calls the LLM
   # or: agent.generate                              # resume without a new turn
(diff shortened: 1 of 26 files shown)


docs/AGENT_LIFECYCLE.md
```diff
@@ -306,18 +305,44 @@ agent.context[:skills]        # the Skills::Context, if skills configured
...
-| `interrupt_reason`     | `String` / `Symbol` / `nil` | The reason passed to `throw :riffer_interrupt`                                                   |
 | `messages`             | `Array`                     | Full message history from the conversation                                                       |
 | `healed_tool_call_ids` | `Array[String]`             | `tool_call` ids filled with placeholder results during interrupt healing (else `[]`)             |
 | `token_usage`          | `TokenUsage` / `nil`        | Aggregate `Riffer::Providers::TokenUsage` across this run's LLM calls (`nil` when none reported) |
 | `steps`                | `Integer`                   | LLM calls made during this run (`0` when a before-guardrail blocks first); not the session's cumulative count |
 
+### response.outcome
+
+`response.outcome` is a `Riffer::Agent::Outcome` — the single place to read how the run ended. `reason` is always one of the values below; `detail` is a `String` with the specifics when there are any, else `nil`. `outcome.success?` is shorthand for `reason == :completed`.
```
cloetzi, line 320: suggestion(non-blocking): I'd recommend documenting this in a separate paragraph at the top.

lib/riffer/agent/outcome.rb
```diff
@@ -0,0 +1,48 @@
...
+#
+#   response = agent.generate("Analyze this")
+#   case response.outcome.reason
+#   when :completed then puts response.structured_output
+#   when :invalid_structured_output then warn response.outcome.detail
+#   end
+class Riffer::Agent::Outcome
+  # Finish reasons that mean the provider cut the turn short; they surface as
+  # the run's outcome verbatim.
+  PROVIDER_STOP_REASONS = %i[length content_filter context_window malformed_output error other].freeze #: Array[Symbol]
```
cloetzi: issue(non-blocking): We are duplicating values here, why can't we use a shared constant? That's a significant future bug-risk.

lib/riffer/agent/run.rb
```diff
@@ -215,18 +219,39 @@ def tripwire_response(agent, stream_yielder, tripwire, all_modifications, token_
...
+      outcome: final_outcome(message, result, interrupted: interrupted, interrupt_reason: interrupt_reason),
       modifications: all_modifications,
-      structured_output: validate_structured_output(agent, response),
+      structured_output: result&.object,
       **extra,
     )
   end
 
+  # Ordered most-causal first: a stop the caller forced outranks what the
+  # provider reported, which outranks how riffer post-processed the content.
```
cloetzi: nitpick(non-blocking): Wut?

lib/riffer/agent/run.rb
```diff
@@ -215,18 +219,39 @@ def tripwire_response(agent, stream_yielder, tripwire, all_modifications, token_
...
+  #--
+  #: (Riffer::Messages::Assistant?, Riffer::Agent::StructuredOutput::Result?, interrupted: bool, interrupt_reason: (String | Symbol)?) -> Riffer::Agent::Outcome
+  def final_outcome(message, result, interrupted:, interrupt_reason:)
```
cloetzi, line 242: question(non-blocking): Can we harden this somehow and make it not order dependent? I think as-is it is bug-prone in the future. E.g. if we get a new provider stop reason, but we miss to add it to the list, it is now reported as structured output failure or completed?

I think the chances are pretty unlikely, but this single statement with somewhat unrelated checks makes me a bit nervous.

lib/riffer/agent/run.rb
```diff
@@ -215,18 +219,39 @@ def tripwire_response(agent, stream_yielder, tripwire, all_modifications, token_
   end
 
   #--
-  #: (Riffer::Agent, Array[Riffer::Guardrails::Modification], **untyped) -> Riffer::Agent::Response
-  def final_response(agent, all_modifications, **extra)
-    response = agent.session.final_assistant_message
+  #: (Riffer::Agent, Array[Riffer::Guardrails::Modification], ?interrupted: bool, ?interrupt_reason: (String | Symbol)?, **untyped) -> Riffer::Agent::Response
+  def final_response(agent, all_modifications, interrupted: false, interrupt_reason: nil, **extra)
+    message = agent.session.final_assistant_message
+    result = structured_output_result(agent, message)
```
cloetzi: suggestion(non-blocking): Move the `agent.structured_output` check here. I got initially confused and thought we are in a structured-output-only code path.

lib/riffer/agent/run.rb
```diff
@@ -470,4 +497,13 @@ def record_run_outcome(span, response)
     span.set_attribute("riffer.tripwire.reason", tripwire.reason)
     span.set_attribute("riffer.tripwire.phase", tripwire.phase.to_s)
   end
+
+  #--
+  #: (Riffer::Agent::Outcome) -> String?
+  def interrupt_reason_attribute(outcome)
+    case outcome.reason
+    when :max_steps then Riffer::Agent::INTERRUPT_MAX_STEPS.to_s
```
cloetzi, line 507: question(non-blocking): Should we start using constants for all those symbols and have one central module that maintains them?

cloetzi (approved):
Overall, a huge improvement. I think there's some potential to further harden the implementation against future bugs and improve the code-readability. Up-to-you how far you wanna push it.

docs/AGENT_LIFECYCLE.md
bottrall: Done in ef26314. The Response Attributes section now opens with a short paragraph: start with `outcome.reason`, everything else on the response is detail for that reason, and `content` / `structured_output` only mean something when it is `:completed`.

— 🤖 Claude (beep boop)

lib/riffer/agent/outcome.rb
bottrall: Agreed. In ef26314 `PROVIDER_STOP_REASONS` is derived from `Riffer::Providers::FinishReason::VALUES` minus a new `NORMAL_FINISH_REASONS` (`:stop`, `:tool_calls`), so there is one list to maintain. A test in `outcome_test.rb` asserts every provider finish reason is either normal or an outcome value, so a future addition cannot slip through.

— 🤖 Claude (beep boop)

lib/riffer/agent/run.rb
bottrall: Fair, that sentence was doing too much. Reworded in ef26314: checked in the order things happened, the loop being stopped (max_steps or interrupt) beats the provider's finish reason, which beats riffer's own validation of the content. A truncated response that also fails the schema therefore reports `:length`, not `:invalid_structured_output`.

— 🤖 Claude (beep boop)

lib/riffer/agent/run.rb
bottrall: The ordering itself is intentional rather than incidental: several of these can be true at once (interrupted and `:length`, or `:length` and invalid JSON) and something has to pick which one the caller sees. So I kept the chain, but closed the specific risk you describe. In ef26314 the provider branch now fails closed: any finish reason outside `NORMAL_FINISH_REASONS` is treated as a provider stop, so a newly added reason cannot fall through to `:completed` or `:invalid_structured_output`. Combined with deriving the list (thread above) and the new invariant test, adding a finish reason to `FinishReason::VALUES` is the only edit needed.

— 🤖 Claude (beep boop)

lib/riffer/agent/run.rb
bottrall: Done in ef26314. The call site is now `result = agent.structured_output && structured_output_result(agent, message)`, so it is obvious that most runs skip this and the helper only handles the message-present case.

— 🤖 Claude (beep boop)

lib/riffer/agent/run.rb
bottrall: I would hold off on per-symbol constants. Symbols as enum values are how the rest of the gem already works (`FinishReason::VALUES`, every provider's finish-reason mapping), and `Outcome::MAX_STEPS` next to `when :length` would sit oddly. The idiomatic hardening here is a type rather than constants: an RBS literal union for `Outcome#reason` (`:completed | :guardrail_blocked | ...`) makes a typo a Steep error at typecheck time. That needs `FinishReason#reason` typed the same way since it flows straight into `Outcome.new`, so it is a small follow-up of its own rather than something to fold in here.

— 🤖 Claude (beep boop)

PR #429, opened 2026-09-11
fix: accept JSON integers for Float params

cloetzi:
## What broke

Any whole-number value for a `Float` param failed validation:

```ruby
p = Riffer::Params.new
p.required :treatments, Array do
  required :price, Float
end
Riffer::Agent::StructuredOutput.new(p).parse_and_validate('{"treatments":[{"price":120}]}').error
# => "Validation error: treatments[0].price must be a number"
```

## Why

`Param#valid_type?` checked `value.is_a?(type)`. `Float` maps to JSON Schema `"number"`, and `"number"` includes integers — so a model returning `120` is valid per the schema we emit, but `JSON.parse` yields an `Integer` and the `is_a?(Float)` check rejected it. `Params#validate_typed_array` had the same defect for `of: Float` arrays.

## Fix

- `Param#valid_type?`: a `Float` param accepts any `Numeric`. `true`/`false` are not `Numeric` in Ruby, so booleans are still rejected.
- `Params#validate_typed_array`: same rule for `of: Float` items.
- `Integer` stays strict — JSON Schema `"integer"` excludes `1.0`.

## Coercion decision

The validated value for a `Float` param is coerced with `to_f`, and so are `of: Float` array items, so callers always receive a `Float` rather than a type that varies with whether the model happened to emit a decimal point. Nothing else is coerced — this is not a general string-to-number conversion.

After the fix the repro returns `{treatments: [{price: 120.0}]}` with no error.

`bundle exec rake` (test + rubocop + steep) is green; RBS regenerated.

## Summary by CodeRabbit

- **New Features**
  - Float parameters now accept whole-number and decimal numeric values, converting them to floating-point values.
  - The same behavior applies to nested parameters and typed arrays containing Float values.
  - Invalid strings and boolean values continue to be rejected.

- **Bug Fixes**
  - Integer parameters continue to reject fractional numeric values.

- **Documentation**
  - Updated tool documentation to clarify numeric coercion and validation behavior.

### DIFF
--- a/lib/riffer/params.rb
+++ b/lib/riffer/params.rb
@@ -79,6 +79,11 @@ class Riffer::Params
 
   # Validates arguments against parameter definitions.
   #
+  # A Float param accepts an Integer (JSON Schema <tt>"number"</tt> covers
+  # integers) and its value is coerced with +to_f+, so callers always get a
+  # Float. The same holds for the items of an <tt>of: Float</tt> array. No other
+  # type is coerced.
+  #
   # Raises Riffer::ValidationError if validation fails.
   #
   #--
@@ -112,7 +117,7 @@ class Riffer::Params
 
       value = validate_nested(param, value, errors)
 
-      validated[param.name] = value
+      validated[param.name] = param.type == Float ? value.to_f : value
     end
 
     raise Riffer::ValidationError, errors.join("; ") if errors.any?
@@ -178,7 +183,6 @@ class Riffer::Params
       validate_nested_array_of_objects(param, value, errors)
     elsif param.type == Array && param.item_type
       validate_typed_array(param, value, errors)
-      value
     else
       value
     end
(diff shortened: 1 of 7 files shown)


lib/riffer/params/param.rb
```diff
@@ -114,13 +114,20 @@ def initialize(name:, type:, required:, description: nil, enum: nil, default: ni
 
   # Validates that a value matches the expected type.
   #
+  # A Float param maps to JSON Schema <tt>"number"</tt>, which covers integers,
+  # so a whole number arrives as an Integer and must be accepted. Integer stays
+  # strict, matching JSON Schema <tt>"integer"</tt>, which excludes
+  # <tt>1.0</tt>. +true+ and +false+ are not Numeric in Ruby, so they are still
+  # rejected for a Float param.
```
bottrall: suggestion(non-blocking): remove this comment.

I don't think it adds any information that you can't figure out by looking at the code. Or it should be a much smaller comment next to `elsif type == Float`.

lib/riffer/params.rb
```diff
@@ -112,7 +117,7 @@ def validate(arguments)
 
       value = validate_nested(param, value, errors)
 
-      validated[param.name] = value
+      validated[param.name] = param.type == Float ? value.to_f : value
```
bottrall: suggestion(non-blocking): Add a `coerce_value` method below with a case statement.

This likely won't be the last time we want to coerce some values and it will help us avoid continuing to throw the logic in here.

bottrall (approved):

PR #438, opened 2026-09-11
chore(steep): adopt all_error diagnostics preset

bottrall:
## Problem

Steep was running with the `strict` preset, which leaves several diagnostics at `:hint` or off. In particular `UnreachableBranch` was silent, so a handful of methods carried runtime type guards that their own signatures already made impossible to reach, and `MethodDefinitionMissing` was hidden, so nothing checked that every RBS member has an implementation. The private RBS override for `Anthropic::Resources::Messages#stream` was also silently redefining the gem-shipped signature rather than extending it after #436 adopted the rbs collection.

## Solution

Switch the `Steepfile` to `D::Ruby.all_error` with no overrides and fix everything that surfaced, taking the strict reading of "fix" rather than loosening types or downgrading diagnostics.

**Missing method implementations (~170).** Steep 2.1 only registers `def` nodes as implementations, so every RBS `attr_reader`/`attr_accessor` member trips `MethodDefinitionMissing` (upstream soutaro/steep#1036, open since 2024 with two stalled PRs). Steep's documented answer is the `# @dynamic` annotation, so each attribute line carries one as a trailing comment after its existing `#:` type:

```ruby
attr_reader :content #: String # @dynamic content
attr_accessor :model_options #: Hash[Symbol, untyped] # @dynamic model_options, model_options=
```

Steep scans trailing comments for annotations, the RBS type parser treats the second `#` as a comment so rbs-inline output is unchanged, and RDoc ignores trailing comments. Steep's `UnexpectedDynamicMethod` fires if a listed name has no RBS member, so the annotations are self-enforcing in both directions.

**Unreachable branches (10).** These were all defensive guards in our own code where the declared type already rules the branch out, so the guards go rather than the types widening. `Messages::Base.from_hash`, `FilePart.from_hash` and `Agent::Run#normalize_tags` drop their `is_a?(Hash)` raises, and `Providers::Repository#register` drops its no-block raise, along with the four tests and the one docs sentence that asserted them. Where a `Hash#[]` or `Array#first` lookup can genuinely miss at runtime, the lookup-then-guard becomes a single `fetch { raise }` with the same exception class and message. `semconv_provider_name` becomes `class_name = name or return "unknown"`, keeping anonymous classes uncached. The Bedrock cache ttl ternary becomes a guarded assignment because Steep's union unification otherwise types it as `nil`.

**Duplicate method definition (2).** The private `stream` override now ends in `| ...` so it extends the gem's overloads instead of replacing them.

Behaviour change to be aware of: passing a non-Hash to those three public methods or calling `register` without a block now fails with a `NoMethodError` or `ArgumentError` from Ruby itself instead of a `Riffer::ArgumentError`. The type signatures were already the contract.

## Proof

CI runs `bin/rake` (test + rubocop + steep:check). Locally, `bundle exec steep check --no-daemon` reports no type errors (with a negative control confirming the trailing annotations are what satisfy `MethodDefinitionMissing`), `bin/lint` reports no offenses, and `bin/test` passes (2360 runs, 0 failures, 2 pre-existing skips). Regenerating `sig/generated` from source produces a byte-identical tree, and generated RDoc contains no `@dynamic` text.

## Summary by CodeRabbit

* **Changes**
  * Ruby diagnostics for the library are now reported as errors, improving visibility of issues.
  * Registration APIs now permit entries without a factory.
  * Invalid tag and message inputs may now fail according to their own behavior rather than with a library-specific argument error.
  * Cache TTL values are ignored when cache settings are not provided as key-value data.
  * Providers without a class name now use `"unknown"` as their provider name.

* **Documentation**
  * Updated tag handling documentation to reflect current behavior.

### DIFF
--- a/lib/riffer/agent/config.rb
+++ b/lib/riffer/agent/config.rb
@@ -5,6 +5,9 @@
 # Riffer::Agent subclass. Procs are stored unresolved and resolved per-instance
 # later.
 class Riffer::Agent::Config
+  #--
+  # @dynamic identifier, model, instructions, model_options, model_options=, structured_output, max_steps, max_steps=, tools_config, tools_config=, mcp_configs, tool_runtime, skills_config, skills_config=, guardrails
+
   DEFAULT_MAX_STEPS = 16 #: Integer
 
   # The configured agent identifier.
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -67,12 +67,12 @@ Style/ClassAndModuleChildren:
 Lint/UnusedMethodArgument:
   AllowUnusedKeywordArguments: true
 
-# Additions, not overrides: RBS-inline annotations must lint.
+# Additions, not overrides: RBS-inline and Steep annotations must lint.
 Layout/LeadingCommentSpace:
   AllowRBSInlineAnnotation: true
 
 Layout/LineLength:
-  AllowedPatterns: ['#:']
+  AllowedPatterns: ['#:', '# @dynamic']
 
 Minitest/MultipleAssertions:
   Max: 3
--- a/Steepfile
+++ b/Steepfile
@@ -9,11 +9,5 @@ target :lib do
 
   check "lib"
 
-  # Steep 2.1 only registers `def` nodes as implementations, so it reports every
-  # RBS `attr_reader`/`attr_accessor` member as unimplemented. The only in-source
-  # escape is a `# @dynamic` comment beside each attribute, so the whole
-  # diagnostic stays a hint (its level under the `strict` preset).
-  configure_code_diagnostics(D::Ruby.all_error) do |hash|
-    hash[D::Ruby::MethodDefinitionMissing] = :hint
-  end
+  configure_code_diagnostics(D::Ruby.all_error)
 end
--- a/Steepfile
+++ b/Steepfile
@@ -9,5 +9,11 @@ target :lib do
 
   check "lib"
 
-  configure_code_diagnostics(D::Ruby.strict)
+  # Steep 2.1 only registers `def` nodes as implementations, so it reports every
+  # RBS `attr_reader`/`attr_accessor` member as unimplemented. The only in-source
+  # escape is a `# @dynamic` comment beside each attribute, so the whole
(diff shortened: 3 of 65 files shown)


lib/riffer/agent/config.rb
```diff
@@ -5,6 +5,9 @@
 # Riffer::Agent subclass. Procs are stored unresolved and resolved per-instance
 # later.
 class Riffer::Agent::Config
+  #--
+  # @dynamic identifier, model, instructions, model_options, model_options=, structured_output, max_steps, max_steps=, tools_config, tools_config=, mcp_configs, tool_runtime, skills_config, skills_config=, guardrails
```
Thibaut: question(non-blocking): those can't be inlined next to the method/attribute in question?

Thibaut (approved):

lib/riffer/agent/config.rb
bottrall: Good call, they can. Switched to a trailing form on each attribute line in b5735fb:

```ruby
attr_reader :content #: String # @dynamic content
attr_accessor :model_options #: Hash[Symbol, untyped] # @dynamic model_options, model_options=
```

I tested three placements before picking this one:

- **Standalone `# @dynamic` line above each attribute** works with Steep, but rbs-inline copies the doc block into the generated RBS, so the annotation leaked into `sig/`. It also needed a `#--` stop per attribute to stay out of RDoc.
- **Trailing after the `#:` type** (chosen) works because Steep scans trailing comments for annotations and the RBS type parser treats the second `#` as a comment. Verified: `steep check --no-daemon` clean with a negative control confirming the trailing annotation is what satisfies it, regenerated `sig/generated` byte-identical, zero `@dynamic` in RDoc output, rubocop clean, full suite green.
- **The grouped header** I had before is now gone, along with its 52 `#--` stop lines and the `.rubocop.yml` line-length change (every one of these lines already carries `#:`, which the existing allowlist covers).

Net effect against main is 177 one-line edits instead of 104 added lines. Steep's `UnexpectedDynamicMethod` still fires for any name without an RBS member, so the annotations stay self-enforcing.

— 🤖 Claude (beep boop)

PR #440, opened 2026-09-18
feat(agent): inherit configuration on subclass

ChadMcCallum-Jane:
## Problem

A subclass of a Riffer agent starts from an empty `Config`, so it loses every setting its parent declared and falls back to each setting's default. Reusing an agent's configuration means restating it.

That blocks a pattern where a subclass is a deliberate variation of the agent above it, declaring only the handful of settings under test and taking the rest from its parent. The copy belongs in the gem rather than in a consumer's own `inherited` hook: the class that owns the fields is the only place a field added in a later version can't be silently dropped from every subclass.

## Solution

`Config` and `Params` each override `initialize_copy`, so `dup` produces a copy holding no collection in common with the original. `Agent.inherited` is then:

```ruby
def self.inherited(subclass)
  super
  copy = config.dup
  copy.identifier = nil
  subclass.instance_variable_set(:@config, copy)
end
```

The hook fires before the subclass body runs, so the subclass's own DSL calls override on top. `super` is load-bearing: `Registrable#inherited` busts the identifier-registry memo.

`identifier` is the only field a subclass must not inherit — configuration on the parent, identity on the subclass. Copied, two classes would claim one identifier and `Registrable` raises `DuplicateIdentifierError` at the first lookup. Cleared, the subclass derives its own from its class name.

An Array `tools_config` is duplicated, since `attr_accessor` leaves it open to `<<` from outside; a Proc is shared, having nothing to alias. `Params` rebuilds its parameters and recurses through `nested_params`, `Param` itself being read-only.

### Things worth a reviewer's attention

**Two behaviour changes, not one.**

Existing subclasses inherit rather than defaulting: a subclass that didn't redeclare `max_steps` used to get `DEFAULT_MAX_STEPS` and now gets its parent's value. The test that asserted the old behaviour is updated here, which is the honest statement of it.

And `Config` now stores `nil` for an undeclared `tool_runtime`, resolving `Riffer.config.tool_runtime` in the reader instead of at construction. So an agent declaring no runtime follows a later change to the global, where it previously kept whatever was current when its class loaded. `docs/AGENTS.md` promised "captured at agent class definition time" and is corrected.

That second change is what removes the special case. Resolving the global eagerly meant a copy couldn't tell a declared runtime from a defaulted one, so the field had to be excluded — which left a parent declaring `tool_runtime Threaded` with a child that silently took the global while every other setting inherited.

It's all landing as a minor bump because no consumer we know of has agent subclasses relying on the old fallbacks. Happy to re-cut as a major if you'd rather.

**Both `use_mcp` and `uses_tools` were documented as not inherited.** Both are now, in `AGENTS.md` and `MCP.md`. The consequence a consumer could trip on: declaring the same MCP tag on a parent and a child registers it twice. That's the part of this change I'd most want a second opinion on.

**`sig/generated` is regenerated** from the inline annotations and committed.

## Proof

- `rake test`, `rubocop`, `steep:check`, and `rake docs:check` all pass, the last confirming links and anchors resolve across all 28 pages. The 63 errors in a full local `rake test` run are AWS credential failures in the Bedrock provider tests, unrelated to this change.
- New tests in `test/riffer/agent_test.rb`: override without touching the parent, two-level inheritance, the `identifier` exclusion, `tool_runtime` inherited when declared and resolved from the global when not, a config following a later change to the global, and isolation for every copied collection — `mcp_configs`, `guardrails`, `skills_config`, an Array tool list, and a schema both flat and nested. Plus a guard that fails when a field is added to `Config`, so the new one gets a deliberate copy-or-exclude decision.
- More in `test/riffer/params_test.rb` covering `dup` directly: isolation at depth, and that the copy produces an identical JSON schema. A copy that isolated but subtly altered the schema would be worse than no copy.
- Mutation-tested, every case failing at least one test: adding `identifier` to the copy; restoring the eager `tool_runtime` default; copying `tool_runtime` through the reader rather than the ivar; stripping either `initialize_copy` to a bare `super`; dropping the `mcp_configs`, `skills_config`, or tool-list duplication; duplicating a Proc tool list; sharing the schema; and making the `Params` copy non-recursive, which fails only the nested test and so shows the recursion earns its place. The `identifier` case needed subclassing directly rather than through `stub_agent`, which assigns an identifier of its own and masked it.
- Verified end to end against a downstream application pinned at this branch via a path override: its full agent suite passes, and removing the hook fails three of its examples.

## Summary by CodeRabbit

* **Enhancements**
  * Agent subclasses now inherit parent configuration while maintaining independent mutable settings.
  * Subclass identifiers are assigned independently rather than inherited.
  * Tool runtime settings use the current global configuration when no agent-specific runtime is set.

* **Bug Fixes**
  * Changes to subclass tools, guardrails, skills, structured output, parameters, and other nested settings no longer unintentionally affect parent agents.

* **Documentation**
  * Updated inheritance and MCP registration guidance to reflect current behavior, including identifier handling.


docs/AGENTS.md
```diff
@@ -126,7 +126,7 @@ end
 
 ### use_mcp
 
-Loads tools from registered [MCP](MCP.md) servers by tag. Like `uses_tools`, **`use_mcp` is not inherited**—add it on each subclass that should include MCP tools.
+Loads tools from registered [MCP](MCP.md) servers by tag. A subclass inherits its parent's registrations and `use_mcp` adds to them, so a tag declared on both is registered twice.
```
bottrall: issue(non-blocking): Additionally we need to change `docs/MCP.md:94` as it still reads that:
> **`use_mcp` is not inherited** from the superclass

lib/riffer/params.rb
```diff
@@ -42,6 +42,21 @@ def self.from_json_schema(schema)
     params
   end
 
+  # Returns a copy holding no collection in common with this one, so a caller can
+  # keep defining parameters on either without reaching the other.
+  #
+  # Param itself is read-only, so its instances are shared rather than rebuilt.
+  # Only +nested_params+ needs following, being a Params with a mutable
+  # collection of its own.
+  #
```
bottrall: issue(non-blocking): This violates the `.claude/rules/comments.md`.

I'd suggest trimming down to:
```rb
# Returns a copy holding no collection in common with this one, so a caller can
# keep defining parameters on either without reaching the other.
#--
#: () -> Riffer::Params
```

test/riffer/agent_test.rb
```diff
@@ -203,9 +203,158 @@
       end
       child = stub_agent("ChildAgent", base: parent)
 
+      # A subclass inherits the parent's settings in a config object of its own.
```
bottrall: suggestion(non-blocking): delete this.

It violates `.claude/rules/comments.md` by restating the below assertions.

lib/riffer/params.rb
```diff
@@ -42,6 +42,21 @@ def self.from_json_schema(schema)
...
+  #--
+  #: () -> Riffer::Params
+  def deep_copy
+    copy = self.class.new
+    @parameters.each { |param| copy.parameters << copy_param(param) }
+    copy
+  end
```
bottrall: suggestion(non-blocking): I think the more idiomatic approach to this would be to override `initialize_copy`, like so:
```rb
#: (Riffer::Params) -> void
def initialize_copy(source)
  super
  @parameters = source.parameters.map { |param| copy_param(param) }
end
```
Callers then write `structured_output&.dup`.

lib/riffer/agent/config.rb
```diff
@@ -175,6 +175,33 @@ def guardrails_for(phase)
...
+      instructions: instructions,
+      model_options: model_options.dup,
+      structured_output: structured_output&.deep_copy,
+      max_steps: max_steps,
+      tools_config: tools_config.is_a?(Array) ? tools_config.dup : tools_config,
+      mcp_configs: mcp_configs.dup,
+      skills_config: skills_config&.dup,
+      guardrails: guardrails.transform_values(&:dup),
+    )
+  end
```
bottrall: suggestion(non-blocking): Similar to the previous comment I think it would be better to override `initialize_copy` to produce a faithful copy of a config. Then in `Riffer::Agent` we adjust as required by the agent:
```rb
def self.inherited(subclass)
  super
  copy = config.dup
  copy.identifier = nil
  copy.tool_runtime = Riffer.config.tool_runtime
  subclass.instance_variable_set(:@config, copy)
end
```

lib/riffer/agent/config.rb
```diff
@@ -175,6 +175,33 @@ def guardrails_for(phase)
...
+  # Returns a copy for a subclass to start from, sharing no mutable collection
+  # with this one.
+  #
+  # An Array +tools_config+ is duplicated because +attr_accessor+ leaves it open
+  # to +<<+ from outside; a Proc is shared, having nothing to alias.
+  #
+  # +identifier+ is excluded because it is configuration here but identity on the
+  # subclass; two classes claiming one identifier raise
+  # Riffer::DuplicateIdentifierError at the next registry lookup. +tool_runtime+
+  # is excluded so a subclass still picks up +Riffer.config.tool_runtime+ instead
```
bottrall: **suggestion(blocking):** Excluding `tool_runtime` from the copy draws an odd line. A parent that explicitly declares `tool_runtime Threaded` gets a child that silently ignores it and takes the global instead, while every other setting inherits.

I dug into this and to me the root cause is that `Config#initialize` eagerly resolves `Riffer.config.tool_runtime` as its default, so the copy can't distinguish "explicitly set" from "defaulted to the global". Rather than exclude the field, we may want to store `nil` when unset and fall back to `Riffer.config.tool_runtime` in the reader. Then copying works as intended: an unset parent's child follows the current global, and an explicit parent's child inherits the explicit value. That removes the special case, its test, and the docs paragraph. Only `identifier` needs excluding, as it is the special case here.

bottrall (changes requested):

lib/riffer/params.rb
ChadMcCallum-Jane: TIL about initialize_copy!  I'll refactor to use that, thanks

docs/AGENTS.md
```diff
@@ -126,7 +126,7 @@ end
 
 ### use_mcp
 
-Loads tools from registered [MCP](MCP.md) servers by tag. Like `uses_tools`, **`use_mcp` is not inherited**—add it on each subclass that should include MCP tools.
+Loads tools from registered [MCP](MCP.md) servers by tag. A subclass inherits its parent's registrations and `use_mcp` adds to them. Duplicate tags are de-duplicated when tools resolve.
```
bottrall: thought(non-blocking): I wonder if we can simply delete the second sentence here? 🤔

Considering we call out explicitly in the "Inheritance" section the subclassing behaviour and we don't have comments explaining the subclassing behaviour on the rest of the DSL, explaining it here seems a bit strange.

docs/AGENTS.md
```diff
@@ -304,6 +304,29 @@ MyAgent.config.max_steps  # => 8
...
+
+TerseAgent.config.model      # => 'openai/gpt-5-mini'
+TerseAgent.config.max_steps  # => 2
+BaseAgent.config.max_steps   # => 8
+```
+
+The copy shares no mutable collection with the parent, so `use_mcp`, `guardrail`, and a `skills` block on a subclass never reach the class it inherits from.
+
+One setting is deliberately not copied. `identifier` is configuration on the parent but identity on the subclass, so a subclass derives its own from its class name; two classes claiming one identifier would raise `Riffer::DuplicateIdentifierError` at the first lookup.
+
```
bottrall, line 327: thought(non-blocking): Does this all deserve to be called out? Or is this mostly expected behaviour from a consumer's perspective? 🤔

Thinking through this the only thing the audience really needs to know about here is that the `identifier` is not inherited. So maybe we simplify this section with only that called out.

docs/MCP.md
```diff
@@ -91,7 +91,7 @@ Tool names must be unique across `uses_tools` and all included MCP servers; dupl
 
 ### Subclassing
 
-Like [`uses_tools`](AGENTS.md#uses_tools), **`use_mcp` is not inherited** from the superclass. Declare `use_mcp` on each agent class that should load MCP tools.
+A subclass [inherits](AGENTS.md#inheritance) its parent's registrations, and `use_mcp` adds to them. Repeating a parent's tag adds a second entry rather than replacing the first. Duplicate tags are de-duplicated when tools resolve, so a server's tools are never loaded twice.
```
bottrall: suggestion(non-blocking): Leading on from the previous two comments, we might be able to just delete this section entirely.

lib/riffer/agent/config.rb
```diff
@@ -175,6 +183,35 @@ def guardrails_for(phase)
...
+  #--
+  #: (untyped) -> untyped
+  def deep_dup(value)
+    case value
+    when Hash then value.transform_values { |entry| deep_dup(entry) }
+    when Array then value.map { |entry| deep_dup(entry) }
+    else value
+    end
+  end
+
```
bottrall: issue(blocking): This should be a private method.

Additionally once it's a private method we can remove comment. It's not needed for public documentation and doesn't provide any extra information that you can't get from reading the code.

lib/riffer/agent/config.rb
```diff
@@ -175,6 +183,35 @@ def guardrails_for(phase)
...
+  #: (Riffer::Agent::Config) -> void
+  def initialize_copy(source)
+    super
+    @model_options = deep_dup(source.model_options)
+    @mcp_configs = deep_dup(source.mcp_configs)
+    @guardrails = deep_dup(source.guardrails)
+    @tools_config = deep_dup(source.tools_config)
+    @skills_config = source.skills_config&.dup
+    @structured_output = source.structured_output&.dup
+  end
```
bottrall, line 202: suggestion(non-blocking): also move this under private.

Ruby makes this private by default so functionally it will be the same, but for consistency and explicitness we should move it below `private`.

lib/riffer/agent.rb
```diff
@@ -28,6 +28,23 @@ def self.config
...
+  # Riffer::DuplicateIdentifierError at the next registry lookup. Cleared, the
+  # subclass derives its own from its class name.
+  #--
+  #: (Class) -> void
+  def self.inherited(subclass)
+    super
+    copy = config.dup
+    copy.identifier = nil
+    subclass.instance_variable_set(:@config, copy)
+  end
```
bottrall, line 41: suggestion(blocking): This should be private.

```rb
def self.inherited(subclass)
  super
  copy = config.dup
  copy.identifier = nil
  subclass.instance_variable_set(:@config, copy)
end
private_class_method :inherited
```

Additionally once it's private the comment can either be deleted or reduced. Personally I think it's self-explanatory, but some might make the argument that it's necessary to explain why we don't copy the identifier.

lib/riffer/params.rb
```diff
@@ -151,6 +151,32 @@ def to_json_schema(strict: false)
...
+    Riffer::Params::Param.new(
+      name: param.name,
+      type: param.type,
+      required: param.required,
+      description: param.description,
+      enum: param.enum,
+      default: param.default,
+      item_type: param.item_type,
+      nested_params: param.nested_params.dup,
+    )
```
bottrall: suggestion(non-blocking): add `initialize_copy` to `Riffer::Params::Param`.

That would simplify everything to:
```rb
# lib/riffer/params.rb

#: (Riffer::Params) -> void
def initialize_copy(source)
  super
  @parameters = source.parameters.map(&:dup)
end

# lib/riffer/params/param.rb

# +dup+ would leave the copy sharing this one's nested Params, so a caller
# defining a parameter on either would reach the other.
#--
#: (Riffer::Params::Param) -> void
def initialize_copy(source)
  super
  @nested_params = source.nested_params&.dup
end
```

bottrall (changes requested):
Thanks @ChadMcCallum-Jane, looking much better but I think there's a few more things we can clean up before moving along.

lib/riffer/skills/config.rb
```diff
@@ -67,4 +67,15 @@ def activate_tool(value = nil)
...
+
+  private
+
+  # +dup+ would leave the copy sharing this one's activation list, so activating
+  # a skill on either would reach the other. A Proc has nothing to alias.
+  #--
+  #: (Riffer::Skills::Config) -> void
+  def initialize_copy(source)
+    super
+    @activate = source.activate.dup if source.activate.is_a?(Array)
```
bottrall: suggestion(non-blocking): Use `Riffer::Helpers::DeepDup` here.

```suggestion
    @activate = Riffer::Helpers::DeepDup.call(source.activate)
```

bottrall (approved):
Thanks @ChadMcCallum-Jane, one last optional thought. I can't help feeling `Riffer::Helpers::DeepDup` papers over a more foundational problem: we allow config hashes and arrays to be mutated rather than freezing them. If riffer froze its own copy of each on write (with `add_guardrail`/`use_mcp` building a new frozen array rather than appending), a plain `dup` would be safe to share, and a consumer who wants a change would just reassign the whole value.

ChadMcCallum-Jane:
yeah I was thinking that about deep_dup as well, it seemed like a awkward fix for an underlying problem.  I'll flag that as a separate thing to fix in another PR later

PR #445, opened 2026-09-21
fix(providers): raise IncompleteStreamError when a stream ends without its terminal event

cloetzi:
## Summary

Bedrock ConverseStream HTTP bodies can close cleanly without a `messageStop` event (observed on long generations). riffer returned each as a normal `Riffer::Messages::Assistant` with `finish_reason: nil`. Empty ones were then rejected by Bedrock on the next turn (`The content field in the Message object at messages.N is empty`); a truncated partial answer was returned as complete.

## Root cause (verified against aws-sdk-core 3.257 / aws-sdk-bedrockruntime 1.85)

- Nothing in the SDK or in riffer checks that a terminal event arrived. The event-stream decoder just decodes frames as bytes arrive; a cleanly closed chunked body is a 200 success. `AmazonBedrock#execute_stream` yields the finish reason only from the `MessageStopEvent` branch.
- `Aws::Binary::EventParser#parse_error_event` turns a `:message-type: error` frame into an `Aws::Errors::EventError` *instance* and hands it to `on_event` as an event. It is never raised by the SDK. `raise_if_stream_exception!` only matched class names ending in `Exception`, so these were silently dropped.
- Anthropic and OpenAI share the shape: their SDK iterators end cleanly at EOF and riffer yields their finish reason only on `message_stop` / `response.completed|incomplete|failed`.

## Changes

- New public `Riffer::IncompleteStreamError < Riffer::Error`, so consumers can classify it as retryable by class name.
- Bedrock, Anthropic, and OpenAI (Azure inherits) raise it when the stream returns without its terminal event. Events already yielded are still delivered; the SDK stream is closed before raising.
- Bedrock re-raises `Aws::Errors::EventError` with a `"code: message"` string while keeping `error_code` / `error_message`. Unknown events stay ignored for forward compatibility.
- Tests: empty stream, truncated stream after deltas, error frame, happy path; existing stub tests gained a `MessageStopEvent`; the Anthropic/OpenAI "close on happy path" tests split into an incomplete-stream case and a real happy path.
- Docs: `docs/STREAM_EVENTS.md` "Incomplete Streams" section, `docs/providers/CUSTOM_PROVIDERS.md` note.
- Private RBS stub for `Aws::Errors::EventError` under `sig/_private` (aws-sdk-core ships none); `sig/generated` regenerated.

## Out of scope

OpenRouter and Gemini share the shape but have no explicit terminal event. Their only signal is `finish_reason` on the last chunk, and Gemini blocked prompts legitimately return no candidates and no `finishReason`, so a naive check would misreport a content filter as truncation. Follow-up needs `promptFeedback` handling first.

## Test plan

- [x] `bin/rake` — 2366 runs, 0 failures, RuboCop clean, Steep clean

## Summary by CodeRabbit

- **New Features**
  - Streaming requests now detect incomplete provider responses and raise a retryable error instead of treating truncated output as complete.
  - Content received before an incomplete stream ends is preserved, along with completed tool-call steps and the original prompt in session history.
  - Streaming errors from Amazon Bedrock now include clearer details.

- **Documentation**
  - Added guidance for handling incomplete streams and retrying requests without resupplying the prompt.
  - Updated custom-provider requirements for detecting missing terminal events.


ChadMcCallum-Jane:
@cloetzi What does a retry look like here?  If the stream is cut off mid-generation, is the entire previous assistant message dropped or is the fragment included in the retry request and the assistant generation picks up where it left off? And if so, does it do it as a new assistant message or do we append it to the previous one?

cloetzi:
@ChadMcCallum-Jane OK, so the doc example was wrong. The implementation doesn't behave any different to any other errors that raise in the context of the stream. This doesn't modify the existing session history. Streaming fragments are not stored, they are just passed into the proc. The implementer is responsible for the retry.

lib/riffer/providers/amazon_bedrock.rb
```diff
@@ -264,20 +266,46 @@ def execute_stream(params, yielder)
...
+          message_stopped = true
           yield_finish_reason(yielder, build_finish_reason(event.stop_reason))
         when Aws::BedrockRuntime::Types::ConverseStreamMetadataEvent
           handle_metadata_usage(event, state: current_state, yielder: yielder) if event.usage
+        when Aws::Errors::EventError
+          # The SDK turns an event-stream +:message-type: error+ frame into an
+          # EventError instance and hands it to this block as an event; it is
+          # never raised. Re-raise it here (Exception#exception clones the
+          # error with a useful message while keeping error_code/error_message)
+          # so the failure surfaces instead of truncating the stream.
```
bottrall: suggestion(non-blocking): Cut out the part that explains how re-raising works.


```suggestion
            # The SDK turns an event-stream +:message-type: error+ frame into an
            # EventError instance and hands it to this block as an event; it is
            # never raised. Re-raise it here so the failure surfaces instead of
            # truncating the stream.
```

lib/riffer/providers/amazon_bedrock.rb
```diff
@@ -264,20 +266,46 @@ def execute_stream(params, yielder)
           handle_content_block_stop_text_delta(event, state: current_state, yielder: yielder) if current_state[:text]
           handle_content_block_stop_tool_use(event, state: current_state, yielder: yielder) if current_state[:tool_call]
         when Aws::BedrockRuntime::Types::MessageStopEvent
+          message_stopped = true
           yield_finish_reason(yielder, build_finish_reason(event.stop_reason))
         when Aws::BedrockRuntime::Types::ConverseStreamMetadataEvent
           handle_metadata_usage(event, state: current_state, yielder: yielder) if event.usage
+        when Aws::Errors::EventError
```
bottrall, line 273: thought(non-blocking): I'm wondering if the bigger issue here is that `raise_if_stream_exception!` assumed all bedrock exception classes will end in `Exception`. So with that line of thinking are we potentially missing other errors that end in `Error`. If so would the safer choice be to change `raise_if_stream_exception!` to also look for classes ending in `Error`.

lib/riffer/providers/open_ai.rb
```diff
@@ -218,6 +218,8 @@ def execute_stream(params, yielder)
       tool_info: {},
     } #: Hash[Symbol, untyped]
 
+    response_finished = false
```
bottrall: nitpick(non-blocking): Name this consistently.

We call it `message_stopped` in anthropic and `response_finished` in openai. Could we pick one? 😅

bottrall (approved):

lib/riffer/providers/open_ai.rb
cloetzi: [607d9b2](https://github.com/janeapp/riffer/pull/445/commits/607d9b2670e7415892853614fb14117c8d13dee2)

lib/riffer/providers/amazon_bedrock.rb
cloetzi: [607d9b2](https://github.com/janeapp/riffer/pull/445/commits/607d9b2670e7415892853614fb14117c8d13dee2)

lib/riffer/providers/amazon_bedrock.rb
cloetzi: So the bot insists that we are good now 😄 The main problem is that `Aws::Errors::EventError` is not actually a bedrock error, but coming from the underlying aws core package. Our error handling currently assumes it would be a bedrock error. We could harden it, but should probably tackle it in a separate PR.

PR #446, opened 2026-09-22
feat(messages): persist reasoning on assistant messages

bottrall:
## Problem

Riffer streams a model's reasoning as `ReasoningDelta` / `ReasoningDone` events and then forgets it. Nothing lands on the `Assistant` message, nothing survives `to_h` / `from_hash`, and nothing is handed back to the provider on the next turn. In an agentic loop every step rebuilds history from stored messages, so the model loses every thought it had, and providers that require their own signed thinking blocks back (Anthropic, Bedrock, OpenAI Responses, Gemini) cannot be given them. Each provider encodes reasoning differently, so a single string is not enough: we need a primitive that can hold every provider's shape losslessly while still feeling native to riffer and being easy for hosts that persist sessions to store and replay.

## Solution

This PR adds the primitive and the transport only. Real adapters adopt it in focused follow-ups, one per provider.

`Riffer::Messages::Assistant::ReasoningPart` is an ordered, verbatim-replayable block nested under `Assistant` (it belongs to exactly one message type, like `Assistant::ToolCall`) and modelled on OpenRouter's `reasoning_details` vocabulary: `type` (`:text`, `:summary`, `:encrypted`, validated), `text`, `data`, `signature`, `id`, and a free-form `format` tag owned by the adapter that produced it (for example `anthropic-claude-v1`). The documented replay contract is that an adapter replays only parts whose `format` it recognises, never reorders or edits them, and skips the rest, so switching models mid-session degrades gracefully.

`Messages::Assistant` carries `reasoning: Array[ReasoningPart]` beside `tool_calls`, with `reasoning?`, `reasoning_text` (joins `:text` and `:summary` parts, ignores `:encrypted`), `to_h` / `from_hash` round-tripping (key absent when empty), and concatenation on `+` so merged turns keep their parts. `StreamEvents::ReasoningDone` now carries a single `part` instead of a `content` string, so there is one source of truth for a finished reasoning block; consumers that displayed `content` read `part.text`. `Agent::Run` accumulates parts from the stream into the session message, `Agent::Response#reasoning` exposes them, `Session#update` accepts `reasoning:` so hosts can strip encrypted blobs before storage, and `Providers::Base#extract_reasoning` is a new optional hook defaulting to `[]` for the non-streaming path.

The Anthropic, OpenAI and OpenRouter adapters wrap the reasoning text they already stream in a `:text` part with no `format`. The replay contract never sends a part without a `format` back, so that text is persisted for display without any risk of a rejected request; each adapter's follow-up PR adds signatures, ids and a `format` tag to make its parts replayable. The Mock provider is the only adapter with full part support: `stub_response(reasoning: [...])` streams a `ReasoningDelta` and a `ReasoningDone` with `part:` per block before the text deltas and returns the parts from `generate_text`, so hosts and riffer's own tests can exercise persistence end to end without a live model. There is deliberately no core opt-out flag; provider-level opt-outs and `Session` hooks cover that need. Tracing capture of reasoning is left for the first real provider PR, when there is a payload to design against.

## Proof

`bin/rake` (tests, RuboCop, Steep) is green locally: 2407 runs, 0 failures. New coverage spans `ReasoningPart` construction, validation, hashing and round-tripping; `Assistant` helpers, `to_h` / `from_hash` and merge; `ReasoningDone#part`; `Run` accumulation through both the streaming and generate paths; `Session#update` preserving and replacing parts; `Response#reasoning`; and Mock event ordering. CI runs the same default task across Ruby 3.3, 3.4 and 4.0.

## Summary by CodeRabbit

* **New Features**
  * Added structured assistant reasoning parts supporting text, summaries, and encrypted data.
  * Reasoning is available on assistant messages and agent responses, with readable text extraction and serialization.
  * Streaming runs now deliver reasoning events before assistant content and preserve reasoning across session updates.
  * Providers can extract, replay, and stream reasoning data, while non-replayable reasoning remains available for display.
  * Added structured reasoning completion events with associated reasoning-part details.

* **Documentation**
  * Expanded guidance for reasoning messages, stream events, provider integration, replay, persistence, and testing.


lib/riffer/messages/base.rb
```diff
@@ -22,10 +22,14 @@ def self.from_hash(msg)
       tool_calls = (msg[:tool_calls] || []).map do |tc|
         tc.is_a?(Riffer::Messages::Assistant::ToolCall) ? tc : Riffer::Messages::Assistant::ToolCall.new(**tc)
       end
+      reasoning = (msg[:reasoning] || []).map do |part|
+        Riffer::Messages::Assistant::ReasoningPart.from_hash(part)
+      end
```
bottrall, line 27: I'm now realizing that we've created an in consistent pattern with nested objects in messages:
- `Riffer::Messages::FilePart` > PORO only for user messages but is not in the user namespace
- `Riffer::Messages::Assistant::ToolCall` > Struct only for assistant messages in the assistant namespace, defined in the assistant message class.
- `Riffer::Messages::Assistant::ReasoningPart` > PORO only for assistant messages in the assistant namespace, defined in its own module.

I'd like to create a consistant pattern here, but will address in a follow up. Likely will make filepart and toolcall follow reasoning parts pattern. PORO in its hosts namespace, in its own module.

lib/riffer/messages/base.rb
bottrall, line 1: Additionally we should probably define a `from_hash` method on each message class. That way `Riffer::Messages::Base.from_hash` doesn't have to concern itself with how to build all of its children.

docs/providers/MOCK_PROVIDER.md
```diff
@@ -47,6 +47,23 @@ provider.stub_response("Based on the tool result, here's my answer.")
 response = agent.generate("Use the tool")
 ```
 
+## Stubbing Reasoning
+
+Stub [reasoning parts](../MESSAGES.md#reasoning) to exercise a host's persistence of them. Hashes are normalized into `Riffer::Messages::Assistant::ReasoningPart`s:
```
ChadMcCallum-Jane: Is this something that Riffer takes care of automatically? (sending stubbed versions of reason chunks when sending back to the model)  Is it based on model capability, or just the values returned from the originally generated reasoning segments?

ChadMcCallum-Jane (approved):

docs/providers/MOCK_PROVIDER.md
bottrall: Not yet. In this PR we only persist reasoning. Replay will lands in the per-provider follow-ups.

docs/MESSAGES.md
```diff
@@ -124,6 +124,61 @@ msg = Riffer::Messages::Assistant.new('{"sentiment":"positive"}', structured_out
 msg.to_h  # => {role: :assistant, content: '{"sentiment":"positive"}', structured_output: {sentiment: "positive"}}
 ```
 
+#### Reasoning
+
+Reasoning models emit thinking blocks alongside their answer, and several providers require those blocks back — verbatim — on the next turn of a tool-calling loop. `Riffer::Messages::Assistant::ReasoningPart` is the neutral container riffer stores them in: a list of parts on the assistant message, in the order the provider emitted them.
```
cloetzi: suggestion(non-blocking): 
```suggestion
Reasoning models emit thinking blocks alongside their answer, and several providers require those blocks back on the next turn of a tool-calling loop. `Riffer::Messages::Assistant::ReasoningPart` is the normalized container riffer stores them in: a list of parts on the assistant message, in the order the provider emitted them.
```

docs/MESSAGES.md
```diff
@@ -124,6 +124,61 @@ msg = Riffer::Messages::Assistant.new('{"sentiment":"positive"}', structured_out
...
+
+```ruby
+thinking = Riffer::Messages::Assistant::ReasoningPart.new(type: :text, text: "The user wants the answer.", format: "mock-v1")
+opaque = Riffer::Messages::Assistant::ReasoningPart.new(type: :encrypted, data: "b3BhcXVl", signature: "sig", format: "mock-v1")
+msg = Riffer::Messages::Assistant.new("42", reasoning: [thinking, opaque])
```
cloetzi: thought(non-blocking): I hope you put 42 as an answer on purpose 😂

cloetzi: question(non-blocking): Is this example realistic? As in, do some providers emit a clear-text and encrypted thinking block in the same message? If not, could we replace it with a more realistic multi-thinking example?

docs/MESSAGES.md
```diff
@@ -124,6 +124,61 @@ msg = Riffer::Messages::Assistant.new('{"sentiment":"positive"}', structured_out
...
+#      {type: :text, text: "The user wants the answer.", format: "mock-v1"},
+#      {type: :encrypted, data: "b3BhcXVl", signature: "sig", format: "mock-v1"}
+#    ]}
+
+Riffer::Messages::Base.from_hash(msg.to_h).reasoning  # => [ReasoningPart, ReasoningPart]
+```
+
+**The replay contract.** A provider adapter replays only the parts whose `format` it recognizes and silently skips the rest, so history that travelled through another provider is never rejected. Parts are never reordered, merged, or edited — riffer treats them as opaque, because the provider's signature covers their exact bytes.
+
+A host that would rather not store parts at all can drop them as messages arrive, with [message ids](#ids) enabled so `Session#update` can address the message:
```
cloetzi: question(non-blocking): I am not following what this means and why we need an ID to drop thinking? Also, the word `host` here is confusing, do you mean `implementer`?

lib/riffer/messages/assistant.rb
```diff
@@ -89,6 +118,7 @@ def to_h
     hash = { role: role, content: content } #: Hash[Symbol, untyped]
     hash[:id] = id if id
     hash[:tool_calls] = tool_calls.map(&:to_h) unless tool_calls.empty?
+    hash[:reasoning] = reasoning.map(&:to_h) unless reasoning.empty?
```
cloetzi: suggestion(non-blocking): 
```suggestion
    hash[:reasoning] = reasoning.map(&:to_h) unless reasoning?
```

cloetzi (approved):
Looks good, left some minor suggestions mostly around docs.

docs/MESSAGES.md
bottrall: Nope, didn't even realize the robot was making a joke 😂

bottrall: Good catch that the example didn't explain itself. It is a real shape, just not obviously so: OpenAI's Responses API returns a reasoning item as summary text plus an `encrypted_content` payload (OpenRouter relays that as `reasoning.summary` and `reasoning.encrypted` side by side), and Anthropic pairs a `thinking` block with a `redacted_thinking` block when it redacts part of the chain of thought.

Updated the example to use `:summary` + `:encrypted`, which matches the OpenAI case exactly, and added a sentence under the snippet naming both providers so it reads as what you'll actually see rather than a contrived pairing.

— 🤖 Claude (beep boop)

docs/MESSAGES.md
bottrall: Fair on both counts. The recipe was over-engineered: `Session#update` is the only in-memory mutation API and it addresses messages by id, which is why the snippet dragged message ids in. But the realistic way to not store parts is simply to leave the key out when you serialize, so the docs now show `msg.to_h.except(:reasoning)` and mention `Session#update` in one sentence as the mid-run alternative that needs ids.

"Host" is gone too. "Implementer" fits the custom-provider page, but the other three uses are about the app persisting a session, so it's "your application" everywhere now.

— 🤖 Claude (beep boop)

PR #449, opened 2026-09-22
refactor(messages): give each message class its own from_hash

bottrall:
## Problem

Follow-up to review feedback on #446 (https://github.com/janeapp/riffer/pull/446#discussion_r4072780971): `Riffer::Messages::Base.from_hash` had to know how to build every one of its subclasses, including their nested parts (files, tool calls, reasoning). Construction knowledge lived in the parent rather than with each message type.

## Solution

`User`, `Assistant`, `System`, and `Tool` each now define `self.from_hash`, following the same pattern as the nested `User::FilePart`, `Assistant::ToolCall`, and `Assistant::ReasoningPart`: accept a hash or an existing instance, return the instance unchanged, otherwise build from symbol keys. `Base.from_hash` keeps the passthrough and role validation, then dispatches to the matching class.

Behaviour of `Base.from_hash` is unchanged for all existing callers. One subtlety: `User.from_hash` (etc.) previously resolved to the inherited `Base.from_hash` and dispatched on `:role`; it now always builds its own class and ignores `:role`. Nothing in `lib/` called a subclass's `from_hash` directly. `docs/MESSAGES.md` mentions the per-role entry point.

## Proof

New `.from_hash` specs per message class (hash construction incl. nested parts, instance passthrough, `to_h` round-trip for `User`, invalid `finish_reason` for `Assistant`); existing `Base.from_hash` specs pass untouched. `bin/rake` (tests + rubocop + steep) green locally; CI covers the rest.

## Summary by CodeRabbit

* **New Features**
  * Added role-specific `from_hash` conversions for user, assistant, system, and tool messages, preserving IDs, file parts, tool-call details, structured output, and finish-reason fields.
  * Updated base conversion to route role creation through the corresponding role implementations.
  * Existing message objects are returned unchanged when converted.
* **Bug Fixes**
  * Invalid assistant tool calls, reasoning parts, or finish reasons now raise a clear argument error.
* **Documentation**
  * Clarified guidance for role-specific `from_hash` normalization.
* **Tests**
  * Added coverage for the new role-specific conversions.


lib/riffer/messages/assistant.rb
```diff
@@ -7,6 +7,25 @@ class Riffer::Messages::Assistant < Riffer::Messages::Base
...
+
+    new(
+      msg[:content],
+      id: msg[:id],
+      tool_calls: (msg[:tool_calls] || []).map { |tc| Riffer::Messages::Assistant::ToolCall.from_hash(tc) },
+      reasoning: (msg[:reasoning] || []).map { |part| Riffer::Messages::Assistant::ReasoningPart.from_hash(part) },
+      structured_output: msg[:structured_output],
+      finish_reason: msg[:finish_reason]&.to_sym,
+      finish_reason_raw: msg[:finish_reason_raw],
+    )
```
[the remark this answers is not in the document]
bottrall: Thanks. This gap was already there before this PR: the old `Base.from_hash` dropped these fields too, and this PR only moves that logic into each class. We'll fix it in a follow-up PR stacked on this one.

Note for the follow-up: passing `msg[:token_usage]` straight through won't work, because `to_h` writes a plain Hash but the constructor expects a `Riffer::Providers::TokenUsage`. It needs a `TokenUsage.from_hash`. `error_type` also needs `&.to_sym` to survive a JSON round trip.

> — 🤖 Claude (beep boop)

lib/riffer/messages/base.rb
```diff
@@ -15,31 +15,11 @@ def self.from_hash(msg)
...
-    when :system
-      Riffer::Messages::System.new(msg[:content], id: msg[:id])
-    when :tool
-      Riffer::Messages::Tool.new(msg[:content], id: msg[:id], tool_call_id: msg[:tool_call_id], name: msg[:name])
-    else
-      raise Riffer::ArgumentError, "Unknown message role: #{msg[:role]}"
+    when :user then Riffer::Messages::User.from_hash(msg)
+    when :assistant then Riffer::Messages::Assistant.from_hash(msg)
+    when :system then Riffer::Messages::System.from_hash(msg)
+    when :tool then Riffer::Messages::Tool.from_hash(msg)
```
Thibaut: suggestion(non-blocking): we can make this even more OO now that they all "quack" the same way. I'd extract a mapping of role/class and then we can do something like:

```ruby
klass = CLASS_BY_ROLE[msg[:role]]
raise Riffer::ArgumentError, "Unknown message role: #{msg[:role]}" unless klass
klass.from_hash(msg)
```

Thibaut (approved):
👌

lib/riffer/messages/base.rb
bottrall: I would agree if there was more than one use for `CLASS_BY_ROLE`. But as it stands it would be around the same amount of code but with some slight indirection. I'm going to keep as is.

PR #450, opened 2026-09-22
fix(messages): preserve tool errors and token usage through from_hash

bottrall:
## Problem

Serializing a message with `to_h` and then loading it with `from_hash` loses data on two message types. This was already true before #449, and CodeRabbit flagged it on that PR ([review](https://github.com/janeapp/riffer/pull/449#pullrequestreview-5284122032)):

- `Tool#to_h` writes `error` and `error_type`, but `Tool.from_hash` never reads them. A failed tool message comes back as a success.
- `Assistant#to_h` writes `token_usage` as a plain hash, but `Assistant.from_hash` drops it.
- `FilePart#to_h` writes both `data` and `url` when a part has both, but `FilePart.from_hash` sends any hash with a `url` through `from_url`, which takes no `data`, so the data is dropped. The new round-trip tests below found this one.

## Solution

`Tool.from_hash` now passes `error` and `error_type`, with `error_type` converted by `&.to_sym`. That matches how `Assistant.from_hash` handles `finish_reason`, and keeps it a `Symbol` after a JSON round trip.

For token usage, this adds `Riffer::Providers::TokenUsage.from_hash`, built the same way as the existing nested `from_hash` methods (`ToolCall`, `ReasoningPart`, `FilePart`): an existing instance passes through unchanged, otherwise it builds from `input_tokens`, `output_tokens` and the optional `cache_write_tokens`, `cache_read_tokens` and `cost`. `Assistant.from_hash` calls it only when a `token_usage` hash is present. CodeRabbit suggested passing the raw hash straight to the constructor, but that would break `message.token_usage.input_tokens`, since the constructor expects a `TokenUsage`. So this PR doesn't take that suggestion.

`FilePart.from_hash` now builds directly with `new(url:, data:, …)`. It still detects `media_type` from the URL when missing, using a private `detect_media_type` that `from_url` now shares. Data the file resolver downloads later isn't affected, since `to_h` leaves it out on purpose.

Following Thibaut's review comment, a shared `assert_round_trips` helper in `test/test_helper.rb` guards against this kind of drift. It reloads an object through `to_h` → JSON → `from_hash` and compares `to_h`. It also requires `to_h` to emit every constructor keyword, so a fixture that doesn't set a newly added attribute fails instead of quietly skipping `from_hash`. Each serializable class (the four messages, `ToolCall`, `ReasoningPart`, `FilePart`, `TokenUsage`) calls it from its own test file with a fully-populated fixture.

Stacked on #449. Target is `feat/messages-children-from-hash`.

## Proof

New tests:
- `TokenUsage.from_hash`: a hash with the optional keys, a hash without them, and instance passthrough.
- An errored `Tool` through `to_h` → `from_hash`, both directly and via `JSON.parse(..., symbolize_names: true)`.
- An `Assistant` with `token_usage` through `to_h` → `from_hash`.
- A fully-populated `assert_round_trips` test for each serializable class. The `FilePart` one failed before the fix. Removing a keyword from a fixture fails with e.g. `FilePart#to_h omits [:data]`.

`bin/rake` passes locally (tests, rubocop, steep). CI covers the rest.

## Summary by CodeRabbit

* **Bug Fixes**
  * Assistant messages retain typed token usage when reconstructed.
  * Tool messages preserve error details and error types through hash and JSON round trips.
  * URL-backed files infer their media type from the file extension when one isn’t provided; unsupported extensions produce an error.
* **Improvements**
  * Token usage can be reconstructed from hashes, including optional token and cost values; existing token usage values remain unchanged.


Thibaut (approved):
thought(non-blocking): I wonder if we could have a test that ensures things never fall out of sync again.

bottrall:
@Thibaut good call — added in d343c99. There's now an `assert_round_trips` helper in `test_helper.rb` that each serializable class calls from its own test file. It reloads through `to_h` → JSON → `from_hash`, and it also requires `to_h` to emit every constructor keyword. So a new attribute fails the test until the fixture sets it, which then puts `from_hash` under test.

It paid off right away: it found that `FilePart.from_hash` dropped `data` when a `url` was also present. Fixed in 5258d2e.

— 🤖 Claude (beep boop)

PR #452, opened 2026-09-22
feat: identify agents and evals with default tags

cloetzi:
## Why

A provider can't tell which agent or eval a call is on behalf of. That's a problem for providers that forward calls to a service needing caller identity, e.g. one that runs each call as a step on an orchestration service and needs the agent identifier.

Other frameworks don't have a separate "attribution" concept; they use a metadata/tags channel plus identity fields the framework fills in itself. riffer already has the channel (`tags:`), so this adds the identity riffer knows about as default tags.

## What

- **Agent runs** carry `kind: "agent"` and `agent: <agent identifier>` on every call, merged under the caller's `tags:` (a caller tag with the same key wins). They travel like any tag: to the provider's native request metadata and as `riffer.tag.*` on spans.
- **Judge calls** made through an evaluator carry `kind: "judge"` and `agent: <evaluator identifier>`.
  - New `Evaluator.identifier`, following the agent pattern: explicit value, else the snake_cased class name, else `riffer/judge` for an anonymous class.
  - `Judge.new` takes `tags:` and sends them on every call, so custom evaluators that call `judge.evaluate` directly keep them.

## Behaviour changes

- Every agent call now sends request metadata. OpenAI / Azure / OpenRouter `metadata` and Bedrock `request_metadata` include `kind` and `agent` even when the caller passes no tags. That's two of the 16 entries those providers allow. Anthropic (forwards `user_id` only) and Gemini (no request field) are unchanged.
- Judge calls through an evaluator now send the same metadata.
- Spans always have `riffer.tag.kind` and `riffer.tag.agent`.
- Custom providers now get `:tags` on every agent call and must strip it before splatting options into an SDK call. The custom provider doc examples are updated.

## Tests

- Run: default tags on generate, stream and every call in the tool loop; caller tags override them; spans carry them.
- Judge: forwards its tags; sends no `:tags` when it has none.
- Evaluator: derived, explicit and anonymous-fallback identifiers; the judge gets both tags.

No cassettes change: VCR tests call providers directly.

## Docs

- `docs/AGENTS.md`: new "Default tags" section.
- `docs/TRACING.md`: default tags on spans.
- `docs/EVALS.md`: `identifier` class method and judge tags.
- `docs/providers/CUSTOM_PROVIDERS.md`: new "Tags" section; examples strip `:tags`.

## Summary by CodeRabbit

- **New Features**
  - Agent and judge calls now include default `kind` and `agent` tags. Caller-provided tags can override these defaults.
  - Evaluators support custom identifiers, with sensible defaults based on their class names.
  - Provider integrations can use tags such as `agent` and `user_id` in service requests.

- **Documentation**
  - Updated guides to explain default tags, evaluator identifiers, and how tags are passed to providers.


lib/riffer/agent/run.rb
```diff
@@ -36,14 +36,15 @@ def stream(agent:, prompt: nil, files: nil, tags: {})
...
+  # +tags+ are normalized and merged over the default tags (a caller tag wins
+  # on a shared key). The clean <tt>String => String</tt> map is then
   # threaded to every span builder in the run as +riffer.tag.*+ and to each
   # provider call (via +merged_model_options+) for native request-metadata
   # mapping.
   #--
   #: (Riffer::Agent, ?tags: Hash[(String | Symbol), untyped]?, ?stream_yielder: Enumerator::Yielder?) -> Riffer::Agent::Response
   def run_loop(agent, tags: {}, stream_yielder: nil)
-    tags = normalize_tags(tags)
+    tags = default_tags(agent).merge(normalize_tags(tags))
```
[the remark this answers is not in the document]
cloetzi: Should be addressed in [f1c9e4e](https://github.com/janeapp/riffer/pull/452/commits/f1c9e4ec2e0b364cd56ca96b0d73b6c17a7b8bfb).

docs/AGENTS.md
```diff
@@ -99,6 +99,8 @@ end
 MyAgent.identifier  # => "custom_agent_name"
 ```
 
+An agent instance also answers `#identifier`. It returns the class identifier, unless the instance was built with its own config (e.g. restored with `Riffer::Agent.from_h`), in which case that config's identifier wins. This is the value used for the default `agent` tag and the `invoke_agent` span.
```
bottrall: suggestion(non-blocking): Remove this section, I think it makes things more confusing than they are.

The scenario it's talking about is:
```rb
some_agent_hash = { identifier: "some_agent" }
my_agent = Riffer::Agent.from_h(some_agent_hash)
my_agent.identifier # "some_agent"
```
I don't think it needs to be called out that it uses the identifier that it was given.

bottrall (approved):

docs/AGENTS.md
cloetzi: [ef3f044](https://github.com/janeapp/riffer/pull/452/commits/ef3f04490773ffdd0044c4245a8eae6d9084d3cb)

PR #453, opened 2026-09-23
feat(open_router): capture and replay reasoning_details

bottrall:
## Problem

#446 added `Assistant::ReasoningPart` and a replay contract, but OpenRouter still only wrapped its flat `reasoning` string in a `:text` part with no `format`. That text was persisted for display, yet it was never sent back, so the model lost its reasoning on every turn. For Anthropic and Gemini models behind OpenRouter this is more than a quality issue: continuing a tool-call loop needs the signed thinking blocks sent back verbatim, and OpenAI reasoning models can only reuse their encrypted reasoning if it is replayed.

## Solution

OpenRouter now reads `reasoning_details` rather than the flat `reasoning` field. On both `generate_text` (`extract_reasoning`) and `stream_text`, each detail maps losslessly onto a `ReasoningPart`: `reasoning.text`, `reasoning.summary` and `reasoning.encrypted` become `:text`, `:summary` and `:encrypted`, and `data`, `signature`, `id` and `format` are copied as they are. Detail types riffer doesn't know are skipped rather than raising.

When streaming, OpenRouter splits one block into many fragments that share an `index`. Fragments are grouped by `[index, type]` in the order they first appear. `text`, `summary` and `data` are concatenated in place, using the same `<<` pattern as the O(n²) streaming fix. `signature`, `id` and `format` take the last non-nil value, because the recorded Anthropic stream sends the signature on a final fragment with no text. `type` is part of the key because OpenRouter documents `index` as optional. Each non-empty prose fragment still yields a `ReasoningDelta`, and each block ends as a single `ReasoningDone`.

On replay, the assistant message's parts go back as `reasoning_details` in their original order and unchanged, with `summary` used as the prose key for summary parts. The key is left out when nothing qualifies. "Recognised format" means an explicit allowlist, `REASONING_FORMATS`, containing the nine formats OpenRouter documents, rather than "any part with a format". That way a `mock-v1` part, or a future format tag from another adapter, can never reach an OpenRouter request. The trade-off: if OpenRouter adds a new format, those parts are skipped until the list is updated. They are not rejected. `index` isn't stored on `ReasoningPart`, and both recorded replays were accepted without it.

`docs/providers/OPENROUTER.md` gains a "Reasoning Replay" section covering the field mapping, how fragments merge, and the format list.

## Proof

`bin/rake` (tests, RuboCop, Steep) is green locally: 2425 runs, 0 failures. It was run with a dummy `OPENROUTER_API_KEY`, so any request not already in a cassette would have failed. New VCR cassettes, each recorded once on cheap models:

- `reasoning/_stream_text/replays_reasoning_details`: `anthropic/claude-haiku-4.5` with `reasoning: { max_tokens: 1024 }`. Turn 1 streams signed reasoning, and turn 2 replays it and gets an answer.
- `reasoning/_generate_text/replays_reasoning_with_tool_result`: Haiku 4.5 with a tool. Turn 1 returns a tool call plus signed reasoning, and turn 2 replays both with the tool result.
- `reasoning/_generate_text/captures_reasoning_details`: `openai/gpt-5-nano` with low effort. It returns a `:summary` part and an `:encrypted` part tagged `openai-responses-v1`.

Unit tests cover type mapping, skipping unknown types, merging fragments with a late signature, one part per index in order, replay order and field mapping, and skipping parts with no format or `mock-v1`. Every cassette's `Authorization` header is filtered to `Bearer <OPENROUTER_API_KEY>`. CI runs the default task on Ruby 3.3, 3.4 and 4.0.

## Summary by CodeRabbit

* **New Features**
  * OpenRouter conversations now preserve supported reasoning details across responses, streamed exchanges, and follow-up turns, including interactions involving tool calls.
  * Streamed reasoning fragments are combined, with text-bearing updates available as they arrive.
* **Bug Fixes**
  * Unsupported reasoning formats and unrecognized detail types are skipped during replay.
* **Documentation**
  * Clarified how OpenRouter reasoning details are handled and replayed.


lib/riffer/providers/open_router.rb
```diff
@@ -188,6 +209,39 @@ def extract_tool_calls(response)
     end
   end
 
+  #--
+  #: (untyped) -> Array[Riffer::Messages::Assistant::ReasoningPart]
+  def extract_reasoning(response)
+    typed_response = response #: OpenAI::Models::Chat::ChatCompletion
```
cloetzi: question(non-blocking): Why is this needed? Couldn't you just use `response`?

lib/riffer/providers/open_router.rb
```diff
@@ -188,6 +209,39 @@ def extract_tool_calls(response)
...
+    message = typed_response.choices.first&.message
+    details = message && reasoning_details(message)
```
cloetzi: nitpick(non-blocking): That tripped me up when reading. I think an one line if else might improve readability.

lib/riffer/providers/open_router.rb
```diff
@@ -188,6 +209,39 @@ def extract_tool_calls(response)
...
+  #--
+  #: (untyped) -> Array[Hash[Symbol, untyped]]?
+  def reasoning_details(model)
+    model[:reasoning_details]
+  end
+
+  #--
+  #: (Hash[Symbol, untyped]) -> Riffer::Messages::Assistant::ReasoningPart?
+  def build_reasoning_part(detail)
+    type = REASONING_TYPES.fetch(detail[:type], nil)
```
cloetzi: nitpick(non-blocking): Isn't this pretty much equal to `REASONING_TYPES[detail[:type]]`?

lib/riffer/providers/open_router.rb
```diff
@@ -224,7 +278,10 @@ def execute_stream(params, yielder)
     emit_tool_call_done_events(state: state, yielder: yielder) unless state[:tool_calls].empty?
 
     yielder << Riffer::StreamEvents::TextDone.new(state[:text]) unless state[:text].empty?
-    yield_reasoning_done(yielder, state[:reasoning]) unless state[:reasoning].empty?
+    state[:reasoning_details].each_value do |detail|
+      part = build_reasoning_part(detail)
+      yielder << Riffer::StreamEvents::ReasoningDone.new(part) if part
```
cloetzi: question(non-blocking): Is this correct behavior? Does this mean we can't use a `ReasoningDone` event as an indicator that reasoning is actually done? The only reason I am bringing this up is that `TextDone` event is some kind of stop marker, and this would divert as a single reasoning message, will result in multiple done events.

cloetzi (approved):

lib/riffer/providers/open_router.rb
bottrall: It's a workaround for an rbs quirk. `response` has to be untyped in the method signature, otherwise `OpenAI::Models::Chat::ChatCompletion` would get pulled into the consumer. If the consumer doesn't have the openai gem installed it then completely breaks steep because we're referencing an undefined type. Therefore to get type safety while using response we have to do this type cast.

lib/riffer/providers/open_router.rb
bottrall: At runtime, yes it's equal. However at type checking time `REASONING_TYPES` is typed `Hash[String, Symbol]`, so Steep assumes `[]` never returns `nil` and reports the` return nil unless type` line as unreachable.

lib/riffer/providers/open_router.rb
bottrall: Yeah, this is intended. `ReasoningDone` means one reasoning block is done, not all reasoning. Each event carries the `ReasoningPart` we replay, and a turn can have several blocks, so you can get more than one. #446 set it up that way and the Mock provider works the same. `TextDone` is still the stop marker.

Fair call on timing though. We currently emit them all at the end of the stream, after `TextDone`, not when each block actually finishes. We could emit them as soon as text or a tool call starts? But if we want to do that let's do that as a followup.

